PR #25, opened 2019-07-30
[KOGITO-66] - Add ServiceAccount to the KogitoApp

ricardozanini:
This PR includes:

1. Operator SDK upgrade to 0.9.0
2. Openapi auto generation of CRDs
3. KogitoApp pods now run with a specified ServiceAccount that can fetch for Kubernetes API Resources within the namespace (like services)


sutaakar:
@ricardozanini I have noticed that some descriptions were deleted from CRD, was it intentional?

ricardozanini:
> @ricardozanini I have noticed that some descriptions were deleted from CRD, was it intentional?

No, I didn't realized  that. The CRD is now being generated automatically, so those descriptions are taken from the comments on `*_types.go` files. Thanks for pointing that out, I'm going to add them.

ricardozanini:
@tchughesiv could you take a look when you have the time, please? :nerd_face:

pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -138,7 +182,7 @@ type KogitoApp struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   KogitoAppSpec   `json:"spec,omitempty"`
+	Spec   *KogitoAppSpec  `json:"spec"`
```
tchughesiv: could you explain why this is being made a pointer? i don't see any corresponding code to indicate a reason for the change changes which is why i ask

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -431,6 +455,40 @@ func (r *ReconcileKogitoApp) newDCForCR(cr *v1alpha1.KogitoApp, serviceBC obuild
...
+
+func newSARoleBinding(cr *v1alpha1.KogitoApp, sa corev1.ServiceAccount) (roles rbacv1.RoleBindingList) {
+	roles = rbacv1.RoleBindingList{}
+	roles.Items = append(roles.Items, rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: sa.Namespace,
+			Name:      fmt.Sprintf("%s-%s", sa.Name, constants.ServiceAccountRole),
+		},
+		RoleRef: rbacv1.RoleRef{
+			Kind: "ClusterRole",
```
tchughesiv: is the intention here to give all kogito deployments `view` rights to the cluster? if so, this seems excessive. what is the purpose of this change?

pkg/apis/app/v1alpha1/kogitoapp_types.go
ricardozanini: the pointer makes the spec required in the crd. But thinking again, is a little too excessive. I believe that there's no need to do that.

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: this is the role that we've been testing with the service discovery mechanism to access the Core Services API. Sometimes, this services are in other namespaces, that's why we need a `ClusterRole`. Makes sense?

tchughesiv:
@ricardozanini do the operator-sdk and go version requirements need to change in the readme? circleci updated?

ricardozanini:
> @ricardozanini do the operator-sdk and go version requirements need to change in the readme? circleci updated?

I changed the circleci when we upgraded the GO version in other PR. The operator-sdk doesn't matter for it.

tchughesiv:
LGTM

PR #26, opened 2019-08-01
[KOGITO-84] - Introduction to definitions package

ricardozanini:
**Note:** I'll leave this PR opened for others to take a look while we work on the redesign. 

This first PR intends to add the "definition" concept to the architecture, which are the data structures that define how the Kubernetes/OpenShift resources will be created to attend the CRD requirement.

The implementation can be viewed on `pkg/controller/<crd>/definitions`. Each file represents a Kubernetes/OpenShift resource  that know how to create each object.

@sutaakar @errantepiphany please chime in. @tchughesiv only if you have the time. :stuck_out_tongue: 

Reference: https://docs.google.com/presentation/d/1xpWaqcTbvb87_uAfNZW1EJg4zJnV5zVLwYRU04D-Kms

### DIFF
--- a/pkg/controller/kogitoapp/constants/constants.go
+++ b/pkg/controller/kogitoapp/constants/constants.go
@@ -9,10 +9,6 @@ const (
 	ImageStreamTag = "0.2.0"
 	// ImageStreamNamespace default namespace for the ImageStreams
 	ImageStreamNamespace = "openshift"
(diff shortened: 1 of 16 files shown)


tchughesiv:
lgtm

ricardozanini:
@tchughesiv  many thanks for reviewing it. I just removed the `factory` thing following Babak's insights. Could you take a second look before I squash them?

PR #31, opened 2019-08-15
[KOGITO-137] - Operator refactoring: packages and functions reorganization

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-137

PR for applying community and peers feedback regarding Operator SDK best practices. See this [comment](https://github.com/kiegroup/kogito-cloud-operator/pull/31#issuecomment-522259273) for a detailed explanation of what this PR is planing to achieve.


ricardozanini:
@xiezhang7 could you take a quick look? Mainly on the package names.

xiezhang7:
Hi @ricardozanini , I would suggest moving the `pkg/controller/kogitoapp/definitions` package under the `builder` package, and rename it to `openshift` or something else. Because I think most of the public functions inside this package are used to create openshift/kubernetes objects. Everything else looks good to me. Thanks.

ricardozanini:
@bmozaffa I'd like to ask for your review in this PR. What we're trying to accomplish here is a clear separation of concerns between what the Operator is doing inside the `Reconcile` loop, support operations and OpenShift/Kubernetes calls. The motivation behind this is to be able to reuse those packages in the CLI and other controllers that we are going to build in the near future.

The `KogitoApp` controller goal is to deploy a Kogito Service from source using s2i builds. This deploy relies on the following OpenShift/Kube objects:

https://raw.githubusercontent.com/kiegroup/kogito-cloud-operator/master/docs/img/kogitoapp_resource_composition.png

The packages are structured like the following:

```
├── cmd
│   ├── kogito
│   └── manager
├── pkg
│   ├── apis
│   │   └── app
│   │       └── v1alpha1
│   ├── client
│   │   ├── kubernetes
│   │   ├── meta
│   │   └── openshift
│   ├── controller
│   │   └── kogitoapp
│   │       ├── builder
│   │       ├── shared
│   │       ├── status
│   │       └── test
│   ├── logger
│   └── util
├── test
│   └── e2e
└── version
```
(omitting support directories for brevity)

A brief explanation of each one:

1. `cmd`: Operator binaries. 
1.1 `manager` is the tool generated by the SDK
1.2 `kogito` is our CLI
2. `pkg`: reusable functions/controller implementation
2.1 `apis`: auto-generated functions by the SDK
2.2 `client`: abstraction of Kubernetes and OpenShift APIs. We're using the same functions for both the controllers and the CLI. This way we can have a high degree of maintainability, reuse and separation of concerns. When the SDK creates a `manager`, also creates a Kubernetes `cli` that we're wrapping inside this package to be used later. When we perform CLI commands, we also creates the same `cli` based on the kubernetes configuration of the host machine (in the future we'll allow the user to pass the cluster credentials). The functions inside this package doesn't perform only CRUDs, but also have meaningful use cases like [triggering a new build](https://github.com/kiegroup/kogito-cloud-operator/blob/6ac7ec79acfae2c72bed895855fc8024bb49095c/pkg/client/openshift/buildconfig.go#L73). Thanks to our client wrapping, we can also use mocks (`fake` clients) to easily perform unit tests.
2.3 `controller`: controllers implementation and functions that only concerns the controller operation
2.3.1 `shared`: to be deprecated by `utils` package
2.3.2 `builder`: has all functions that create and persist the OpenShift/Kubernetes object structure for this controller.
2.4 `logger`: Uber Zap logger wrapping to be used by the entire package
2.5 `util`: utilities functions
2.6: `test`: integration tests (end-to-end)

I understand that it's not perfect, but we're aiming to mature this project thorough the releases and will always try to perform minor refactoring to this according to the non-functional requirements.

As soon as you finish your review, I'm going to update the docs to reflect this refactoring. We intent to freeze this branch on Tuesday, to leave QE with time to verify the Operator Use Cases by Friday for the v0.3 release.

@camilamacedo86, @tchughesiv, @xiezhang7 and @sutaakar please if you guys have the time, I'd like to hear from you too.

Thank you all!

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -6,44 +6,127 @@ import (
...
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	imageClient, err := imagev1.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.Errorf("Error getting image client: %v", err)
+		return &ReconcileKogitoApp{}
```
bmozaffa: Not sure how this bubbles up, but you are expected to return the error here, which would presumably be a second return argument. The caller would check it and decide if to proceed.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -6,44 +6,127 @@ import (
...
+	}
+	imageClientInterface := imagev1.ImageV1Interface(imageClient)
+
+	buildClient, err := buildv1.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.Errorf("Error getting build client: %v", err)
+		return &ReconcileKogitoApp{}
+	}
+
+	buildClientInterface := buildv1.BuildV1Interface(buildClient)
```
bmozaffa: This is unnecessary, and confusing. Just use buildClient directly on line 69.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -6,44 +6,127 @@ import (
...
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	imageClient, err := imagev1.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.Errorf("Error getting image client: %v", err)
+		return &ReconcileKogitoApp{}
+	}
+	imageClientInterface := imagev1.ImageV1Interface(imageClient)
```
bmozaffa: This is unnecessary, and confusing. Just use imageClient directly on line 70.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -67,10 +150,9 @@ func (r *ReconcileKogitoApp) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	// create resources in the cluster that not exists
-	kogitoInv, err := inventory.CreateResources(&inventory.BuilderContext{
```
bmozaffa: This is old code, but on 145, you can simplify by using if "instance.Spec.Runtime != v1alpha1.SpringbootRuntimeType" and nothing else. You can keep the QuarkusRuntimeType check if you think that will be too confusing, but certainly don't need to check against empty.

pkg/client/kubernetes/resource.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ResourceInterface has functions that interacts with any resource object in the Kubernetes cluster
+type ResourceInterface interface {
+	CreateIfNotExists(meta.ResourceObject) (bool, error)
+	FetchWithKey(types.NamespacedName, meta.ResourceObject) (bool, error)
+	Fetch(meta.ResourceObject) (bool, error)
+	ListWithNamespace(string, runtime.Object) error
```
bmozaffa: Make your interfaces more readable by providing argument names, e.g.,
`FetchWithKey(key types.NamespacedName, resource meta.ResourceObject) (bool, error)`

pkg/client/openshift/buildconfig.go
```diff
@@ -0,0 +1,137 @@
...
+type BuildState struct {
+	ImageExists  bool
+	BuildRunning bool
+}
+
+// BuildConfigInterface exposes OpenShift BuildConfig operations
+type BuildConfigInterface interface {
+	// EnsureImageBuild checks for the corresponding image for the build. If there's no image, verifies if the build still running.
+	// Returns a BuildState structure describing it's results. Label selector is used to query for the right bc
+	EnsureImageBuild(bc *buildv1.BuildConfig, labelSelector string) (*BuildState, error)
```
bmozaffa: Why is the return type, build state, a pointer? This question also potentially applies to buildconfig and others.

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: This function is generated by the SDK itself and the interface doesn't have a way to return an error in its signature. We're using the image and build clients for specific OpenShift use cases for this version (should be removed in later versions). Unfortunately those clients builders return an error and there's no other way to handle it (despite logging). Nor the `reconciler` or the `manager` has an interface to set the error.

I completely agree with you, so I'm going to `panic` in this method so the operator won't come up and the user should see in the logs what's happened.

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: LOL. I completely forgot to remove it. Thanks!

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: Absolutely

pkg/client/openshift/buildconfig.go
ricardozanini: Yeah, for those types doesn't make sense returning a pointer.

ricardozanini:
@bmozaffa many thanks for your time (on a weekend!) for reviewing our PR. Much appreciated. I reviewed the code and updated the documentation.

pkg/logger/logger.go
```diff
@@ -13,12 +13,20 @@ import (
...
+
 // Logger shared logger struct
 type Logger struct {
 	Logger        logr.Logger
 	SugaredLogger *zap.SugaredLogger
 }
 
+// GetLoggerWithOut returns a custom named logger to out in a specific writer
+func GetLoggerWithOut(name string, w io.Writer) *zap.SugaredLogger {
+	outputTo = w
```
xiezhang7: Is this going to affect the result of future invocation of function `GetLogger`, which could make `GetLogger` unpredictable?

ricardozanini: Absolutely. Each `GetLogger*` will have its own output since they create a new logger each time is called. Calling `GetLogger` will force the output to the `os.StdErr`. This one has an option to output to another `writer`. Is used during CLI unit tests to verify the output of each command and assert accordingly.

xiezhang7: Then, I think we should make sure calling `GetLogger`  **directly**  always forces the output to `os.StdErr` (e.g. reset `outputTo` to `os.StdErr` before exiting `GetLoggerWithOut` ). Otherwise we won't  be able to know which output it uses after `GetLoggerWithOut` is called. Another option maybe we can provide a function to change the `outputTo` directly.

ricardozanini: Done!

PR #32, opened 2019-08-16
[KOGITO-82] - Add support for service label definitions in Kogito Operator

xiezhang7:
Jira: https://issues.jboss.org/browse/KOGITO-82

Support the capability to define service labels in CR in the form: `spec.service.labels=map[string]string`. 
The service labels defined in CR will be applied to the Kubernetes Service of the Kogito App for service discovery. 
If service labels are not defined in the CR, `service=kogitoApp.Spec.Name` will be applied as the default label for the Service.


xiezhang7:
Hi @ricardozanini  @bmozaffa @tchughesiv , I don't think I can add reviewers or labels. Could you please review this pull request? Thanks.

ricardozanini:
/LGTM

@bmozaffa do you mind taking a quick look? This requirement is to add labels to the application service, so the Service Discovery engine will be able to find it. See: https://github.com/kiegroup/kogito-examples/blob/master/onboarding-example/onboarding/readme.md

This will basically add to the CRD a way to the user inform the label for services.

deploy/crds/app_v1alpha1_kogitoapp_crd.yaml
```diff
@@ -125,6 +125,14 @@ spec:
               description: The name of the runtime used, either quarkus or springboot,
                 defaults to quarkus
               type: string
+            service:
+              description: Service configuration
+              properties:
+                labels:
```
sutaakar: Can you please add a description for this property?

xiezhang7: Description added. Thanks.

deploy/crds/app_v1alpha1_kogitoapp_crd.yaml
```diff
@@ -125,6 +125,15 @@ spec:
...
+                  additionalProperties:
+                    type: string
+                  description: Labels for the application service
+                  type: object
```
bmozaffa: I don't remember off the top of my head, but isn't there a better way to represent labels?

ricardozanini: @bmozaffa we’re using map[string]string for representing the Labels object.

bmozaffa: Right, I mean let's find the right way to represent it in CRD too, so it's validated and now free form.

xiezhang7: @bmozaffa please find this example of [string to string mapping](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#model-with-mapdictionary-properties) for OpenAPI which is the same as what we do. This yaml is auto-generated from the [kogitoapp_types](https://github.com/kiegroup/kogito-cloud-operator/blob/d7b0a6214eb55359e66004c93227bbc33a1ef87a/pkg/apis/app/v1alpha1/kogitoapp_types.go#L80) definition, and the kogitoapp_types definition is based on the type definition from [apimachinery](https://github.com/kubernetes/apimachinery/blob/533d101be9a6450773bb2829bef282b6b7c4ff6d/pkg/apis/meta/v1/types.go#L234) code. Thanks.

bmozaffa: Thanks, Xieshen, for looking into it. The CRD generation is not always the best, but if we don't have any example of trying to describe and enforce labels anywhere, whether possible or not, probably not worth trying to do it.

xiezhang7: @bmozaffa The [validation](https://github.com/kiegroup/kogito-cloud-operator/blob/d7b0a6214eb55359e66004c93227bbc33a1ef87a/deploy/crds/app_v1alpha1_kogitoapp_crd.yaml#L14) is via [OpenAPI V3 Schema](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation). In the OpenAPI V3 Schema Object [example](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#model-with-mapdictionary-properties), string to string mapping is like what we do here.

pkg/apis/app/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -206,6 +206,29 @@ func (in *KogitoAppList) DeepCopyObject() runtime.Object {
 	return nil
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *KogitoAppServiceObject) DeepCopyInto(out *KogitoAppServiceObject) {
```
bmozaffa: Did this not get re-generated?

xiezhang7: It is re-generated.

ricardozanini:
@xiezhang7 Is this PR ready? (I saw that you updated the title)

xiezhang7:
> @xiezhang7 Is this PR ready? (I saw that you updated the title)

Yes. I think so. Thanks.

ricardozanini:
@xiezhang7 could you please then squash your commits into one? Thanks!

xiezhang7:
> @xiezhang7 could you please then squash your commits into one? Thanks!

@ricardozanini I have squashed the commits. Thanks.

PR #53, opened 2019-09-10
[KOGITO-252] - Add an option to Kogito CLI to deploy the Data Index S…

ricardozanini:
…ervice

See:
https://issues.jboss.org/browse/KOGITO-252

### DIFF
--- a/cmd/kogito/command_checks.go
+++ b/cmd/kogito/command_checks.go
@@ -2,9 +2,9 @@ package main
 
 import "fmt"
 
-// checkProjecLocally will verify if the project/namespace exists in the CLI context.
+// checkProjectLocally will verify if the project/namespace exists in the CLI context.
 // Won't fecth the cluster to verify if the project/namespace exists. This is a local validation only.
-func checkProjecLocally(project string) (localProject string, err error) {
+func checkProjectLocally(project string) (localProject string, err error) {
 	if len(project) == 0 {
 		if len(config.Namespace) == 0 {
 			return "", fmt.Errorf("Couldn't find any Project in the current context. Use 'kogito use-project NAME' to set the Kogito Project where the service will be deployed or pass '--project NAME' flag to this one")
--- a/cmd/kogito/deploy_data_index.go
+++ b/cmd/kogito/deploy_data_index.go
@@ -1,0 +1,159 @@
+package main
+
+import (
+	"fmt"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	resdataindex "github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitodataindex/resource"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/util"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/spf13/cobra"
+)
+
+type deployDataIndexFlags struct {
+	deployCommonFlags
+	image              string
+	kafka              v1alpha1.KafkaConnectionProperties
+	infinispan         v1alpha1.InfinispanConnectionProperties
+	infinispanSasl     string
+	infinispanUser     string
+	infinispanPassword string
+}
+
+const (
+	defaultDataIndexName         = "kogito-data-index"
+	defaultInfinispanSecretName  = "kogito-data-index-infinispan-credentials"
+	defaulInfinispanUsernameKey  = "username"
+	defaultInfinispanPasswordKey = "password"
+)
+
+var (
+	deployDataIndexCmd      *cobra.Command
+	deployDataIndexCmdFlags = deployDataIndexFlags{
+		deployCommonFlags: deployCommonFlags{},
+		kafka:             v1alpha1.KafkaConnectionProperties{},
+		infinispan:        v1alpha1.InfinispanConnectionProperties{},
+	}
+)
+
+var _ = RegisterCommandVar(func() {
+	deployDataIndexCmd = &cobra.Command{
+		Use:     "deploy-data-index [flags]",
+		Short:   "Deploys the Kogito Data Index Service in the given Project",
+		Example: "deploy-data-index -p my-project --infinispan-url my-infinispan-server:11222 --kafka-url my-kafka-bootstrap:9092",
+		Long: `deploy-data-index will deploy the Data Index service to enable capturing and indexing data produced by one or more Kogito Runtime Services.
+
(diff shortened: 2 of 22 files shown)


cmd/kogito/deploy_data_index.go
```diff
@@ -0,0 +1,150 @@
...
+		Long: `deploy-data-index will deploy the Data Index service to enable capturing and indexing data produced by one or more Kogito Runtime Services.
+				For more information see: https://github.com/kiegroup/kogito-runtimes/wiki/Data-Index-Service`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return deployDataIndexExec(cmd, args)
+		},
+		PreRun:  preRunF,
+		PostRun: posRunF,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(deployDataIndexCmdFlags.infinispanUser) > 0 && len(deployDataIndexCmdFlags.infinispanPassword) == 0 {
+				return fmt.Errorf("Please set infinispan-user and infinispan-password")
```
sutaakar: I think we should provide more descriptive error message here, something like "infinispan-password wasn't provided, please set both infinispan-user and infinispan-password".
Current message doesn't explicitly say what is missing.

cmd/kogito/deploy_data_index.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+			if len(deployDataIndexCmdFlags.infinispanUser) == 0 && len(deployDataIndexCmdFlags.infinispanPassword) > 0 {
+				return fmt.Errorf("Please set infinispan-user and infinispan-password")
```
sutaakar: Same as above.

sutaakar: Also, I don't see a branch covering the case when both infinispan-user and infinispan-password are not set, is it missing or I overlooked something?

cmd/kogito/deploy_data_index.go
```diff
@@ -0,0 +1,150 @@
...
+		},
+	}
+})
+
+var _ = RegisterCommandInit(func() {
+	rootCmd.AddCommand(deployDataIndexCmd)
+	commonAddDeployFlags(deployDataIndexCmd, &deployDataIndexCmdFlags.deployCommonFlags)
+
+	deployDataIndexCmd.Flags().StringVarP(&deployDataIndexCmdFlags.image, "image", "i", resdataindex.DefaultImage, "Image tag (namespace/name:tag) for the runtime Service, e.g: openshift/kogito-data-index:latest")
+	deployDataIndexCmd.Flags().StringVar(&deployDataIndexCmdFlags.kafka.ServiceURI, "kafka-url", "", "The Kafka cluster internal URI, example: my-kafka-cluster:9092")
```
sutaakar: I think it may be good to mention which properties are mandatory and which are optional, so users don't have to find it out by themselves.

cmd/kogito/deploy_data_index.go
```diff
@@ -0,0 +1,150 @@
...
+func deployDataIndexExec(cmd *cobra.Command, args []string) error {
+	var err error
+	if deployDataIndexCmdFlags.project, err = ensureProject(deployDataIndexCmdFlags.project); err != nil {
+		return err
+	}
+	if err := checkKogitoDataIndexCRDExists(); err != nil {
+		return err
+	}
+
+	// if user and password is sent, let's create a secret to hold it and attach to the CRD
```
sutaakar: The comment mentions creating a secret, is it meant that secret is created on OpenShift or just as an object in the code? As from code below I cannot see any command actually creating the secret on OpenShift.

cmd/kogito/deploy_data_index.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(deployDataIndexCmdFlags.infinispanUser) > 0 && len(deployDataIndexCmdFlags.infinispanPassword) > 0 {
+		infinispanSecret := v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{Name: defaultInfinispanSecretName, Namespace: deployDataIndexCmdFlags.project},
+		}
+
+		if exist, err := kubernetes.ResourceC(kubeCli).Fetch(&infinispanSecret); err != nil {
+			return fmt.Errorf("Error while trying to fecth for the Infinispan Credentials Secret: %s", err)
```
sutaakar:
```suggestion
			return fmt.Errorf("Error while trying to fetch for the Infinispan Credentials Secret: %s", err)
```

cmd/kogito/deploy_data_index.go
ricardozanini: No credentials are used :)

cmd/kogito/resource_checks.go
```diff
@@ -10,9 +10,18 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	kogitoOperatorInstallationInstruc = "https://github.com/kiegroup/kogito-cloud-operator/blob/master/README.md#installation"
-)
+// ensureProject will check if the given project is a valid string in the context and if exists in the cluster
+func ensureProject(project string) (string, error) {
+	var err error
```
spolti: why are you setting this var here?

cmd/kogito/deploy_data_index.go
ricardozanini: You're completely right. I forgot to add a one liner to create the secret. Added that plus a unit test to check.

Many thanks!

cmd/kogito/command_checks.go
```diff
@@ -2,9 +2,9 @@ package main
...
 import "fmt"
 
-// checkProjecLocally will verify if the project/namespace exists in the CLI context.
+// checkProjectLocally will verify if the project/namespace exists in the CLI context.
 // Won't fecth the cluster to verify if the project/namespace exists. This is a local validation only.
-func checkProjecLocally(project string) (localProject string, err error) {
+func checkProjectLocally(project string) (localProject string, err error) {
 	if len(project) == 0 {
 		if len(config.Namespace) == 0 {
 			return "", fmt.Errorf("Couldn't find any Project in the current context. Use 'kogito use-project NAME' to set the Kogito Project where the service will be deployed or pass '--project NAME' flag to this one")
```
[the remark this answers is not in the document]
ricardozanini: We don't need that. The log implementation that we have on CLI is the same for the Operator, which is zapper.

Also, the log output is redirected to the cli, so users will see this error message clearly in the console.

This is nice because we can use `log.Debug` for `-v` flag, `log.Error` or `fmt.Errorf` for error messages and `log.Info` for general output.

cmd/kogito/resource_checks.go
ricardozanini: copy and paste from old code. fixed, thanks!!

cmd/kogito/command_checks.go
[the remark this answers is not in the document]
spolti: ok.

PR #61, opened 2019-09-17
[KOGITO-127] - Limiting s2i builds CPU and Memory resources

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-127

In this PR:

1. Fixed the resources used by s2i builds when deploying Kogito Runtime Services
2. Enhanced e2e tests to make it easy to verify build time

During my tests, the JVM builds (using `drools-quarkus-example` as benchmark) preformed really well with 1GB RAM and 0.5 CPUs.

For Native builds, things got really complicated and I had to raise memory to 4GB and CPU to 1 core. Unfortunately there isn't much to do since GraalVM is a resource killer.

Here's an illustration of the JVM build being limited:

### DIFF
--- a/deploy/crds/app_v1alpha1_kogitoapp_crd-0.4.0.yaml
+++ b/deploy/crds/app_v1alpha1_kogitoapp_crd-0.4.0.yaml
@@ -85,6 +85,45 @@ spec:
                     be compiled to run on native mode when Runtime is quarkus. See:
                     https://www.graalvm.org/docs/reference-manual/aot-compilation/'
                   type: boolean
+                resources:
+                  description: Resources for build pods. Default limits are 1GB RAM/0.5
+                    cpu on jvm and 4GB RAM/1 cpu for native builds.
+                  properties:
+                    limits:
+                      items:
+                        properties:
+                          resource:
+                            description: Resource type like cpu and memory
+                            enum:
+                            - cpu
+                            - memory
+                            type: string
+                          value:
+                            description: Value of this resource in Kubernetes format
+                            type: string
+                        required:
+                        - resource
+                        - value
+                        type: object
+                      type: array
+                    requests:
+                      items:
+                        properties:
+                          resource:
+                            description: Resource type like cpu and memory
+                            enum:
+                            - cpu
+                            - memory
+                            type: string
+                          value:
+                            description: Value of this resource in Kubernetes format
+                            type: string
+                        required:
+                        - resource
+                        - value
+                        type: object
+                      type: array
+                  type: object
                 webhooks:
                   description: WebHook secrets for build configs
                   items:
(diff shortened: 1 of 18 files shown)


ricardozanini:
Aside, I saw that we can limit the memory consumption in the GraalVM side as well. This could be useful. @spolti wdyt? 
https://stackoverflow.com/questions/57166853/quarkus-native-executable-build-high-memory-consumption

I guess we can set this parameter in the s2i images and also limit the build config as this PR is doing.

deploy/olm-catalog/kogito-cloud-operator/0.4.0/app_v1alpha1_kogitoapp_crd.yaml
```diff
@@ -85,6 +85,45 @@ spec:
                     be compiled to run on native mode when Runtime is quarkus. See:
                     https://www.graalvm.org/docs/reference-manual/aot-compilation/'
                   type: boolean
+                resources:
+                  description: Resources for build pods. Default limits are 1GB RAM/0.5
+                    cpu on jvm and 2GB RAM/0.8 cpu for native builds.
```
sutaakar: Can you please update the description to match current defaults for native build?

sutaakar:
```suggestion
                    cpu on jvm and 4GB RAM/1 cpu for native builds.
```

deploy/crds/app_v1alpha1_kogitoapp_crd-0.4.0.yaml
```diff
@@ -85,6 +85,45 @@ spec:
                     be compiled to run on native mode when Runtime is quarkus. See:
                     https://www.graalvm.org/docs/reference-manual/aot-compilation/'
                   type: boolean
+                resources:
+                  description: Resources for build pods. Default limits are 1GB RAM/0.5
+                    cpu on jvm and 2GB RAM/0.8 cpu for native builds.
```
sutaakar:
```suggestion
                    cpu on jvm and 4GB RAM/1 cpu for native builds.
```

spolti:
> sumption in the GraalVM side as well
Yes, I think we should do this on the graalvm side as well.

deploy/crds/app_v1alpha1_kogitoapp_crd-0.4.0.yaml
ricardozanini: @sutaakar thanks! I already changed in my local git. Will push this plus the env vars change soon

ricardozanini:
@spolti : https://issues.jboss.org/browse/KOGITO-306

ricardozanini:
@spolti @sutaakar @xiezhang7 mind to do a second look? Now the build pods will have new env vars `LIMIT_MEMORY` and `LIMIT_CPU`, so the image could use it to calculate the memory for graalvm builds.

PR #62, opened 2019-09-18
[KOGITO-192] Fix and code refactoring for kogito app CR update logic

xiezhang7:
Jira: https://issues.jboss.org/browse/KOGITO-192 https://issues.jboss.org/browse/KOGITO-107

Description: fix the CR update code


xiezhang7:
Hi @ricardozanini @bmozaffa @tchughesiv , could you please review this pull request? Thanks.

xiezhang7:
I'll create the unit tests for [manage_status.go](https://github.com/kiegroup/kogito-cloud-operator/blob/08df2ff172fb4f891bc02c9423c105bcc5851e3d/pkg/controller/kogitoapp/status/manage_status.go#L125) in a separate pull request later.

pkg/controller/kogitoapp/resource/manage_resources.go
```diff
@@ -0,0 +1,360 @@
...
+	return false, nil
+}
+
+func ensureServiceLabels(instance *v1alpha1.KogitoApp, serviceLabels map[string]string, log *zap.SugaredLogger) bool {
+	update := false
+
+	labels := map[string]string{}
+	labels[LabelKeyAppName] = instance.Name
+	if len(instance.Spec.Service.Labels) > 0 {
+
```
ricardozanini: Unneeded line break

pkg/controller/kogitoapp/resource/manage_resources.go
```diff
@@ -0,0 +1,360 @@
...
+		for key, value := range instance.Spec.Service.Labels {
+			labels[key] = value
+		}
+	} else {
+		labels[LabelKeyServiceName] = instance.Name
```
ricardozanini: This is the default label, right? WDYT having a func to add default service labels and then we reuse it here and when creating the service?

pkg/controller/kogitoapp/resource/manage_resources.go
```diff
@@ -0,0 +1,360 @@
...
+
+			if err := kubernetes.ResourceC(client).Update(buildConfigS2I); err != nil {
+				return updateS2I, err
+			}
+
+			if _, err := openshift.BuildConfigC(client).TriggerBuild(buildConfigS2I, instance.Name); err != nil {
+				return updateS2I, err
+			}
+
+			return updateS2I, nil
```
ricardozanini: In your tests, updating the BC instead of recreating it in case of updates was easier? I guess that we could safely delete the BC and recreate it in here since might have some sort of "unexpected" results if updated.

pkg/controller/kogitoapp/resource/manage_resources.go
```diff
@@ -0,0 +1,360 @@
...
+
+			if err := kubernetes.ResourceC(client).Update(buildConfigRuntime); err != nil {
+				return updateRuntime, err
+			}
+
+			if _, err := openshift.BuildConfigC(client).TriggerBuild(buildConfigRuntime, instance.Name); err != nil {
+				return updateRuntime, err
+			}
+
+			return updateRuntime, nil
```
ricardozanini: Same here.

pkg/controller/kogitoapp/resource/manage_resources.go
```diff
@@ -0,0 +1,360 @@
...
+
+func mapToEnv(env map[string]string) (envVars []corev1.EnvVar) {
+	for key, value := range env {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  key,
+			Value: value,
+		})
+	}
+
+	return envVars
```
ricardozanini: Could `envVarToMap` and `mapToEnv` functions be moved to `pkg/util/resources.go` file? I guess we have some functions there that are doing almost the same. Even an `env_var.go` file there would be useful.

docs/ARCHITECTURE.MD
```diff
@@ -6,13 +6,13 @@ The actual architecture has two [controllers](https://godoc.org/github.com/kuber
...
 
-We aim to avoid having a huge monolith inside the reconcile loop that does it all. With that in mind, we separated the responsibility of making the Kubernetes and OpenShift API calls to a package that we call [`client`](../pkg/client). Kubernetes/OpenShift resources that the controller need is defined and created inside the [`builder`](../pkg/controller/kogitoapp/builder) package. `Builder` communicates with the `client` package to bind or create the objects in the cluster. The `Controller` also make calls to `client` to perform certain taks during the `reconcile` loop.
+We aim to avoid having a huge monolith inside the reconcile loop that does it all. With that in mind, we separated the responsibility of making the Kubernetes and OpenShift API calls to a package that we call [`client`](../pkg/client). Kubernetes/OpenShift resources, which the controller needs, are defined and created inside the [`resource`](../pkg/controller/kogitoapp/resource) package. `resource` communicates with the `client` package to bind or create the objects in the cluster. The `controller` also make calls to `client` to perform certain tasks during the `reconcile` loop.
```
spolti: that does what? from user perspective it is not clear.

pkg/client/openshift/route.go
```diff
@@ -41,19 +41,19 @@ type route struct {
 	client *client.Client
 }
 
-func (r *route) GetHostFromRoute(routeKey types.NamespacedName) (string, error) {
+func (r *route) GetHostFromRoute(routeKey types.NamespacedName) (bool, string, error) {
 	route := &routev1.Route{}
 
 	for i := 1; i < 60; i++ {
```
spolti: why this for loop is needed?
is there a feasible way than a loop + sleep to do it?
If my understanding is correct the route should be returned right way, yes?

docs/ARCHITECTURE.MD
ricardozanini: Well, since I wrote this I believe I can answer: this is a small section to the contributors doc. This is intended to help others to understand the operator architecture before submitting PRs. It's not meant for final users, that's why it's in a separate doc.

pkg/client/openshift/route.go
bmozaffa: I haven't followed the full execution stack well enough to know how easy it would be, but: after creating the route, you could return with a Requeue of say, 300ms, to allow the route to likely be created and assigned a hostname. Then at this point, if no hostname is found, you could also just return with a shorter Requeue, e.g., 100ms, and let reconcile call again.

tchughesiv: @bmozaffa that's what we've done in the past in BA Operator, works well.
https://github.com/kiegroup/kie-cloud-operator/blob/0f35e826bf276d3f2d23fabbb266f21434bcae21/pkg/controller/kieapp/kieapp_controller.go#L104

xiezhang7:
> Outstanding work, @xiezhang7
> I can tell that you had a very hard time putting all this together, and I thank you for that.
> I just left some small comments.
> 
> One more thing, this would fix [KOGITO-107](https://issues.jboss.org/browse/KOGITO-107) as well, right?

Thanks, @ricardozanini . Yes, this also fixes KOGITO-107.

pkg/controller/kogitoapp/resource/manage_resources.go
xiezhang7: @ricardozanini , I my test, after updating the BC, I explicitly [start/trigger a new Build](https://github.com/kiegroup/kogito-cloud-operator/blob/08df2ff172fb4f891bc02c9423c105bcc5851e3d/pkg/controller/kogitoapp/resource/manage_resources.go#L269) with the BuildConfig. Otherwise, the Build won't run. We can keep track of all the previous Builds if we don't recreate the BC.

ricardozanini: Seems fair having the build history. This is what an administrator would expect. Let’s leave this way then and keep an eye on every possible modification that we could do on BC and update accordingly. 

I’m afraid we need a step to check if the BC is meant to run native builds or not, right?

pkg/controller/kogitoapp/resource/build_config_test.go
```diff
@@ -44,7 +44,7 @@ func Test_BuidConfig_NonNativeBuild(t *testing.T) {
 		},
 	}
 	bcS2I, _ := NewBuildConfigS2I(kogitoApp)
-	bcService, _ := NewBuildConfigService(kogitoApp, &bcS2I)
+	bcService, _ := NewBuildConfigRuntime(kogitoApp, &bcS2I)
```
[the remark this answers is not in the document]
xiezhang7: renamed

pkg/controller/kogitoapp/resource/manage_resources.go
xiezhang7: deleted

pkg/controller/kogitoapp/resource/manage_resources.go
xiezhang7: moved the function to [meta.go](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/controller/kogitoapp/resource/meta.go#L69)

pkg/controller/kogitoapp/resource/manage_resources.go
xiezhang7: moved the functions to [env_var.go](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/util/env_var.go#L23) and [envs.go](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/util/envs.go#L51)

pkg/controller/kogitoapp/resource/create_resources.go
```diff
@@ -140,19 +141,30 @@ func imageStreamBuilder(chain *builderChain) *builderChain {
 	return callPostCreate(created, is, chain)
 }
 
+func buildTrigger(chain *builderChain) *builderChain {
```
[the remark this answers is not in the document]
xiezhang7: Thanks @ricardozanini , I wanted to trigger the build after the creation of `ImageStream`s. So it cannot be added to the `postCreate` of `build_config_s2i`. And in the `postCreate` of the `ImagesStream`s, the information of the `BuildConfigS2I` is not available. So I moved the code to the [reconcile loop](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/controller/kogitoapp/kogitoapp_controller.go#L168) instead.

pkg/client/openshift/route.go
xiezhang7: Thanks @spolti @bmozaffa @tchughesiv . The sleep loop is [deleted](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/client/openshift/route.go#L45). If the route is not found, a [requeue](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/controller/kogitoapp/status/manage_status.go#L275) will be used.

pkg/controller/kogitoapp/resource/manage_resources.go
xiezhang7: @ricardozanini do you mean checking the BC after updating the CR to use native build? The [code](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/controller/kogitoapp/resource/manage_resources.go#L227) checks if it is configured to use native build.

ricardozanini: Yeah, right. But the native flag build will use other image to build it.

For example: if I have a JVM build set on the BCS2I and I change my mind and change it to native, what happens?

The update hook should update `.Spec.Build.Native=true` and also performs the image build update. We could separate this use case into a function that is called during the BC creation and by the update hook. WDYT?

xiezhang7: This part is moved to the function [resolveRuntimeImage](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/controller/kogitoapp/resource/build_config_runtime.go#L61) for the runtime BC creation. And the [update BC code](https://github.com/kiegroup/kogito-cloud-operator/blob/33e9918d0604a0593a609031c21facc071cb286f/pkg/controller/kogitoapp/resource/manage_resources.go#L295) calls this function to get the expected image, and compares this image with the image which is currently deployed. And also created the function [resolveS2IImage](https://github.com/kiegroup/kogito-cloud-operator/blob/aec718ab497fe4ab136f6a4a4faf69a57003c552/pkg/controller/kogitoapp/resource/build_config_s2i.go#L83) for the [s2i BC update](https://github.com/kiegroup/kogito-cloud-operator/blob/aec718ab497fe4ab136f6a4a4faf69a57003c552/pkg/controller/kogitoapp/resource/manage_resources.go#L203).

ricardozanini:
@spolti @bmozaffa mind to take a second look on this? We should merge after 0.4.0 release on Monday.

Many thanks!

bmozaffa:
> @spolti @bmozaffa mind to take a second look on this? We should merge after 0.4.0 release on Monday.
> 
> Many thanks!

@ricardozanini At a glance, this looks good. However, I also discussed with @xiezhang7 how our newer contributions to operator-utils may make full reconciliation easier. That doesn't necessarily have to hold this PR up, though.

ricardozanini:
Thanks @bmozaffa, I just opened https://issues.jboss.org/browse/KOGITO-322 to follow this up.

sutaakar:
I have noticed that deployment yamls were refactored as lists.
However, the List is not a valid resource on OpenShift 4. I understand that oc client is able to parse the List and create referenced resources separately. Still, I think it would be better to stick to available resources to allow creating objects from yaml files even manually.

xiezhang7:
> I have noticed that deployment yamls were refactored as lists.
> However, the List is not a valid resource on OpenShift 4. I understand that oc client is able to parse the List and create referenced resources separately. Still, I think it would be better to stick to available resources to allow creating objects from yaml files even manually.

I'll change it back.

ricardozanini:
@xiezhang7 I'll do this, since I'm using those yaml files to deploy the operator via CLI. Your chan ge broke my code, actually. :laughing:

PR #66, opened 2019-09-23
[KOGITO-178] Adding reference to the kogito-usertaskinstances-events …

ricardozanini:
…kafka topic

See: https://github.com/kiegroup/kogito-runtimes/pull/157#discussion_r325658169


spolti:
I didn't check the operator code, but did you also added the "kogito-processdomain-events" ?

ricardozanini:
> I didn't check the operator code, but did you also added the "kogito-processdomain-events" ?

We don't have any references in the code to the topics, since the Kafka installation should be handled manually by users. :(

We will add this feature in the future versions (to integrate with Kafka properly): https://issues.jboss.org/browse/KOGITO-246

PR #69, opened 2019-09-30
Bump to 0.5.0

ricardozanini:
Small changes to bump to the next version.

I'm about to make some changes in the code that will reflect in the 0.4.0 CSV, to avoid that we should bump and create a new CSV.

This should be in the CI after publishing the CSV to quay.


ricardozanini:
@xiezhang7  could you please take a look? Thanks!

PR #70, opened 2019-10-01
[KOGITO-160] - Add capability to create Kogito image streams in case …

spolti:
…they are missing

Signed-off-by: Filippe Spolti <fspolti@redhat.com>


pkg/controller/kogitoapp/resource/create_resources.go
```diff
@@ -140,6 +159,62 @@ func imageStreamBuilder(chain *builderChain) *builderChain {
...
+
+	for _, key := range ImageStreamNameList {
+		isTag, _ := ocp.FetchTag(types.NamespacedName{Name: key, Namespace: ImageStreamNamespace}, ImageStreamTag)
+		if isTag == nil {
+			need2create = append(need2create, key)
+		}
+	}
+
+	if len(need2create) > 0 {
+		imageStreamLocation := fmt.Sprintf(ImageStreamLocation, ImageStreamTag+"/1")
```
sutaakar: What is the purpose of /1 ?

spolti: none, garbage left, still wip

pkg/controller/kogitoapp/resource/create_resources.go
```diff
@@ -140,6 +159,62 @@ func imageStreamBuilder(chain *builderChain) *builderChain {
...
+			}
+
+			if err := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(bodyBytes), len([]byte(bodyBytes))).Decode(&imagestreamlist); err != nil {
+				log.Errorf("Error while unmarshal file '%s': %s ", resp.Body, err)
+			}
+
+			for _, is2create := range need2create {
+				for _, is := range imagestreamlist.Items {
+					if is2create == is.Name {
+						is.Namespace = ImageStreamNamespace
```
sutaakar: If I understand it correctly the operator will create image streams in openshift namespace. Does operator have privileges to do so?

pkg/controller/kogitoapp/resource/create_resources.go
```diff
@@ -140,6 +159,62 @@ func imageStreamBuilder(chain *builderChain) *builderChain {
...
+			need2create = append(need2create, key)
+		}
+	}
+
+	if len(need2create) > 0 {
+		imageStreamLocation := fmt.Sprintf(ImageStreamLocation, ImageStreamTag+"/1")
+		log.Infof("Using imagestream %s", imageStreamLocation)
+		resp, err := http.Get(imageStreamLocation)
+		if err != nil {
+			log.Errorf("Failed to fetch imagestream %s", err.Error())
```
sutaakar: In case error happens it is logged, however the method processing continue. Is it expected behaviour?

pkg/controller/kogitoapp/resource/create_resources.go
spolti: in this case yes.
I rewrote this logic, it will look, initially, for the image streams on the openshift and current namespace.
If a specific imagestream is not found on openshift and current namespace, then create it on the currenct namespace only.

pkg/controller/kogitoapp/resource/create_resources.go
spolti: not really.
for now it is returning the error for the reconciler, it will fail and no build will start.

ricardozanini:
I'm going to wait to review once the `WIP` label is gone. :)

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -220,3 +230,91 @@ func (r *ReconcileKogitoApp) setDefaultBuildLimits(instance *v1alpha1.KogitoApp)
...
+	// All images must be present, if not create it on current namespace
+	for _, key := range kogitores.ImageStreamNameList {
+		isOnOCPNamespace, _ := openshift.ImageStreamC(r.client).FetchTag(types.NamespacedName{Name: key, Namespace: kogitores.ImageStreamNamespace}, imageStreamTag)
+		if isOnOCPNamespace == nil {
+			need2createIS = true
+		}
+	}
+
+	if need2createIS {
+		imageStreamLocation := fmt.Sprintf(kogitores.ImageStreamLocation, imageStreamTag)
```
sutaakar: Image stream tag is expected to be always equal to GitHub branch?

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -220,3 +230,91 @@ func (r *ReconcileKogitoApp) setDefaultBuildLimits(instance *v1alpha1.KogitoApp)
...
+			return reconcile.Result{}, err
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode == http.StatusOK {
+
+			var imagestreamlist oimagev1.ImageStreamList
+			bodyBytes, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				log.Warnf("Failed to read response: %s", err.Error())
```
sutaakar: Wouldn't it have sense to return from method in case response body cannot be read? In this implementation only warning is shown and method execution continue with decoding of body content.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -220,3 +230,91 @@ func (r *ReconcileKogitoApp) setDefaultBuildLimits(instance *v1alpha1.KogitoApp)
...
+						}
+						if imageTag != nil && imageTag.Image.Name != "" {
+							log.Infof("Image %s ready.", imageTag.Name)
+							break
+						}
+						if counter >= 10 {
+							return reconcile.Result{}, fmt.Errorf("Failed to get information about image tag ref %s on namespace %s", tagRef, is.Namespace)
+						}
+						counter++
+						time.Sleep(500 * time.Microsecond)
```
sutaakar: Isn't 500 microseconds too small step? If I count correctly the loop will wait for 5 milliseconds to create tag, then it returns error.

pkg/controller/kogitoapp/resource/build_config.go
```diff
@@ -39,44 +39,50 @@ const (
 	tagLatest          = "latest"
 	// ImageStreamNamespace default namespace for the ImageStreams
 	ImageStreamNamespace = "openshift"
+	// ImageStreamLocation if no Kogito images found, use this to fetch the imagestream and import it on cluster under the default namespace
+	ImageStreamLocation = "https://raw.githubusercontent.com/kiegroup/kogito-cloud/%s/s2i/kogito-imagestream.yaml"
```
sutaakar: I am a bit worried of maintainability of these hardcoded values, specially if we will productize it.

ricardozanini: I'm also worried. Do you have a suggestion for this particular case? I mean, those should follow the tag/release from the `kogito-cloud` repository to not break old operators installations.

sutaakar: I think that we may use config files to define image names and tags, this way we can easily supply different values if needed. Also it will remove external dependency on GitHub (useful for cases when cluster is in offline environment).

ricardozanini: The removal of the Github reference is a must. I'd say that the Image Stream should be created like we do today with other resources, but not in the "resource creation fase", before.

[Also we already have an input for the image tags](https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/apis/app/v1alpha1/kogitoapp_types.go#L88-L90) that should be used as a reference to create the IS. If it's not given, we just use the default version which is the one from the operator itself.

pkg/controller/kogitoapp/kogitoapp_controller.go
spolti: not really, it wait for 5 seconds in total.
note that if the counter is greater or equal to 10 if returns a error.

pkg/controller/kogitoapp/kogitoapp_controller.go
spolti: +1, updated.

ricardozanini:
@spolti I believe you have to review this PR because of this update: https://github.com/kiegroup/kogito-cloud-operator/pull/72

mswiderski:
@spolti @ricardozanini  guys, can we make this merged? It will make the use of kogito significantly easier :)

ricardozanini:
@mswiderski unfortunately not yet, we need a lot of changes in this PR.
@spolti do you think you can work on this by the end of this sprint?

spolti:
I think I can.

pkg/controller/kogitoapp/resource/build_config.go
spolti: @ricardozanini @mswiderski @sutaakar ^

mswiderski: I’d say that given version of operator should support only given version of the images. Otherwise we could run out of sync

Another question - in what namespace will these image streams be installed? Openshift or default or?

spolti: @mswiderski the operator will first look for the imagestreams on the openshift namespace and in the namespace you are creating the kogito app.

If no imagestream are found, the operator will create it on the current namespace.

spolti: not really, IIRC the first place it will be looking is on the local namespace then openshift.
Only if all imagestreams are available on openshift namespace we set the Instance.Namespace to openshift.

@ricardozanini can confirm it.

spolti: @mswiderski yeah, right, that is the current behaviour. with this PR it will change: https://github.com/kiegroup/kogito-cloud-operator/pull/70/files#diff-4be0017824908f3c2c4610bbd4bfef96L55

by default it will first look on the current namespace.
Only if all image stream are available on openshift namespace, we set it manually.

ricardozanini: @mswiderski the default behavior is to look to openshift namespace, unless you specify the image details in the CLI via: `--image-s2i=mynamespace/kogito-quarkus-ubi8-s2i:latest`

You can run `kogito deploy --help` to see more details about this option.

Also, you can easily set this via CR: `.Spec.Build.ImageS2i`.

@spolti I guess you're considering those options too?

ricardozanini: @spolti looking at the code you linked above, why are you removing the ns?

spolti: @ricardozanini I think we spoke about it a few days ago, I removed it as part of this change.
only if all imagestreams are available on openshift namespace I set it manually here: https://github.com/kiegroup/kogito-cloud-operator/pull/70/files#diff-06053a5d4cb74bb9cf01187cb724f70dR423

this way, if the IS are created, it will look on the current namespace.

sutaakar: @spolti Ok for me to use GitHub IS as a temporary solution.

mswiderski: > @mswiderski the default behavior is to look to openshift namespace, unless you specify the image details in the CLI via: `--image-s2i=mynamespace/kogito-quarkus-ubi8-s2i:latest`

in that case we should always install image streams in openshift namespace as it is used by default and most users won't bother to use the `--image-s2i` argument

ricardozanini:
@spolti @sutaakar @mswiderski I guess we need a sum up for this PR to get merged:

1. The operator **will not** rely on Github to take the IS yaml file
2. The operator will create the IS in the **openshift** namespace if it's not there (please confirm if the operator user has the required permissions)
3. An user can use the CLI/CR to change where the images is located, e.g.:  `--image-s2i=mynamespace/kogito-quarkus-ubi8-s2i:latest`

Are we in the same page?

spolti:
1. yes
2. I don't think it is a good idea. and operator user don't have permissions.
3. yes.

ricardozanini:
>     2. I don't think it is a good idea. and operator user don't have permissions.

We can add this permission in the `role.yaml` file, for the operator user to have permissions to read and write ImageStream in the openshift namespace.

@sutaakar @xiezhang7 WDYT?

mswiderski:
all three points make sense to me.

About @spolti comment to not use openshift namespace - I am not against it as long as the operator upon builds will not reference openshift namespace by default. So what is the most important is the consistency and then the first time experience.

spolti:
About point 2, I think we should only grant permissions to operator user in cases there is no other option, in this case, we can rely on the current namespace, I don't think it is necessary.

ricardozanini:
So we have divergent opinions here:

1. The default namespace for images is `openshift`, but we don't have the required permissions to create them there (we could have, but might be too over regarding security issues)
2. We could just create the images in the operator namespace, but on the other hand will divert from the default path

I'd say to go with #2.

One last thing, we should have the option to let users to set where their IS are. This is a very common use cases for CI where the IS is on a different namespace.

sutaakar:
#2 sounds good to me.

pkg/controller/kogitoapp/resource/build_config.go
spolti: Github reference removed.

pkg/client/openshift/imagestream.go
```diff
@@ -107,3 +108,17 @@ func (i *imageStream) CreateTagIfNotExists(is *imgv1.ImageStreamTag) (bool, erro
 	log.Debugf("Image Stream Tag %s created in namespace %s", is.Name, is.Namespace)
 	return true, nil
 }
+
+// CreateTagIfNotExists will create a new ImageStream if not exists
```
ricardozanini: There's a typo in the documentation

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +334,79 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
+		}
+	}
+
+	// look for the Kogito image stream on the default namespace (openshift)
+	// All images must be present, if not create it on current namespace
+	if !r.isImagestreamReady {
+		for _, key := range kogitores.ImageStreamNameList {
+			isOcpNs, _ := openshift.ImageStreamC(r.client).FetchTag(types.NamespacedName{Name: key, Namespace: kogitores.ImageStreamNamespace}, imageStreamTag)
+			if isOcpNs == nil {
+				isISPresentOnOpenshiftNamespace = false
```
ricardozanini: Have you tested this condition in an actual cluster? The operator user has the permissions to fetch `is` from `openshift` namespace?

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +334,79 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
+					}
+					if imageTag != nil && imageTag.Image.Name != "" {
+						log.Infof("Image %s ready.", imageTag.Name)
+						break
+					}
+					if counter >= 10 {
+						return reconcile.Result{}, fmt.Errorf("Failed to get information about image tag ref %s on namespace %s", tagRef, is.Namespace)
+					}
+					counter++
+					time.Sleep(500 * time.Millisecond)
```
ricardozanini: Maybe `time.Sleep(5 * time.Second)`?

pkg/controller/kogitoapp/resource/build_config.go
```diff
@@ -44,39 +44,43 @@ const (
...
 )
 
 // BuildImageStreams are the image streams needed to perform the initial builds
 var BuildImageStreams = map[BuildType]map[v1alpha1.RuntimeType]v1alpha1.Image{
 	BuildTypeS2I: {
 		v1alpha1.QuarkusRuntimeType: v1alpha1.Image{
-			ImageStreamName:      "kogito-quarkus-ubi8-s2i",
-			ImageStreamNamespace: ImageStreamNamespace,
-			ImageStreamTag:       ImageStreamTag,
+			ImageStreamName: "kogito-quarkus-ubi8-s2i",
```
ricardozanini: Instead of duplicating the imageStreamName we can have `const`s and then reference them here and in the splice  `ImageStreamNameList`

pkg/controller/kogitoapp/resource/kogito_imagestream.go
```diff
@@ -0,0 +1,143 @@
...
+
+		tagAnnotations := make(map[string]string)
+		tagAnnotations["iconClass"] = "icon-jbpm"
+		tagAnnotations["version"] = targetTag
+
+		imageStreamAnnotations := make(map[string]string)
+		imageStreamAnnotations["openshift.io/provider-display-name"] = "Kie Group."
+
+		switch imageName {
+		case "kogito-quarkus-ubi8":
```
ricardozanini: We could use the `const`s mentioned before here too.

pkg/controller/kogitoapp/kogitoapp_controller.go
spolti: Tested on my local cluster running with operator-sdk.
worth a test using olm as well.

pkg/controller/kogitoapp/kogitoapp_controller.go
spolti: if we change to 5*seconds, the for loop will wait for 50 seconds in total if no IS is created and 5 seconds for every try.
with 0.5s it will wait 5 seconds tops before thrown an error and 0.5 seconds between every try.

pkg/controller/kogitoapp/resource/build_config.go
spolti: do you mean create constants for the imagestream names, i.e.:

quarkusS2i = "kogito-quarkus-s2i"

?

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: Oh ok, you're right I misread the parameter.

pkg/controller/kogitoapp/resource/build_config.go
ricardozanini: Yes

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: Should work, but since now that we have our own env, do you mind testing there?

ricardozanini: Yes, we have the required permissions: https://github.com/kiegroup/kogito-cloud-operator/blob/master/deploy/role.yaml#L11

```yaml
- apiGroups:
      - ""
      - app.kiegroup.org
      - apps.openshift.io
      - image.openshift.io
      - build.openshift.io
      - rbac.authorization.k8s.io
      - route.openshift.io
    resources:
      - "*"
    verbs:
      - "*"
```

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -125,9 +127,10 @@ var _ reconcile.Reconciler = &ReconcileKogitoApp{}
 type ReconcileKogitoApp struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client *kogitocli.Client
-	scheme *runtime.Scheme
-	cache  cachev1.Cache
+	client             *kogitocli.Client
+	scheme             *runtime.Scheme
+	cache              cachev1.Cache
+	isImagestreamReady bool
```
xiezhang7: It seems this variable is only used in function `ensureKogitoImageStream`, it could be just a local variable.

xiezhang7: Sorry, I didn't notice the comments `set the isImagestreamReady to true to avoid it to be checked every time the reconcile runs`.

But I don't think we want to maintain any state by the `Reconciler`. [It should re-evaluate the entire state as opposed to just what changed](https://blog.openshift.com/kubernetes-operators-best-practices/). Otherwise, we will have to figure out how to sync the states between the `Reconciler` and the cluster. 

If you think the checking is a waste in most cases, maybe we can just check the single image stream used by this particular kogito app and create only that single image stream. Because in many cases, only one image stream will used by one kogito app. It is not very helpful to create all the image streams.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +334,79 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
+			}
+
+			if hasIs == nil {
+				r.isImagestreamReady = false
+				_, err := openshift.ImageStreamC(r.client).CreateImageStream(&is)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+
+				// wait for 5 seconds to the image get ready
```
xiezhang7: Instead of waiting for the image stream to create, we usually create the resources and  requeue the reconcile loop for a period of time, like this [line](https://github.com/kiegroup/kogito-cloud-operator/blob/31d5dc655e1ab86bf42a5b328b4a8ecda1356ddd/pkg/controller/kogitoapp/kogitoapp_controller.go#L251) . The reconcile loop will be triggered again after the specified period of time.

pkg/controller/kogitoapp/kogitoapp_controller.go
xiezhang7: With `requeue`, we don't need to store the state of the image stream, it will be checked in every iteration of the loop. 

In case the image stream is deleted, it will be created again.

ricardozanini: I liked the idea of creating just the image**s** that would be used by this deployment.
And also, wait for the object to be create is a best practice indeed, thanks for pointing this out @xiezhang7 

Instead, create the ImageStream and return a Reconcile Request after 5/10 seconds and them your algorithm will check for the existence of the IS. If it's not there, reconcile again.

Always with a message to the user: "ImageStream not ready/found, scheduling reconcile".

Thanks!

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: Just read your comment after writing mine :(

spolti: +1
I liked this idea, now this is the behaviour.

pkg/controller/kogitoapp/kogitoapp_controller.go
spolti: done, thanks.

deploy/olm-catalog/kogito-cloud-operator/0.6.0/kogito-cloud-operator.v0.6.0.clusterserviceversion.yaml
```diff
@@ -60,87 +60,87 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
     owned:
```
ricardozanini: There's a lot of changes in this file. Have you upgraded to 0.11 SDK?

pkg/apis/app/v1alpha1/zz_generated.openapi.go
```diff
@@ -50,7 +50,6 @@ func schema_pkg_apis_app_v1alpha1_Builds(ref common.ReferenceCallback) common.Op
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
 				Description: "Builds ...",
-				Type:        []string{"object"},
```
ricardozanini: Same question here. This file shouldn't be changed. If it's, ok, but just double check if you upgrade to 0.11.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +332,67 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
+
+		for _, is := range kogitores.KogitoImageStream(instance.Namespace, imageStreamTag, instance.Spec.Runtime, instance.Spec.Build.Native).Items {
+			is.Namespace = instance.Namespace
+			hasIs, err := openshift.ImageStreamC(r.client).FetchTag(types.NamespacedName{Name: is.Name, Namespace: instance.Namespace}, imageStreamTag)
+			if err != nil {
+				log.Error(err.Error())
+			}
+
+			if hasIs == nil {
+				_, err := openshift.ImageStreamC(r.client).CreateImageStream(&is)
```
ricardozanini: When the second reconcile loop reaches here, there's an error reporting that the IS is already created? Or we could just fetch first and create after.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +332,67 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
 	return k8sRes
 }
+
+// Ensure that all Kogito images are available before the build.
+func (r *ReconcileKogitoApp) ensureKogitoImageStream(instance *v1alpha1.KogitoApp) (reconcile.Result, error) {
+	isISPresentOnOpenshiftNamespace := true
+	isImagestreamReady := true
+
+	imageStreamTag := kogitores.ImageStreamTag
+	if instance.Spec.Build.ImageS2I.ImageStreamTag != "" {
```
ricardozanini: Where's the check for image stream in the Runtime Build?

ricardozanini:
Looks good, minor comments. I guess that the reconcile handling is a much nicer and cleaner way of doing things. Thanks for the hard work!

pkg/apis/app/v1alpha1/zz_generated.openapi.go
spolti: I didn't, was using sdk version 0.10

deploy/olm-catalog/kogito-cloud-operator/0.6.0/kogito-cloud-operator.v0.6.0.clusterserviceversion.yaml
spolti: updated.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +332,67 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
+		imageStreamTag = instance.Spec.Build.ImageS2I.ImageStreamTag
+		log.Infof("Using ImageS2I Tag from CR: %s", imageStreamTag)
+	}
+
+	// First look on the current namespace, if the imagestreams are available, return
+	// and don't check for anything else.
+	// look for the Kogito image stream on the default namespace (openshift)
+	// All needed images must be present, if not create it on current namespace
+	for _, key := range kogitores.ImageStreamNameList {
+		is, _ := openshift.ImageStreamC(r.client).FetchTag(types.NamespacedName{Name: key, Namespace: instance.Namespace}, imageStreamTag)
```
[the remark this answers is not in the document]
spolti: The point to check both is that, the next section will be called only if the IS are not available on openshift or current namespace.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -327,3 +332,67 @@ func getKubernetesResources(kogitoRes *kogitores.KogitoAppResources) []utilsres.
...
+				}
+
+				tagRef := fmt.Sprintf("%s:%s", is.Name, imageStreamTag)
+				imageTag, getISerr := openshift.ImageStreamC(r.client).FetchTag(types.NamespacedName{Name: is.Name, Namespace: is.Namespace}, imageStreamTag)
+				if getISerr != nil {
+					log.Error(getISerr)
+				}
+				if imageTag != nil && imageTag.Image.Name != "" {
+					log.Infof("Image %s ready.", imageTag.Name)
+					break
```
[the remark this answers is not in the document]
spolti: Agreed with break part, but the second check is to make sure the image is available. If all the imageStream is available, this second step would not be reached.

spolti:
@xiezhang7 thanks for your tips, I think I addressed all your concerns
@ricardozanini 
do you guys mind to review it again?

deploy/olm-catalog/kogito-cloud-operator/0.6.0/kogito-cloud-operator.v0.6.0.clusterserviceversion.yaml
```diff
@@ -4,21 +4,6 @@ metadata:
   annotations:
     alm-examples: |-
       [
-        {
```
ricardozanini: I guess this is just a matter of formatting the yaml, but please just double check that your SDK is set to 0.11.0

PR #71, opened 2019-10-03
Small documentation review for OperatorHub

ricardozanini:
Small documentation review (organizational, more info)

@sterobin could you please take a look?


sterobin:
@ricardozanini For the sake of time, maybe you should exclude my review and just merge, since it's community-only anyway, and then I'll plan on reviewing the README overall later anyway. I'm on PTO until Monday and even then must prioritize other things first. So rather than make you wait, I thought you could just merge as is. It's up to you.

PR #74, opened 2019-10-07
[KOGITO-376] - Removing 'LastSavedKogitoApp' to avoid marshall problems

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-376

The problem here was a problem with an unmarshal feature from Viper when handling timestamps. Since we are not using the "saved" KogitoApp in the config file, I removed. Once we need a cache from the created CR, we can then handle the unmarshal problem and also only save the data that will be needed.

Also, I panic'ed all errors regarding manipulating the configuration file since it's being write/read in the user's home. If we got an error there, the command flow should interrupt right away to not spawn the bug to other features.

Now let's see the CI behavior.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,5 @@ jobs:
       - run: go mod vendor
       - run: go fmt ./...
       - run: go vet ./...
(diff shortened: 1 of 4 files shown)


ricardozanini:
It's working :smile_cat:

PR #82, opened 2019-10-15
[KOGITO-414] - Saving project in use every time if -p flag is used

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-414

In this PR we introduce a fix/feature that persists the `--project` flag on each command in the user's machine. This means that all subsequent commands will use the persisted project name if the flag is absent.

Plus, this should fix the intermittent unit test error regarding having or not the project in:
https://github.com/kiegroup/kogito-cloud-operator/blob/v0.5.0/cmd/kogito/command/install/kogito_operator_test.go#L39
 

ricardozanini:
This annoying error that only happens on CI is freaking me out. Will work on it tomorrow.

ricardozanini:
Looks like there’s a race condition to access the config file, that’s why sometimes the test just works.

ricardozanini:
Solved by isolating the `cmd` test cases to have each test access the file at a time. There was a lot of process accessing the same file, writing different namespaces in there. Some tests need to write and then check at the end, this caused the tests to not have a defined behavior, failing some times.

ricardozanini:
@xiezhang7 @spolti please take a look when you have the time?

PR #83, opened 2019-10-17
[KOGITO-287] - Injecting proto files into DataIndex ConfigMap

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-287

In this PR we added the possibility to add protobuf files from image labels `org.ki/persistence/protobuf` to the Data Index configMap.

The controller will now listen to the `BuildConfig` objects created by `KogitoApps` each time the reconcile occur, the kogitoApp image is read, the file is extracted and added to the ConfigMap.

Doing this way we can remove, update or add those files from kogitoApps within the namespace.


ricardozanini:
Will only work on OCP 3.11 because [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1758305) prevents the protobuf files to be added to our images.

pkg/resource/main.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resource
+
+import "github.com/kiegroup/kogito-cloud-operator/pkg/logger"
+
+var log = logger.GetLogger("resource_api")
```
spolti: why call it main? shouldn't be something related with log?

pkg/util/arrays.go
```diff
@@ -15,12 +15,26 @@
...
+	"bytes"
+	"crypto/md5"
 	"fmt"
 	"strings"
 )
 
 const keyPairSeparator = "="
 
+// GenerateMD5Hash will generate a MD5 hash from the given map
+func GenerateMD5Hash(source map[string]string) string {
```
spolti: what is the purpose of this?

pkg/resource/configmap_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{name: "config map has no values", args: args{
+			files: files,
+			cm:    emptyCm,
+		}},
+		{name: "config map has values", args: args{
```
sutaakar: What is the purpose of this test step? If I understand it correctly then we take a CM with data, calling AddFilesToConfigMap with this CM as argument and same data as other argument, in the end checking that the data are still there.Is it so?

pkg/util/arrays_test.go
```diff
@@ -123,3 +124,30 @@ func TestContainsAll(t *testing.T) {
...
+		source map[string]string
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
+		{"Should have the same hash as map1", args{map2}, hashToCompare},
+		{"Should have no hash", args{map[string]string{}}, ""},
+		{"Should have no hash because nil", args{nil}, ""},
```
sutaakar: Please add a test case making sure that maps with different contents should have different hash.

pkg/controller/kogitoapp/resource/route.go
```diff
@@ -33,7 +33,7 @@ func NewRoute(kogitoApp *v1alpha1.KogitoApp, service *corev1.Service) (route *ro
 		ObjectMeta: *service.ObjectMeta.DeepCopy(),
 		Spec: routev1.RouteSpec{
 			Port: &routev1.RoutePort{
-				TargetPort: intstr.FromString(defaultExportedProtocol),
+				TargetPort: intstr.FromString(resource.DefaultExportedProtocol),
```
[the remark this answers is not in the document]
ricardozanini: Yeah..I guess

pkg/resource/main.go
ricardozanini: This is just a convention, like the entry point for the package.

pkg/util/arrays.go
ricardozanini: To compare the content between the image protobuf files and the configMap files:
https://github.com/kiegroup/kogito-cloud-operator/pull/83/files#diff-a35a3b1905a737e2fc36cb08fac9a195R70

pkg/resource/configmap_test.go
ricardozanini: This is a leftover of my TDD. I tend to test the interface first for basic functionality and then the implementation.

pkg/util/arrays_test.go
ricardozanini: Done

pkg/resource/configmap_test.go
ricardozanini: Replying direct your question: yes, it's that. Because the function should override same values, and in the other use case we verify if the files are being added to the CM.

ricardozanini:
@spolti @sutaakar could you guys please take another look in this new commit?

Thanks!

pkg/util/arrays.go
xiezhang7: I would suggest moving the function to a separate file, e.g. string.go or hash.go

ricardozanini:
@sutaakar are we ok to go?

pkg/resource/configmap_test.go
sutaakar: ok, thanks

sutaakar:
@ricardozanini  Yes, good for me.

PR #86, opened 2019-10-22
[KOGITO-478] Add possibility to run e2e tests with external image 

sutaakar:
Set operator image tag based on tag parameter

https://issues.jboss.org/browse/KOGITO-478

### DIFF
--- a/hack/run-e2e.sh
+++ b/hack/run-e2e.sh
@@ -19,27 +19,39 @@ namespace=$1
 tag=$2
 native=$3
 maven_mirror=$4
+image=$5
 
 if [ -z "$namespace" ]; then
   echo "Please inform the namespace where the tests will run"
   exit 1
 fi
 
+# creates the namespace
+oc create namespace ${namespace}
+
 if [ -z "$tag" ]; then
   echo "-------- tag is empty, assuming default from the Operator"
 fi
 
-echo "-------- Running e2e tests with namespace=${namespace}, tag=${tag}, native=${native} and maven_mirror=${maven_mirror}"
+if [ -n "$image" ]; then
+  echo "-------- image is set, using Kogito operator image ${image}"
+  # operator permissions are created by operator-sdk
+  E2E_PARAMS="${E2E_PARAMS} --image ${image}"
+else
+  echo "-------- using local operator code for testing"
 
-# creates the namespace
-oc create namespace ${namespace}
-# gives permissions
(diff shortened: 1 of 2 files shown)


sutaakar:
Still thinking about this a bit.
What should be the usage of these e2e tests?
Should they support execution of local implementation (tests would have to run sequentially, tests will have to take care of cleaning up the namespace in the end) or rather focus more or executing tests against specific existing images (allowing parallel test runs in separate namespaces)?
Or both?

ricardozanini:
@sutaakar thanks for your time having a look at this. Let me try to reply your questions.

> What should be the usage of these e2e tests?
For now, I'm using it as a final assessment before submiting a PR. In the future I intend to have those to be run in CI, so we could automate at least basic usage if this tests.

> Should they support execution of local implementation (tests would have to run sequentially, tests will have to take care of cleaning up the namespace in the end) or rather focus more or executing tests against specific existing images (allowing parallel test runs in separate namespaces)?
Or both?
I'd say that we can have those tests to check basic use cases in complex environments. This means deploying a native and non-native Kogito service on OCP 3.11, 4.x and Kubernetes, for example. With persitence and without. With data index and without and so on.

hack/run-e2e.sh
```diff
@@ -27,19 +27,17 @@ fi
...
   echo "-------- tag is empty, assuming default from the Operator"
+else
+  E2E_PARAMS="${E2E_PARAMS} --image quay.io/kiegroup/kogito-cloud-operator:${tag}"
 fi
 
 echo "-------- Running e2e tests with namespace=${namespace}, tag=${tag}, native=${native} and maven_mirror=${maven_mirror}"
 
 # creates the namespace
 oc create namespace ${namespace}
-# gives permissions
```
ricardozanini: Why did you remove this? Without the required permissions the operator won't run correctly.

sutaakar: Originally I thought that these permissions are handled by operator-sdk when running the tests. But as I check the documentation, the role.yaml and role_binding.yaml are not applied. I will return it back.

sutaakar:
Adjusted implementation.
Now it is possible to run tests against specific docker image or using local operator code.
In case of docker image the service accounts and roles are created by operator-sdk.
Local operator code functionality stays as it is.

ricardozanini:
@sutaakar could you please open a JIRA to track this change?

sutaakar:
Adjusted commit message with JIRA reference

PR #94, opened 2019-10-24
[KOGITO-341] - Checking if the Kogito Operator is in the hub before r…

ricardozanini:
…unning install

See:
https://issues.jboss.org/browse/KOGITO-341

If running the CLI in the OpenShift with OLM installed, the user will see:

```bash
kogito install operator -p kogito
Error: Kogito Operator is available in the OperatorHub, please install it via Web Console, see: https://github.com/kiegroup/kogito-cloud-operator#via-operatorhub-automatically 
Usage:
  kogito install operator [flags]

Examples:
install operator -p my-project

Flags:
  -h, --help             help for operator
  -i, --image string     The operator image (default "quay.io/kiegroup/kogito-cloud-operator:0.6.0")
  -p, --project string   The project name where the operator will be deployed

Global Flags:
      --config string   config file (default is $HOME/.kogito/config.yaml)
  -v, --verbose         verbose output
      --version         display version

Kogito Operator is available in the OperatorHub, please install it via Web Console, see: https://github.com/kiegroup/kogito-cloud-operator#via-operatorhub-automatically 
```

The CLI won't create and Operator Subscriptions since this task is managed by the Operator Marketplace. We will probably mess with things plus every update from the market place could cause the CLI to stop working. Thoughts?


ricardozanini:
> Seems good to me. One question though:
> Does it also happen when launching command `kogito new-project` or `kogito use-project` ?

Yes, the verification of the operator installation happens in those commands. Manly on `new-project` because we will install the whole infra there in the future. The first piece is the operator.

radtriste:
That would mean that each time the user creates a project via `new-project` and Operator is in the OperatorHub, the user would get an error ?

radtriste:
IMO, the automatic installation of Operator when using `new-project` or `use-project` should be optional (as a config parameter of the command or in the config file) and not forced.
Then, if the user try to do an install or deployment, then we should check for operator installed or not.
And if not installed, then
- If in OperatorHub then say that it should be installed by this way
- If not, then install it

Also, if in OperatorHub, the user should have the choice to force the installation via the CLI.

wdyt ? @ricardozanini

ricardozanini:
> That would mean that each time the user creates a project via new-project and Operator is in the OperatorHub, the user would get an error ?

Very good catch! Yes, `new-project` would raise an error. I'm about to push a new commit that will only warn the user and then proceed with the new project creation.

> IMO, the automatic installation of Operator when using new-project or use-project should be optional (as a config parameter of the command or in the config file) and not forced.

For `new-project` the CLI will emit a warn message that the operator is available in the OperatorHub and it should be installed by he/she. If it's not, the operator will be installed via the `yaml` files method like it's today. The operator MUST be installed anyway, since the CLI is just a shortcut for `oc create -f kogitoapp_cr.yaml`. There's no option. We're just skipping the operator installation for the `new-project` because otherwise the namespace won't be created whatsoever. Hence any other forthcoming operation will require the operator to be installed.

For `use-project` we'll force the installation based on what I explained above.

> If in OperatorHub then say that it should be installed by this way; If not, then install it

That's what we're doing.

> Also, if in OperatorHub, the user should have the choice to force the installation via the CLI.

I was planning to have this use case, mainly for us testers to force an installation using other image like `kogito install -f -i mycustomoperatorimage -p namespace`. Could you kindly open a JIRA for this one?

Many thanks for your review, much appreciated!

ricardozanini:
Here's the result of the `new-project` call:

```bash
kogito new-project kogito2
Project 'kogito2' created successfully
Kogito Operator is available in the OperatorHub, please install it via Web Console. See: https://github.com/kiegroup/kogito-cloud-operator#via-operatorhub-automatically 
```

radtriste:
ok for me with the warning message.
Missed a bit the definition of the kogito CLI but now it is more clear :)

> > Also, if in OperatorHub, the user should have the choice to force the installation via the CLI.
> 
> I was planning to have this use case, mainly for us testers to force an installation using other image like `kogito install -f -i mycustomoperatorimage -p namespace`. Could you kindly open a JIRA for this one?

Yes will do !

cmd/kogito/command/shared/install_operator.go
```diff
@@ -74,7 +93,11 @@ func InstallOperatorIfNotExists(namespace string, operatorImage string, client *
 		return nil
 	}
 
-	//TODO: check if we have operator available at the hub, warn the user and exit
+	if available, err := isOperatorAvailableInOperatorHub(client); err != nil {
+		log.Warnf("Couldn't find if the Kogito Operator is available in the cluster: %s ", err)
+	} else if available {
```
spolti: why return a error in case of success?

ricardozanini: This could be an security error like "user doesn't have permission to fetch this resource". It's valueable info to the end user to understand why we failed to fetch for the operator hub.

Then we continue to install the operator using the standard way.

PR #95, opened 2019-10-25
KOGITO-482: Kogito CLI tests are changing the kogito config file

radtriste:
https://issues.jboss.org/browse/KOGITO-482

Changed tests to use a different config file
To avoid conflicts with current machine configuration

### DIFF
--- a/cmd/kogito/command/context/config.go
+++ b/cmd/kogito/command/context/config.go
@@ -21,6 +21,7 @@ import (
 	"gopkg.in/yaml.v2"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/viper"
 )
@@ -46,7 +47,8 @@ type Configuration struct {
 func InitConfig() {
 	if rootCmd.ConfigFile() != "" {
 		// Use config file from the flag.
-		viper.SetConfigFile(rootCmd.ConfigFile())
+		fullPath, file := filepath.Split(rootCmd.ConfigFile())
+		updateConfigFile(fullPath, file)
 	} else {
 		// Find home directory.
 		home, err := homedir.Dir()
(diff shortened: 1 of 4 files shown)


ricardozanini:
@radtriste you can just leave the checkbox unchecked when writing the PR description and then just check them using the UI. :stuck_out_tongue:

radtriste:
ahah. Godd to know. Thx !

radtriste:
@ricardozanini You can begin the review ;)

cmd/kogito/command/test/common.go
```diff
@@ -18,10 +18,12 @@ import (
 	"bytes"
 	"github.com/spf13/cobra"
 	"strings"
-
+	"path/filepath"
+	
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/meta"
+	"github.com/mitchellh/go-homedir"
```
ricardozanini: Since we're working on this, can you please change the implementation so we reuse the util package to get the home dir? Please see: https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/util/envs.go#L43

I'm not against using this dependency, but would be nice to trim a little bit OR use it in the `util` package instead.

cmd/kogito/command/test/common.go
```diff
@@ -33,6 +35,14 @@ var (
 	rootCommand *cobra.Command
 )
 
+const (
+	// Test constants for config file, based on context ones
+	DefaultConfigPath = context.DefaultConfigPath
```
ricardozanini: We really need to export those `const`s? If not, make them private please or add a comment to it so the lint tool won't make noise. Run the `make test` command in the root directory and observe the lint fails for more info.

cmd/kogito/command/test/common.go
```diff
@@ -74,3 +85,24 @@ func ExecuteCli() (string, string, error) {
 
 	return testOut.String(), testErr.String(), err
 }
+
+func InitConfigWithTestConfigFile(){
```
ricardozanini: Please add docs to the exported function

cmd/kogito/command/test/common.go
```diff
@@ -74,3 +85,24 @@ func ExecuteCli() (string, string, error) {
...
+	SetupCliTest("", context.CommandFactory{BuildCommands: buildCommands})
+	context.InitConfig()
+}
+
+func buildCommands(ctx *context.CommandContext, rootCommand *cobra.Command) []context.KogitoCommand {
+	return []context.KogitoCommand{}
+}
+
+func GetTestConfigFilePath() string {
```
ricardozanini: Docs

cmd/kogito/command/test/common.go
```diff
@@ -74,3 +85,24 @@ func ExecuteCli() (string, string, error) {
...
+	return filepath.Join(GetTestConfigPath(), DefaultConfigFinalName)
+}
+
+func GetTestConfigPath() string {
```
ricardozanini: Docs

cmd/kogito/command/test/common.go
```diff
@@ -51,6 +61,7 @@ func SetupCliTest(cli string, factory context.CommandFactory, kubeObjects ...run
 	kogitoRootCmd.Command().SetArgs(strings.Split(cli, " "))
 	kogitoRootCmd.Command().SetOut(testOut)
 	kogitoRootCmd.Command().SetErr(testErr)
+	kogitoRootCmd.Command().ParseFlags([]string{"--config", GetTestConfigFilePath()})
```
ricardozanini: Nice that this is working. I would've join this args to the `cli` command `string` instead. Don't need to change, just an observation. =)

Actually is a better solution since the test could've send `--config` args and here we are overriding it.

cmd/kogito/command/context/config.go
```diff
@@ -46,36 +47,44 @@ type Configuration struct {
...
-						panic(fmt.Errorf("Error while trying to write config file: %s ", err))
-					}
-				}
+		updateConfigFile(fullPath, DefaultConfigFinalName)
+	}
+
+	viper.AutomaticEnv() // read in environment variables that match
+}
+
+func updateConfigFile(fullPath, file string) {
```
ricardozanini: Have you tested the built CLI with the `--config` flag? I mean something like `kogito use-project kogito --config /tmp/myfile.yaml`?

cmd/kogito/command/test/common.go
radtriste: updated

cmd/kogito/command/test/common.go
radtriste: updated

cmd/kogito/command/test/common.go
radtriste: updated

cmd/kogito/command/test/common.go
radtriste: updated

cmd/kogito/command/context/config.go
radtriste: tested and approved, with and without the --config flag. Config file is correctly used :)

cmd/kogito/command/test/common.go
radtriste: updated

cmd/kogito/command/test/common.go
radtriste: updated

cmd/kogito/command/test/common.go
ricardozanini: Oh my bad, I meant that **your solution** is better! Please keep yours! I'm sorry, reading the comment again is really open for misinterpretation.

radtriste: ok, I will set back the original ;)

radtriste: done

ricardozanini:
@radtriste could you please squash your commits? Many thanks!

radtriste:
squash done !

PR #96, opened 2019-10-28
[KOGITO-493] - Adding CreateOptions to CLI jube client plus automate …

ricardozanini:
…smoke test for CLI

See:
https://issues.jboss.org/browse/KOGITO-493

In this PR, we added the missing Schema for the `CreateOptions` type within the Operator namespaces.

Plus, to avoid errors like this in the future, I added a simple `make` task to run a smoke test with the CLI. Please run this test whenever sending a PR while we don't have a proper automated CI.

@sutaakar @radtriste please take a look into the smoke test for the CLI. This is not a rocket science like you guys are used to, but should at least cover the basics to avoid running into similar issues in the future.

@xiezhang7 could you take a look please? In 0.11.0 they introduced the `CreateOptions` type in the client that requires one more schema mapper.


radtriste:
> Looks good to me.
> For the future I was thinking about making CLI tests part of current e2e test suite, @ricardozanini what do you think about it?

Would be a good idea. But need to check how it can be done internally, as part of the e2e tests...
We also need to decide what we want to test in there. The final CLI or the functionality inside ?
I would say that we need to test the final CLI in an environment and check everything is ok.

We should continue the dicussion here: https://issues.jboss.org/browse/KOGITO-491

ricardozanini:
Replied in the JIRA.

hack/run-e2e-cli.sh
```diff
@@ -0,0 +1,80 @@
...
+echo "-------- Route is ${route}"
+response=$(curl "http://${route}/hello")
+
+if [ "$response" != "Mario is older than Mark" ]; then
+  echo "Failed to deploy application, response: '${response}' not expected"
+  exit 1
+fi
+
+echo "------- KogitoApp successful deployed, cleaning up"
+oc delete kogitoapp/kogito-example -n ${namespace}
```
xiezhang7: For the cleaning up, we could test deleting the app with cli and verify the results.

ricardozanini: We don't have the undeploy feature yet. Tracked here: https://issues.jboss.org/browse/KOGITO-149

ricardozanini: I'm so stupid.. there's the `kogito delete-service` feature. I implemented and completely forget about it.

Will change this.

PR #98, opened 2019-10-30
[KOGITO-378]  Integrate Kogito Operator with Prometheus Operator

xiezhang7:
Jira: https://issues.jboss.org/browse/KOGITO-378

Description: deploy a `Service Monitor` for scraping the metrics of the Kogitio app, if `Prometheus Operator` is installed.


pkg/client/client.go
```diff
@@ -52,10 +54,11 @@ var (
...
 	// ControlCli is a reference for the controller-runtime client, normally built by a Manager inside the controller context.
-	ControlCli controllercli.Client
-	BuildCli   buildv1.BuildV1Interface
-	ImageCli   imagev1.ImageV1Interface
-	Discovery  discovery.DiscoveryInterface
+	ControlCli    controllercli.Client
+	BuildCli      buildv1.BuildV1Interface
+	ImageCli      imagev1.ImageV1Interface
+	Discovery     discovery.DiscoveryInterface
+	PrometheusCli monclientv1.MonitoringV1Interface
```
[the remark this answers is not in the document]
xiezhang7: The prometheus operator api uses pointers in the item list, e.g. [ServiceMonitorList](https://github.com/coreos/prometheus-operator/blob/25144969947563ccc480079b2533375a84d84adf/pkg/apis/monitoring/v1/types.go#L774). But, ControlCli cannot handle this kind of api. So I use the prometheus operator client instead.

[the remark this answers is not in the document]
ricardozanini: Thanks for your clarification, @xiezhang7. I'll take a closer look tomorrow in this PR.

sutaakar:
Briefly went through the PR, didn't find there anything suspicious.
I will make more detailed review later, possible today or tomorrow.
If I won't make it this week then don't wait with merging for me.

ricardozanini:
@xiezhang7 is it possible to use the `Discovery` client to find if the Prometheus API is available in the cluster like we do today with [`isOpenShift`](https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/client/client.go#L62) method in the `client` package? 

This way you just create the `ServiceMonitor` if the API is available using the CR and the standard client.

xiezhang7:
> @xiezhang7 is it possible to use the `Discovery` client to find if the Prometheus API is available in the cluster like we do today with [`isOpenShift`](https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/client/client.go#L62) method in the `client` package?
> 
> This way you just create the `ServiceMonitor` if the API is available using the CR and the standard client.

@ricardozanini yes, we can do that. But we will still need the prometheus client to get the list of deployed service monitors, because the standard client doesn't support that.

ricardozanini:
> > @xiezhang7 is it possible to use the `Discovery` client to find if the Prometheus API is available in the cluster like we do today with [`isOpenShift`](https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/client/client.go#L62) method in the `client` package?
> > This way you just create the `ServiceMonitor` if the API is available using the CR and the standard client.
> 
> @ricardozanini yes, we can do that. But we will still need the prometheus client to get the list of deployed service monitors, because the standard client doesn't support that.

Cool! But why do we need to list the ServiceMonitors?

xiezhang7:
> Cool! But why do we need to list the ServiceMonitors?

Because [operator-utils](https://github.com/RHsyseng/operator-utils/tree/k8s-1.14/pkg/resource) is used to maintain the resources. It lists the resources and filters them by the ownership. It doesn't make any assumption on the resource name or the amount of the resources.

ricardozanini:
> > Cool! But why do we need to list the ServiceMonitors?
> 
> Because [operator-utils](https://github.com/RHsyseng/operator-utils/tree/k8s-1.14/pkg/resource) is used to maintain the resources. It lists the resources and filters them by the ownership. It doesn't make any assumption on the resource name or the amount of the resources.

hmm I see. Well, I'm not that happy with this solution, but I see value using the operator-utils to manage our resources. It is what it is. 

Could you please them rebase your commit?

xiezhang7:
Thanks, @ricardozanini . I have rebased the code, and changed the [code](https://github.com/kiegroup/kogito-cloud-operator/blob/021766a57419a8b169e45ff399a0112c0ac4a415/pkg/controller/kogitoapp/resource/service_monitor.go#L90) to use `Discovery` client to check the `Prometheus` API.

pkg/controller/kogitoapp/resource/service_monitor_test.go
```diff
@@ -0,0 +1,260 @@
...
+				&corev1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"test": "test",
+						},
+					},
+				},
+				&client.Client{
+					ControlCli: fake.NewFakeClient(),
+					Discovery: &discfake.FakeDiscovery{
```
ricardozanini: I see that you're using this fake client in many places now, wdyt using [this package](https://github.com/kiegroup/kogito-cloud-operator/tree/master/pkg/test) to return a function with this fake client and  also to have the fake client in a `const`, so we avoid repeating ourselves?

pkg/resource/image_metadata.go
```diff
@@ -43,6 +43,15 @@ const (
 	// LabelKeyOrgKieProtoBuf is the label key for ProtoBuf metadata
 	LabelKeyOrgKieProtoBuf = "org.kie" + labelNamespaceSep + "persistence" + labelNamespaceSep + "proto"
 
+	// LabelPrometheusScrape is the label key for prometheus scrape configuration
+	LabelPrometheusScrape = LabelKeyPrometheus + "/scrape"
```
ricardozanini: I remember setting those constants somewhere else. Is they new?

.circleci/config.yml
```diff
@@ -15,4 +15,4 @@ jobs:
       - run: go fmt ./...
       - run: go vet ./...
       - run: CGO_ENABLED=0 go test -mod=vendor ./cmd/... -count=1 -p=1
-      - run: CGO_ENABLED=0 go test -mod=vendor ./pkg/... -count=1
+      - run: CGO_ENABLED=0 go test -mod=vendor ./pkg/... -count=1 -p=1
```
ricardozanini: Are we having concurrency problems in the `pkg` tests as well? :disappointed:

pkg/controller/kogitoapp/resource/service_monitor_test.go
xiezhang7: Moved it to the [test package](https://github.com/kiegroup/kogito-cloud-operator/pull/98/files#diff-eb7b36bde76ac14decc92ec9f8976b75).

pkg/resource/image_metadata.go
xiezhang7: They are used in some unit tests. I have modified the unit tests to use the const variables.

.circleci/config.yml
xiezhang7: yes

spolti:
I didn't find anything suspicious as well, but, as a user, it is no clear how it works, or what exactly needs to be done to enable/access the metrics, and how to take advantage of it.
Do you mind to add a few detailed information about it?

xiezhang7:
> I didn't find anything suspicious as well, but, as a user, it is no clear how it works, or what exactly needs to be done to enable/access the metrics, and how to take advantage of this integration, the user will need to do some manual configuration besides add the prometheus addon on their project, or will the operator take care of the prometheus operator integration and run it automatically?
> Do you mind to add a few detailed information about it?

Thanks, @spolti . I have updated the README.

1. Accessing the metrics is configured by the prometheus operator. The integration only works when there is already a prometheus operator install in the cluster, which means the user is supposed to have some knowledge of the prometheus operator.  Mainly, the user will have to create or configure a `Prometheus` instance for the prometheus operator. I added the information in the [README](https://github.com/kiegroup/kogito-cloud-operator/blob/ec8e92f61102744b1b50a3f23665e987255c6922/README.md#prometheus-operator) .

2. Enabling/generating metrics is configured in the kogito source project. It is mentioned in the first part of the [Prometheus Integration](https://github.com/kiegroup/kogito-cloud-operator/blob/ec8e92f61102744b1b50a3f23665e987255c6922/README.md#prometheus-annotations) section.

3. The integration is to let the kogito operator create a `Service Monitor` instance of the prometheus operator automatically, when prometheus operator is installed and metrics for the kogito app is enabled. The `Service Monitor` instance is responsible for describing how the `Prometheus` instance should scrape the metrics of the Kogito App. 

So the user is supposed to finish 1 and 2. The operator will do 3.

PR #100, opened 2019-11-05
[KOGITO-494] Injecting KOGITO_DATAINDEX_URL in kogito runtime services

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-494

In this PR we introduce the `integration` package that will handle calls to the deployed infrastructure to inject service/URLs to the Kogito Runtime Services. For this JIRA, we're adding the Data Index route if one exists in the namespace.

Also, a small refactoring to the `CreateFakeClient` function to avoid repetition in the unit tests. If you need a mock kubernetes client, use this one from the `pkg/test` package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -624,7 +624,7 @@ Change log level at runtime with the `DEBUG` environment variable. e.g. -
 ```bash
 $ make mod
 $ make clean
-$ DEBUG="true" operator-sdk up local --namespace=<namespace>
+$ DEBUG=true operator-sdk up local --namespace=<namespace>
 ```
 
 Before submitting PR, please be sure to read the [contributors guide](docs/CONTRIBUTING.MD).
--- a/cmd/kogito/command/test/common.go
+++ b/cmd/kogito/command/test/common.go
@@ -16,17 +16,16 @@ package test
 
 import (
 	"bytes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/test"
 	"github.com/spf13/cobra"
 	"path/filepath"
 	"strings"
 
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
-	"github.com/kiegroup/kogito-cloud-operator/pkg/client/meta"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/util"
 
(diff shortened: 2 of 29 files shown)


pkg/controller/kogitoapp/integration/integration.go
```diff
@@ -0,0 +1,50 @@
...
+func InjectEnvVarsFromExternalServices(kogitoApp *v1alpha1.KogitoApp, dc *v1.DeploymentConfig, client *client.Client) error {
+	// We look for a deployed data index to inject into the runtime service
+	// later we could also integrate with other external services like Kafka, Infinispan and SSO
+	route, err := getKogitoDataIndexRoute(client, kogitoApp.GetNamespace())
+	if err != nil {
+		return err
+	}
+
+	if len(dc.Spec.Template.Spec.Containers) > 0 {
+		oldRoute := util.GetEnvVarFromContainer(kogitoDataIndexRouteEnv, dc.Spec.Template.Spec.Containers[0])
```
xiezhang7: I don't think it is necessary to compare with the old value here, only need to describe the expected deployment. The comparison with the deployed DC will be done in [resource_comparator.go](https://github.com/kiegroup/kogito-cloud-operator/blob/0569bc0a936135dcc44e5340d5c0564aae84c393/pkg/controller/kogitoapp/resource/resource_comparator.go#L57) . 

Everything else looks good to me. Thanks!

ricardozanini: Yeah, I was in doubt wether or not adding the comparison in here to not update the resource of not needed. I’ll run some tests removing this line: deleting the data index should remove the variable.

ricardozanini:
@xiezhang7 could you take a second look, please?
I've changed a lot. After some tests, I've figured out that the reconciliation phase on `kogitoapp_controller` won't work to watch `KogitoDataIndex` objects since we receive from the request object the value `kogito-data-index/namespace` and not the `Namespaced` object for the `KogitoApp`, hence we can't fetch for the kogitoapp!! The loop just returns without doing anything.

The solution is to have two functions: one to grab the route once we reconcile the `KogitoApp` or create a new one and the other to update every app during the `KogitoDataIndex` reconciliation. This way we guarantee that once we don't have a route anymore, every `KogitoApp` got it's injection point deleted instead of waiting for the default reconciliation time.

It's working like a charm.

PR #110, opened 2019-11-15
[KOGITO-600] - Removing default build limits

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-600

Simple update to the KogitoApp to remove default builds, avoiding side effects like the one exemplified in the JIRA

### DIFF
--- a/pkg/controller/kogitoapp/kogitoapp_controller.go
+++ b/pkg/controller/kogitoapp/kogitoapp_controller.go
@@ -29,7 +29,6 @@ import (
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/openshift"
 	kogitores "github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoapp/resource"
-	"github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoapp/shared"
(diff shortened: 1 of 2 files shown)


ricardozanini:
> That means that the workaround on https://issues.jboss.org/browse/KOGITO-504 don't apply anymore or ?

Exactly. But this can be consider a workaround anyway, since will use all resources that are available in the node.

ricardozanini:
> there is only this tests which does limits tests?

I didn't get it. What did you mean? 

In the scenario that we tackled, a cluster admin set the Limit below 1 CPU and 2Gi of memory for builds.

That's fine since the user can workaround this by setting a limit below this values in the Kogito App CR.

The problem is that once we set a value for resources limits in BC, the init containers for s2i builds, also set their own limits and we **don't have control over it**. We don't have an option "blank limits" in our CR since we set default limit builds, that's the problem. 

Removing the defaults and leaving this values to the cluster admin/dev to set would be better and more "wider" approach, since setting default limits bring other consequences like the exemplified above.

spolti:
I meant, in the operator tests, all tests that  is related with build limits are the one you removed?

ricardozanini:
> I meant, in the operator tests, all tests that is related with build limits are the one you removed?

I just removed the ones that consider setting the default.

ricardozanini:
@xiezhang7 could you please take a look?

PR #125, opened 2019-12-11
[KOGITO-711] Implement integration test for Kogito persistence

sutaakar:
https://issues.redhat.com/browse/KOGITO-711


sutaakar:
Sometimes I have observed that deployed application didn't properly return persisted order when the application pod was respinned (returning empty list of orders). Does Kogito retrieve persisted content once it start up or already during deployment?

go.mod
```diff
@@ -6,13 +6,15 @@ require (
...
 	github.com/go-openapi/spec v0.19.3
+	github.com/go-openapi/validate v0.18.0 // indirect
 	github.com/gobuffalo/packr/v2 v2.7.1
 	github.com/google/go-cmp v0.3.1
 	github.com/google/uuid v1.1.1
 	github.com/infinispan/infinispan-operator v0.0.0-20191028165532-917f400bee62
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/openshift/api v0.0.0-00010101000000-000000000000
-	github.com/openshift/client-go v0.0.0-00010101000000-000000000000
+	github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible
```
ricardozanini: @xiezhang7 could you confirm if this is the right version of the openshift API? I understand that we should replace it with 4.x.

pkg/apis/kafka/v1beta1/zz_generated.openapi.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !ignore_autogenerated
```
ricardozanini: Why this one was removed?

test/e2e/kogitoapp_test.go
```diff
@@ -179,6 +232,34 @@ func initializeKogitoOperator(t *testing.T, f *framework.Framework, ctx *framewo
...
+
+	log.Info("Installing Infinispan operator")
+
+	operatorGroup := getOperatorGroup(namespace, namespace)
+	err = f.Client.Create(goctx.TODO(), operatorGroup, &framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
+	if err != nil {
+		log.Fatalf("could not create operator group: %v", err)
+	}
+
+	subscription := getSubscriptionSingleNamespace("infinispan", namespace, "infinispan", "community-operators", "stable")
```
ricardozanini: by only creating a subscription reference were you able to deploy Infinispan on OCP 4.x? We might need that also here: https://issues.redhat.com/browse/KOGITO-703
Looking at the docs, looks like it's needed other objects to be feasible.

pkg/apis/kafka/v1beta1/zz_generated.openapi.go
sutaakar: That was removed by build, should I put it back?

test/e2e/kogitoapp_test.go
sutaakar: I followed https://docs.openshift.com/container-platform/4.2/operators/olm-adding-operators-to-cluster.html#olm-installing-operator-from-operatorhub-using-cli_olm-adding-operators-to-a-cluster for this. According my local experiments it should be enough (tested on OCP 4.2).

pkg/apis/kafka/v1beta1/zz_generated.openapi.go
ricardozanini: Are you using golang 1.12? If so, keep it this way.

test/e2e/kogitoapp_test.go
ricardozanini: ... and the `OperatorGroup` as well. I think it's safe to follow this approach in the CLI as well.
The operator appears in the `Installed Operators` tab like if it was installed via UI?

pkg/apis/kafka/v1beta1/zz_generated.openapi.go
sutaakar: go1.12.14

test/e2e/kogitoapp_test.go
sutaakar: yes, it is shown in UI, though it takes quite some time (around 2 minutes) until the operator itself is spinned up (not sure why).

test/e2e/example_verifier.go
```diff
@@ -56,3 +62,83 @@ func verifyJbpmSpringBootExample(t *testing.T, kogitoService *v1alpha1.KogitoApp
...
+	kogitoApp.Spec.Replicas = &convertedReplicas
+
+	// Update Kogito application
+	if err := f.Client.Update(goctx.TODO(), kogitoApp); err != nil {
+		verifierLog.Fatalf("Error while trying to update Kogito application: %s", err)
+	}
+
+	//wait until desired number of replicas are available
+	dc := appsv1.DeploymentConfig{}
+	for i := 1; i <= 60; i++ {
```
radtriste: Maybe we could factorize those waiting loops to have cleaner code

something like `waitFor(fn function(interface{}) error)` or similar ?

go.mod
xiezhang7: I don't think the version here matters, because of the `replace` at line 35.

ricardozanini:
@sutaakar do you mind writing a proper e2e section in the README or a new markdown doc inside the `doc` dir and link it into the README? Like the steps we need to perform those tests would be awesome.

sutaakar:
@ricardozanini What should I write about? The steps should be already covered by https://github.com/kiegroup/kogito-cloud-operator#running-end-to-end-e2e-tests .

ricardozanini:
> @ricardozanini What should I write about? The steps should be already covered by https://github.com/kiegroup/kogito-cloud-operator#running-end-to-end-e2e-tests .

We can't choose which test to run? I mean, this PR will just enhance the already existent test?

go.mod
ricardozanini: It's the same major version. And I think we need to support 3.x since we're still running some stories in this OCP version.

sutaakar:
The test group to run is selected by "tests" parameter in make run-e2e. This PR just add another test into jvm and full test groups.
AFAIK currently there is no possibility to run single test (without commenting out other tests in code).

ricardozanini:
Yeah, didn't give too much attention to e2e yet TBH. Just wrote the first implementation based on a very basic example. I trust in your judgement, though. :)

sutaakar:
That is usually the way how large and complex test frameworks starts :) 
Anyway, these tests may get transformed into the Cucumber (based on the feedback and usability).

sutaakar:
Reimplemented waiting loops with dedicated method.

ricardozanini:
@spolti this means that you have approved it? =D

spolti:
hmmm, I didn't checked the approve option LOL

PR #136, opened 2020-01-09
[KOGITO-852] Detect running operator using subscription name

sutaakar:
https://issues.redhat.com/browse/KOGITO-852

### DIFF
--- a/pkg/infrastructure/operator.go
+++ b/pkg/infrastructure/operator.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/operator"
 
 	v1 "k8s.io/api/apps/v1"
@@ -31,7 +32,7 @@ func CheckKogitoOperatorExists(kubeCli *client.Client, namespace string) (bool, 
 	return CheckOperatorExists(kubeCli, namespace, operator.Name)
 }
 
-// CheckOperatorExists checks whether operator is existing and running.
+// CheckOperatorExists checks whether operator is existing and running. It expects that deployment name is equal to operator name.
 // If it is existing but not running, it returns true and an error
 func CheckOperatorExists(kubeCli *client.Client, namespace, operatorName string) (bool, error) {
 	log.Debugf("Checking if %s Operator is deployed in namespace %s", operatorName, namespace)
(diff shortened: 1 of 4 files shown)


pkg/infrastructure/operator.go
```diff
@@ -54,3 +55,42 @@ func CheckOperatorExists(kubeCli *client.Client, namespace, operatorName string)
 
 	return true, nil
 }
+
+// CheckOperatorExistsUsingSubscription checks whether operator is existing and running. For this check informations from subscription are used.
+// If it is existing but not running, it returns true and an error
+func CheckOperatorExistsUsingSubscription(kubeCli *client.Client, namespace, subscriptionName string) (bool, error) {
```
radtriste: That one seems really nice and more generic :)
I would suggest we keep the CheckKogitoOperatorExists as it was before (means code from CheckOperatorExists would be set back to CheckKogitoOperatorExists). That one would be used for all other operators.
Or even use it for Kogito operator ? @ricardozanini wdyt ?

sutaakar: This code can be used to detect operators deployed using OLM.
In case operator is deployed using YAML files then there is no Subscription to carry the informations.
Current implementation is limited to single namespace subscriptions, for global subscriptions the adjustments need to be investigated.

radtriste: ok you are right, forgot this option ...
good for me then

pkg/infrastructure/operator.go
```diff
@@ -54,3 +55,42 @@ func CheckOperatorExists(kubeCli *client.Client, namespace, operatorName string)
...
+	log.Debugf("Checking if Operator with %s Subscription is deployed in namespace %s", subscriptionName, namespace)
+
+	subscription := &olmapiv1alpha1.Subscription{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      subscriptionName,
+			Namespace: namespace,
+		},
+	}
+	if exists, err := kubernetes.ResourceC(kubeCli).Fetch(subscription); err != nil {
```
ricardozanini: I don't think that fetching only by name will do, since subscriptions can have generated names, that's why we fetch for `Package` and `CatalogSource` as well.

Won't be better to reuse this function? And externalize it to the `framework` (new with KOGITO-721 PR) package to be used by tests. They are very alike.

https://github.com/kiegroup/kogito-cloud-operator/blob/master/cmd/kogito/command/shared/operatorhub.go#L77

sutaakar: Good point, will take a look on it.

ricardozanini:
Waiting for OLM PR to be merged to merge this one.

pkg/infrastructure/operator.go
radtriste: As this one was merged, created https://issues.redhat.com/browse/KOGITO-916 to follow-up on this (and assigned to you @sutaakar if you have time. If not I can have a look)

PR #143, opened 2020-01-17
[KOGITO-614] Deploy Kafka for Data Index with Infra

xiezhang7:
Jira: https://issues.redhat.com/browse/KOGITO-614

Description:  use KogitoInfra to also deploy the Kafka instance for Data Index.


ricardozanini:
@xiezhang7 could you please rebase your branch? Many thanks!

deploy/crds/app.kiegroup.org_v1alpha1_kogitodataindex_cr.yaml
```diff
@@ -16,8 +16,8 @@ spec:
   cpuRequest: ""
   # details about the kafka connection
   kafka:
-    # the Kafka service instance to be used. Example: kogito-kafka
-    instance: kogito-kafka
+    # let's leave this burden to KogitoInfra CR to deploy a new Kafka instance for us
```
sutaakar: The comment seems to be misleading. From the CRD description above:
> UseKogitoInfra flags if the instance will use a provided
                    infrastructure by KogitoInfra CR. Setting this to true will configure
                    a KogitoInfra CR to install Kafka via Strimzi Operator.
From what I understand setting this property to false means that user will provide Kafka connection informations manually, not relying on KogitoInfra.

pkg/controller/kogitodataindex/resource/kafka_test.go
```diff
@@ -94,7 +98,7 @@ func TestIsKafkaServerURIResolved(t *testing.T) {
 				&v1alpha1.KogitoDataIndex{},
 				cli,
 			},
-			true,
+			false,
```
sutaakar: What is the difference in "AnyInstanceURI" test and "NoServer" test? Both provide empty KogitoDataIndex and expect URL not to be resolved.

pkg/infrastructure/kafka.go
```diff
@@ -101,3 +103,75 @@ func GetKafkaServiceURI(cli *client.Client, infra *v1alpha1.KogitoInfra) (uri st
...
+	return ""
+}
+
+// ResolveKafkaServerReplicas returns the number of replicas of the kafka instance
+func ResolveKafkaServerReplicas(kafka *v1beta1.Kafka) int32 {
+	if kafka != nil {
+		if kafka.Spec.Kafka.Replicas > 0 {
+			return kafka.Spec.Kafka.Replicas
+		}
+		return 1
```
sutaakar: Why the 1 is returned here when 'kafka.Spec.Kafka.Replicas' is 0 or less?

pkg/infrastructure/kafka.go
```diff
@@ -101,3 +103,75 @@ func GetKafkaServiceURI(cli *client.Client, infra *v1alpha1.KogitoInfra) (uri st
...
+}
+
+// GetKafkaInstanceInNamespace fetches one Kafka instance in the given namespace
+func GetKafkaInstanceInNamespace(namespace string, client *client.Client) (*v1beta1.Kafka, error) {
+	kafkaList := &v1beta1.KafkaList{}
+	if err := kubernetes.ResourceC(client).ListWithNamespace(namespace, kafkaList); err != nil {
+		return nil, err
+	} else if len(kafkaList.Items) == 1 {
+		return &kafkaList.Items[0], nil
+	}
```
sutaakar: What if there are several kafka instances in namespace?

sutaakar:
Few minor comments, otherwise looks ok.

pkg/infrastructure/kafka.go
xiezhang7: Because the code sets the default value of this property is 0 if no other value is provided, and a value which is less than or equal to 0 doesn't make any sense. So set the replicas to 1 in this case.

pkg/infrastructure/kafka.go
xiezhang7: Actually, this function is deprecated. I'll delete it.

pkg/controller/kogitodataindex/resource/kafka_test.go
xiezhang7: The test case "AnyInstanceURI" doesn't make any sense anymore. I'll change it.

deploy/crds/app.kiegroup.org_v1alpha1_kogitodataindex_cr.yaml
xiezhang7: I'll change the values. Actually, the operator will change the value to true if the user doesn't explicitly specify which external kafka or infinispan to use.

PR #148, opened 2020-01-20
[KOGITO-692] Allowing specific images to be set for build configurations

ricardozanini:
See:
https://issues.redhat.com/browse/KOGITO-692

In this PR we modified the `KogitoApp` CR to include `imageVersion`, so users could set their preferred version for the builds.

Also, we now accept the full tag name in the `Spec.Build.ImageRuntime|S2I.imageStreamTag` attribute. This will create a new custom `ImageStream` resource with the given tag. See examples in the `deploy/examples/example-custom-build-images.yaml` file.

Needs more tests in the cluster, which I'll do it tomorrow before flying :flight_departure: 


cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -149,10 +150,11 @@ func (i *deployCommand) InitHook() {
...
+	i.command.Flags().StringVar(&i.flags.imageRuntime, "image-runtime", "", "Custom image tag for the s2i build, e.g: quay.io/mynamespace/myimage:latest")
 	i.command.Flags().StringVar(&i.flags.installInfinispan, "install-infinispan", defaultInstallInfinispan, "Infinispan installation mode: \"Always\", \"Never\" or \"Auto\". \"Always\" will install Infinispan in the same namespace no matter what, \"Never\" won't install Infinispan even if the service requires it and \"Auto\" will install only if the service requires persistence.")
 	i.command.Flags().StringVar(&i.flags.installKafka, "install-kafka", defaultInstallKafka, "Kafka installation mode: \"Always\" or \"Never\". \"Always\" will use the Strimzi Operator to install a Kafka cluster. The environment variable 'KAFKA_BOOTSTRAP_SERVERS' will be available for the service during runtime.")
+	i.command.Flags().StringVar(&i.flags.imageVersion, "image-version", "", "Image version for standard Kogito build images. Ignored if a custom image is set for image-s2i or image-runtime.")
 }
```
radtriste: no way to specify the `imageStreamNamespace` ?

radtriste: no way to specify the `imageStreamNamespace` ?

ricardozanini:
> If you leave the namespace, why not do like for the version ?
> and then, `imageRuntime` and `imageS2I` have only one attribute, so we could factorize, and at the end have something like that (all at same level):
> 
> ```
> imageVersion: ###
> imageStreamNamespace: ###
> imageRuntimeTag: ###
> imageS2ITag: ###
> ```

TBH I think we should drop the `namespace` attribute.

Doesn't make sense to have the images across namespaces at this moment and we will refactor the build process to work on Kubernetes in the future. The CI use case that I mentioned in the JIRA is still valid, but in that case the IS will be set on the "build" namespace. 

This namespace thing came into a time where we weren't creating IS. So, this should be refactored / dropped when we introduced this feature.

Will be simpler as well to implement and do understand. wdyt?

@spolti @sutaakar, mind to chime in?

sutaakar:
Sounds good to me.

cmd/kogito/command/deploy/deploy_service.go
ricardozanini: No way :p

hahaha let's see where our discussions regarding dropping namespace definitely goes.

deploy/examples/example-custom-build-images.yaml
```diff
@@ -0,0 +1,19 @@
...
+    #  value: "http://nexus3.kogito-tools.svc.cluster.local:8081/repository/maven-public/"
+    gitSource:
+      contextDir: jbpm-quarkus-example
+      uri: 'https://github.com/kiegroup/kogito-examples'
+      reference: 0.7.0-rc4
+    # images that don't have a custom tag, will use this version
+    imageVersion: 0.7.0-rc4
+    imageRuntime:
+      # the image runtime build will use this image
+      imageStreamTag: quay.io/ricardozanini/kogito-quarkus-jvm-ubi8:0.7.0-rc1
```
spolti: maybe use the kiegroup tag 0.7.7-rc4?

pkg/apis/app/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,17 +1,3 @@
-// Copyright 2020 Red Hat, Inc. and/or its affiliates
```
spolti: do you know why was it removed?

deploy/examples/example-custom-build-images.yaml
ricardozanini: This is to show that we can have a custom tag, thus the namespace naming (and with an actual tag that exists).

pkg/apis/app/v1alpha1/zz_generated.deepcopy.go
ricardozanini: It's auto-generated. Some times it removes, next commit will add it again.

deploy/examples/example-custom-build-images.yaml
```diff
@@ -0,0 +1,18 @@
...
+    #- name: MAVEN_MIRROR_URL
+    #  value: "http://nexus3.kogito-tools.svc.cluster.local:8081/repository/maven-public/"
+    gitSource:
+      contextDir: jbpm-quarkus-example
+      uri: 'https://github.com/kiegroup/kogito-examples'
+      reference: 0.7.0-rc4
+    # images that don't have a custom tag, will use this version
+    imageVersion: 0.7.0-rc4
+    # the image runtime build will use this image
+    imageRuntimeTag:  quay.io/ricardozanini/kogito-quarkus-jvm-ubi8:0.7.0-rc1
```
spolti: maybe use kiegroup and tag 0.7.0.rc4?

deploy/examples/travelapp-kogito-travel-agency.yaml
```diff
@@ -17,10 +17,7 @@ spec:
       contextDir: 05-kogito-travel-agency
       reference: override-bootstrap
       uri: https://github.com/ricardozanini/kogito-travel-agency-tutorial
```
spolti: isn't this example under kiegroup org?

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -53,6 +54,11 @@ import (
 
 var log = logger.GetLogger("controller_kogitoapp")
 
+const (
+	// imageStreamNamespace default namespace for the ImageStreams
+	imageStreamNamespace = "openshift"
```
spolti: Isn't this same env declared somewhere? I do remember to see it but nor sure where exactly

ricardozanini: This will be removed.

deploy/examples/travelapp-kogito-travel-agency.yaml
ricardozanini: Yes, I'll change it.

deploy/examples/example-custom-build-images.yaml
ricardozanini: The tag is already used by the `version`. The point here is to have a different namespace, so others can understand that they might have their own image in their own namespace.

deploy/examples/travelapp-kogito-travel-agency.yaml
ricardozanini: Actually, only on Maciej's repo. Since it's the source, I'll add it.

PR #153, opened 2020-01-24
[KOGITO-956] Disable smoke test for springboot-example with persistence because of KOGITO-948

radtriste:
https://issues.redhat.com/browse/KOGITO-956

Will be later enabled in https://issues.redhat.com/browse/KOGITO-957, when problem is solved...

### DIFF
--- a/test/smoke/features/cli/cli_deploy_springboot_service.feature
+++ b/test/smoke/features/cli/cli_deploy_springboot_service.feature
@@ -13,24 +13,25 @@ Feature: CLI: Deploy Spring Boot service
 
 #####
 
-  Scenario: CLI deploy jbpm-springboot-example service with persistence
-    Given Kogito Operator is deployed with dependencies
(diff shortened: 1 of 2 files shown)


sutaakar:
@radtriste What about disabling tests using tags?
```
 -t, --tags=          Filter scenarios by tags. Expression can be:
                           - "@wip": run all scenarios with wip tag
                           - "~@wip": exclude all scenarios with wip tag
                           - "@wip && ~@new": run wip scenarios, but exclude new
                           - "@wip,@undone": run wip or undone scenarios
```

We can define for example Ignore as disabling tag

radtriste:
@sutaakar 
I will update that as part of https://issues.redhat.com/browse/KOGITO-958 then :)
We could then put more tags like @persistence, @messaging, @cli,... instead of different folders ?Will add your suggestions to the specified issue.
Need to be discussed on how we want to organize the features folder :)

Is that ok ?

sutaakar:
Sounds ok

PR #158, opened 2020-01-30
[KOGITO-972] Allow to change the image-builder when building the operator

radtriste:
https://issues.redhat.com/browse/KOGITO-972

Also added some parameters in Makefile to change the image information, so we can build easily a local image with own registry, name and/or tag.

### DIFF
--- a/hack/go-build.sh
+++ b/hack/go-build.sh
@@ -13,9 +13,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+SCRIPT_NAME=`basename $0`
+
+function usage(){
+  printf "Build the operator."
+  printf "\n"
+  printf "\n${SCRIPT_NAME} [options]*"
+  printf "\n"
+  printf "\nOptions:"
+  printf "\n"
+  printf "\n-h | --help\n\tPrint the usage of this script."
+  printf "\n--image_registry {IMAGE_REGISTRY}\n\tSet the Image Registry."
+  printf "\n--image_name {IMAGE_NAME}\n\tSet the Image Name."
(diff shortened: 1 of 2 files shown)


hack/go-build.sh
```diff
@@ -24,12 +41,46 @@ TAG=0.7.0
...
 
 setGoModEnv
 go generate ./...
 if [[ -z ${CI} ]]; then
     ./hack/go-test.sh
-    operator-sdk build ${REGISTRY}/${IMAGE}:${TAG}
+    BUILD_PARAMS=""
+    if [[ ! -z ${BUILDER} ]]; then BUILD_PARAMS="${BUILD_PARAMS} --image-builder ${BUILDER}"; fi
+    operator-sdk build ${REGISTRY}/${IMAGE}:${TAG} ${BUILD_PARAMS}
     if [[ ${1} == "rhel" ]]; then
```
ricardozanini: I think we can drop this, let's take care of productization once we have the Cekit enabled.

spolti: not sure what would be better, still with cekit or use operator-sdk build.
Maybe have both enabled?

ricardozanini: I'd rather going for Cekit, like we are doing on downstream, or maybe a `operator-sdk` plugin for cekit :thinking:

spolti: I don't think cekit will have operator-sdk plugin.
but let's use cekit only for now.

spolti: what about leave it as it is and update once we migrate to cekit?

radtriste:
referencing the issue for Cekit migration: https://issues.redhat.com/browse/KOGITO-233

PR #165, opened 2020-02-04
[KOGITO-950] Store test logs into local filesystem

sutaakar:
https://issues.redhat.com/browse/KOGITO-950

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -32,6 +32,7 @@ import (
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	controllercli "sigs.k8s.io/controller-runtime/pkg/client"
@@ -62,12 +63,13 @@ var (
 // Client wraps clients functions from controller-runtime, Kube and OpenShift cli for generic API calls to the cluster
 type Client struct {
 	// ControlCli is a reference for the controller-runtime client, normally built by a Manager inside the controller context.
-	ControlCli    controllercli.Client
-	BuildCli      buildv1.BuildV1Interface
-	ImageCli      imagev1.ImageV1Interface
-	Discovery     discovery.DiscoveryInterface
-	PrometheusCli monclientv1.MonitoringV1Interface
-	DeploymentCli appsv1.AppsV1Interface
+	ControlCli             controllercli.Client
+	BuildCli               buildv1.BuildV1Interface
+	ImageCli               imagev1.ImageV1Interface
+	Discovery              discovery.DiscoveryInterface
+	PrometheusCli          monclientv1.MonitoringV1Interface
+	DeploymentCli          appsv1.AppsV1Interface
+	KubernetesExtensionCli kubernetes.Interface
 }
 
 // NewForConsole will create a brand new client using the local machine
--- a/pkg/client/client_builder.go
+++ b/pkg/client/client_builder.go
@@ -16,10 +16,12 @@ package client
 
 import (
 	"fmt"
+
 	appsv1 "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	buildv1 "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	imagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	controllercli "sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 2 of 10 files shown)


sutaakar:
Still need to make minor adjustments to cover also case with several containers in pod, but otherwise should work.
@radtriste Please take a look if the code is (moreless) understandable.

pkg/client/client.go
```diff
@@ -153,7 +153,7 @@ func newKubeClient(config *restclient.Config) (controllercli.Client, error) {
 	return controlCli, nil
 }
 
-func buildKubeConnectionConfig() (*restclient.Config, error) {
+func BuildKubeConnectionConfig() (*restclient.Config, error) {
```
sutaakar: @ricardozanini To access the pod logs I use client from "k8s.io/client-go/kubernetes", do you know about any simpler way?
Also to use the client I need access to kubeconfig (that is why I exported this function). Should I move it to infrastructure?

ricardozanini: @sutaakar if you need a specific method from the Pod API, please exposed it instead of the `buildKube..` function. On our code base, we aren't accessing the kube client directly, please see this implementation as an example:
https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/client/openshift/deploymentconfig.go

In this case, you create a `pod.go` file within the `kubernetes` package. :)

sutaakar: Please take another look, implementation refactored and moved to `pod.go`

test/smoke/framework/logging.go
```diff
@@ -0,0 +1,123 @@
...
+		return errors.New("namespace is not monitored")
+	}
+	monitoredNamespace.stopMonitoring <- true
+	waitUntilNamespacePodsLogFinish(monitoredNamespace)
+	return nil
+}
+
+func waitUntilNamespacePodsLogFinish(monitoredNamespace monitoredNamespace) {
+	for _, loggingFinished := range monitoredNamespace.pods {
+		<-loggingFinished
```
radtriste: So the logging is finished when the pod is not running anymore and log stream is over right and here we are waiting for that right ?

sutaakar: Yes, waiting until all log finishes, as log files are retrireved when pod terminates.

radtriste: What happens if the pod is never deleted ? for example, if we don't delete the namespace ?
(could be useful sometimes)

spolti: Are chains/channels auto closed on Go?

spolti:
This looks good, same concern than @radtriste

test/smoke/framework/logging.go
sutaakar: Good point, will address it.

pkg/client/client_builder.go
```diff
@@ -47,6 +49,8 @@ type Builder interface {
 	WithPrometheusClient() Builder
 	// WithDeploymentClient tells the builder to create the deployment client
 	WithDeploymentClient() Builder
+	// WithKubernetesClient tells the builder to create the kubernetes client
+	WithKubernetesClient() Builder
```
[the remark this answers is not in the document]
sutaakar: Done.

sutaakar:
@radtriste Partly refactored and reformatted.
Now when StopPodLogCollector is called then logs of all currently running pods are stored. In case the pod is terminated later then his log will be ignored.

test/smoke/framework/logging.go
sutaakar: > Are chains/channels auto closed on Go?

They aren't, thanks for reminding. Will close the opened channels.

sutaakar:
Adjusted to log containers instead of pods.
Ready for review.

pkg/client/client.go
```diff
@@ -68,6 +69,7 @@ type Client struct {
 	Discovery     discovery.DiscoveryInterface
 	PrometheusCli monclientv1.MonitoringV1Interface
 	DeploymentCli appsv1.AppsV1Interface
+	KubernetesCli kubernetes.Interface
```
ricardozanini: I might be missing something, but why this? The client itself is a Kubernetes client, the others are extensions. We already have the `Namespace` interface that's suppose to work on the same way as the `Pod` you introduced. Sorry, I haven't enough time to go through everything yet. Could you please clarify? Thanks!!

sutaakar: Well, the problem is that log cannot be retrieved by CRUD on Kubernetes objects as log isn't a usual object.
It is retrieved as a subresource [1] using client. I didn't find any better way to retrieve it. Do you have any idea how can this be improved?

[1] https://github.com/alankritr/kubernative_lib/blob/4e35750abcfc7848719b84dd04346488fc2aa9fa/staging/src/k8s.io/client-go/kubernetes/typed/core/v1/pod_expansion.go#L44

sutaakar: Should we use k8s.io/client-go (currently used) dependency or github.com/kubernetes/client-go?
I can see that we already use github.com/openshift/client-go

ricardozanini: the openshift one is to query specifics APIs from Openshift such as ImageStream. Please stick with k8s.io if possible.

We could rename the `Kubernetes` API to `KubernetesExtension` instead, so won't cause confusion in the future.

sutaakar: Done, please take a look

PR #168, opened 2020-02-06
[KOGITO-983] Check first that the CLI exists

radtriste:
https://issues.redhat.com/browse/KOGITO-983

Change has been done as well how the smoke tests are started. No more direct call to godog application but call of go test. That means that we don't need to install godog binary anymore. 
Also, it gives us much more flexibility of configuration (--smoke.* options) and setup.

Also added `timeout` and `debug` options.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,3 +15,6 @@ jobs:
       - run: go vet ./...
       - run: CGO_ENABLED=0 go test ./cmd/... -count=1 -p=1
       - run: CGO_ENABLED=0 go test ./pkg/... -count=1 -p=1
+      - run: git clone https://github.com/bats-core/bats-core.git
+      - run: cd bats-core && ./install.sh $HOME
+      - run: cd hack/ && $HOME/bin/bats *.bats
--- a/test/smoke/framework/cli.go
+++ b/test/smoke/framework/cli.go
@@ -16,8 +16,25 @@ package framework
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 )
+
+// CheckCliBinaryExist checks if the CLI binary does exist
+func CheckCliBinaryExist() (bool, error) {
+	path, err := getEnvOperatorCliPath()
+	if err != nil {
+		return false, err
+	}
+
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
 
 // ExecuteCliCommand executes a kogito cli command for a given namespace
 func ExecuteCliCommand(namespace string, args ...string) (string, error) {
--- a/test/smoke/main_test.go
+++ b/test/smoke/main_test.go
@@ -15,11 +15,74 @@
 package operator
 
 import (
+	"flag"
+	"fmt"
+	"os"
+	"strings"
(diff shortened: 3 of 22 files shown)


test/smoke/framework/cli.go
```diff
@@ -16,9 +16,23 @@ package framework
...
 
+// CheckCliBinaryExist checks if the CLI binary does exist
+func CheckCliBinaryExist() (bool, error) {
+	path, err := getEnvOperatorCliPath()
+	if err != nil {
+		return false, err
+	}
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return false, nil
```
sutaakar: Please cover also branch when err is not nil but also not IsNotExist()

radtriste: updated

test/smoke/main_test.go
```diff
@@ -15,12 +15,64 @@
...
+
+	if !strings.Contains(opt.Tags, defaultDisabledTag) {
+		if opt.Tags != "" {
+			opt.Tags += " && "
+		}
+		opt.Tags += defaultDisabledTag
+	}
+
+	// Check CLI binary is existing
+	if exits, err := framework.CheckCliBinaryExist(); err != nil {
```
sutaakar: This means that CLI binary is required for running any test, right?

radtriste: yes... will modify that...

.circleci/config.yml
```diff
@@ -15,3 +15,6 @@ jobs:
       - run: go vet ./...
       - run: CGO_ENABLED=0 go test ./cmd/... -count=1 -p=1
       - run: CGO_ENABLED=0 go test ./pkg/... -count=1 -p=1
+      - run: git clone https://github.com/bats-core/bats-core.git
```
radtriste: @spolti It seems that the circle ci job is only running the `build` job (https://circleci.com/gh/kiegroup/kogito-cloud-operator/783#build-parameters/containers/0)
So for now, I included it into the `build` job, but I would like better to have a second job, like `bats-tests` job.
Do you know how we can update that ?

test/smoke/main_test.go
```diff
@@ -100,3 +112,12 @@ func deleteNamespaceIfExists(namespace string) {
 		}
 	}
 }
+
+func matchingFeature(tags string, features []*gherkin.Feature) bool {
```
sutaakar: tags parameter is unused

.circleci/config.yml
spolti: @radtriste for use multiple jobs, then a workflow is necessary:

https://circleci.com/docs/2.0/workflows/

but, in this case, I think that a sequential steps is ok, or, is there a specific reason to use multiple jobs for it?

spolti:
@radtriste is it ready to merge or are you going to modify the circleci job?

.circleci/config.yml
radtriste: If you are ok with the current solution, we can stick to it (and refactor later if needed)

radtriste:
@sutaakar are you ok with that ?

.circleci/config.yml
spolti: It is ok for me

radtriste:
Rebased and squashed

PR #188, opened 2020-02-25
[KOGITO-1114] Jobs service tests with Kogito app

sutaakar:
https://issues.redhat.com/browse/KOGITO-1114

### DIFF
--- a/test/features/cr/deploy_quarkus_service.feature
+++ b/test/features/cr/deploy_quarkus_service.feature
@@ -57,3 +57,33 @@ Feature: CR: Deploy quarkus service
     Examples:
       | native | minutes |
       | "enabled" | 20 |
+
+#####
+
+  # Disabled as long as https://issues.redhat.com/browse/KOGITO-1163 and https://issues.redhat.com/browse/KOGITO-1166 is not solved
+  @disabled
+  Scenario Outline: CR deploy timer-quarkus-example service with Jobs service
+    Given Kogito Operator is deployed
+    And Deploy Kogito Jobs Service with 1 replicas
+    And Deploy quarkus example service "timer-quarkus-example" with native <native>
+    And Kogito application "timer-quarkus-example" has 1 pods running within <minutes> minutes
+
+    When HTTP POST request on service "timer-quarkus-example" is successful within 2 minutes with path "timer" and body:
+      """json
+      { }
+      """
+
+    # Implement retrieving of job information from Jobs service once https://issues.redhat.com/browse/KOGITO-1163 is fixed
+    Then Kogito application "timer-quarkus-example" log contains text "Before timer" within 1 minutes
+    And Kogito application "timer-quarkus-example" log contains text "After timer" within 1 minutes
+
+    Examples: Non Native
+      | native | minutes |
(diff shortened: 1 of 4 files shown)


test/features/cr/deploy_quarkus_service.feature
```diff
@@ -56,4 +56,34 @@ Feature: CR: Deploy quarkus service
     @native
     Examples:
       | native | minutes |
-      | "enabled" | 20 |
\ No newline at end of file
+      | "enabled" | 20 |
+
+#####
+
+  # Disabled as long as https://issues.redhat.com/browse/KOGITO-1163 and https://issues.redhat.com/browse/KOGITO-1166 is not solved
```
radtriste: Do we have a follow-up task on those issues to enable the test when they are solved ? :)

test/features/cr/deploy_quarkus_service.feature
```diff
@@ -56,4 +56,34 @@ Feature: CR: Deploy quarkus service
...
+
+    # Implement retrieving of job information from Jobs service once https://issues.redhat.com/browse/KOGITO-1163 is fixed
+    Then Kogito application "timer-quarkus-example" log contains text "Before timer" within 1 minutes
+    And Kogito application "timer-quarkus-example" log contains text "After timer" within 1 minutes
+
+    Examples: Non Native
+      | native | minutes |
+      | "disabled" | 10 |
+
+    # Disabled as long as https://issues.redhat.com/browse/KOGITO-1179 is not solved
```
radtriste: same here, do we have a follow-up task on this issue to enable the test when it is solved ? :)

test/framework/kubernetes.go
```diff
@@ -133,3 +134,31 @@ func loadResource(namespace, uri string, resourceRef meta.ResourceObject, before
...
+
+		// Container name is equal to deployment config name
+		return checkPodsContainingTextInLog(namespace, pods, dcName, logText)
+	})
+}
+
+func checkPodsContainingTextInLog(namespace string, pods *corev1.PodList, containerName, text string) (bool, error) {
+	for _, pod := range pods.Items {
+		containsText, err := isPodContainingTextInLog(namespace, &pod, containerName, text)
+		if err != nil || !containsText {
```
radtriste: small mistake here I think
you should stop looking into pods if you find something or ?

sutaakar: Actually not. The purpose is to check all pods than they contain required text.
If test is found then another pod is inspected (for cycle continue), if text is not found then false is returned from the method.

test/features/cr/deploy_quarkus_service.feature
sutaakar: Not yet, will create one.

sutaakar: Created https://issues.redhat.com/browse/KOGITO-1265

test/framework/kubernetes.go
radtriste: Can we then call the method `*AllPods*` ? would be clearer I think

sutaakar: Done

PR #210, opened 2020-03-12
[KOGITO-1408] Install missing dependent crd before start of test

radtriste:
https://issues.redhat.com/browse/KOGITO-1408

Workaround int tests until KOGITO-1376 is implemented, as we would like the kogito pipelines to run correctly.

### DIFF
--- a/test/framework/operator.go
+++ b/test/framework/operator.go
@@ -40,6 +40,7 @@ type dependentOperator struct {
 	operatorPackageName string
 	timeoutInMin        int
 	channel             string
+	crdName             string
 }
 
 var (
@@ -49,16 +50,19 @@ var (
 			operatorPackageName: "infinispan",
 			timeoutInMin:        20,
 			channel:             "stable",
+			crdName:             "infinispans.infinispan.org",
 		},
 		"Kafka": {
 			operatorPackageName: "strimzi-kafka-operator",
 			timeoutInMin:        20,
 			channel:             "stable",
+			crdName:             "kafkas.kafka.strimzi.io",
 		},
 		"Keycloak": {
 			operatorPackageName: "keycloak-operator",
 			timeoutInMin:        20,
 			channel:             "alpha",
+			crdName:             "keycloaks.keycloak.org",
 		},
 	}
 )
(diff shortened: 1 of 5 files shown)


test/main_test.go
```diff
@@ -187,3 +191,54 @@ func showScenarios(features []*gherkin.Feature) {
...
+}
+
+// Should be deleted once https://issues.redhat.com/browse/KOGITO-1376 is solved
+func installDependentOperatorsIfNeeded(namespace string) error {
+	var operatorsToInstall []string
+	for operatorName := range framework.KogitoOperatorCommunityDependencies {
+		if installed, err := framework.IsCommunityOperatorCrdAvailable(operatorName); err != nil {
+			return err
+		} else if !installed {
+			operatorsToInstall = append(operatorsToInstall, operatorName)
```
ricardozanini: I understand this is a temporary workaround, so take into account if you want, but I'd say that you could install and wait for the dependency in the same loop instead.

sutaakar: AFAIK the intention is to speed up the test by allowing several dependencies to be installed in parallel

radtriste: exactly. the `waitFor` loop is blocking.
What I can do also to speed it up a bit is waiting for CRD to be installed, we do not care here that the operator pod is running. This should be quite quicker then.

radtriste: updated, should be a bit faster

test/framework/operator.go
```diff
@@ -216,3 +220,23 @@ func CreateNamespacedSubscriptionIfNotExist(namespace string, subscriptionName s
...
+		return false, fmt.Errorf("Unknown community operator %s", operatorName)
+	}
+	return IsCrdAvailable(operator.crdName)
+
+}
+
+// WaitForKogitoOperatorCrdAvailable waits for dependent operator main crd to be available
+func WaitForKogitoOperatorCrdAvailable(namespace, dependentOperator string) error {
+	if operatorInfo, exists := KogitoOperatorCommunityDependencies[dependentOperator]; exists {
+		return WaitFor(namespace, fmt.Sprintf("%s operator crd is available", dependentOperator), time.Minute*time.Duration(operatorInfo.timeoutInMin), func() (bool, error) {
```
MarianMacik: Do we want to incorporate the load factor also here?

test/main_test.go
```diff
@@ -187,3 +191,54 @@ func showScenarios(features []*gherkin.Feature) {
...
+		}
+	}
+
+	for _, operatorName := range operatorsToInstall {
+		if err := framework.InstallCommunityKogitoOperatorDependency(namespace, operatorName); err != nil {
```
MarianMacik: So here we will install all dependent operators even if they are not needed? Because in this step:

https://github.com/kiegroup/kogito-cloud-operator/blob/01238f460145419857c7985f843bbf246081ad74/test/steps/operator.go#L55

We only install operators which are needed. If this PR gets merged, then that part of the step is not needed anymore I guess? Because the newly added code will always make sure that all operators (thus all CRDs) are installed before running the scenario.

radtriste: We just want the crds to be installed as a workaround.
So we don't even wait for them to be running.
when deleting the namespace, they will be automatically deleted.
But for the Kogito operator to be running, we need all dependent operators' crds installed. If not, the Kogito operator will not be running, even if we don't need dependent operators...
That is why this quick fix. Will take less than 2 min but will avoid all tests to fail because of missing crds.

test/framework/operator.go
radtriste: will be when kogito-1338 is merged into master ;)

test/main_test.go
radtriste: in the future, KOGITO-1376 should solve this problem in Kogito Operator

MarianMacik: ah, forgot about the different namespace, sorry :)

test/main_test.go
ricardozanini: Inside this loop could have the second loop though.

radtriste: done :)

PR #213, opened 2020-03-13
[KOGITO-1439] Cucumber tests: Set CLI or CR tests via test option

radtriste:
https://issues.redhat.com/browse/KOGITO-1439

removed CR / CLI differenciation.
By default, it will try to use the CLI to deploy. When executed, CLI command is creating a CR to deploy on Openshift.

Use `make run-tests cr_deployment_only=true` to tell the tests that we want only to test via CR and not CLI (because, for example, CLI is not available)

Also solve https://issues.redhat.com/browse/KOGITO-1435 (code removal)

### DIFF
--- a/test/features/deploy_quarkus_service.feature
+++ b/test/features/deploy_quarkus_service.feature
@@ -7,7 +7,7 @@ Feature: Deploy quarkus service
   Scenario Outline: Deploy drools-quarkus-example service without persistence
     Given Kogito Operator is deployed
     
-    When "<installer>" deploy quarkus example service "ruleunit-quarkus-example" with native "<native>"
+    When Deploy quarkus example service "ruleunit-quarkus-example" with native <native>
 
     Then Kogito application "ruleunit-quarkus-example" has 1 pods running within <minutes> minutes
     And HTTP POST request on service "ruleunit-quarkus-example" is successful within 2 minutes with path "find-approved" and body:
@@ -41,36 +41,23 @@ Feature: Deploy quarkus service
           }]
         }
       """
-    
-    @cr
-    Examples: CR Non Native
-      | installer | native | minutes |
-      | CR        | false  | 10      |
 
-    @cr
+    @smoke  
+    Examples: Non Native
+      | native   | minutes |
+      | disabled | 10      |
+
     @native
-    Examples: CR Native
-      | installer | native | minutes |
-      | CR        | true   | 20      |
-
-    @smoke
-    @cli
-    Examples: CLI Non Native
-      | installer | native | minutes |
-      | CLI       | false  | 10      |
-
-    @cli
-    @native
-    Examples: CLI Native
-      | installer | native | minutes |
-      | CLI       | true   | 20      |
+    Examples: Native
+      | native  | minutes |
+      | enabled | 20      |
 
 #####
 
   @persistence
   Scenario Outline: Deploy jbpm-quarkus-example service with persistence
     Given Kogito Operator is deployed with Infinispan operator
-    And "<installer>" deploy quarkus example service "jbpm-quarkus-example" with native "<native>" and persistence
+    And Deploy quarkus example service "jbpm-quarkus-example" with native <native> and persistence
     And Kogito application "jbpm-quarkus-example" has 1 pods running within <minutes> minutes
     And HTTP GET request on service "jbpm-quarkus-example" with path "orders" is successful within 3 minutes
     And HTTP POST request on service "jbpm-quarkus-example" with path "orders" and body:
@@ -90,38 +77,24 @@ Feature: Deploy quarkus service
     
     Then HTTP GET request on service "jbpm-quarkus-example" with path "orders" should return an array of size 1 within 2 minutes
     
-    @cr
-    Examples: CR Non Native
-      | installer | native | minutes |
-      | CR        | false  | 10      |
+    Examples: Non Native
+      | native   | minutes |
+      | disabled | 10      |
 
-    @cr
     @native
-    Examples: CR Native
-      | installer | native | minutes |
-      | CR        | true   | 20      |
-
-    @cli
-    Examples: CLI Non Native
-      | installer | native | minutes |
-      | CLI       | false  | 10      |
-
-    @cli
-    @native
-    Examples: CLI Native
-      | installer | native | minutes |
-      | CLI       | true   | 20      |
+    Examples: Native
+      | native  | minutes |
+      | enabled | 20      |
 
 #####
 
   # Disabled as long as https://issues.redhat.com/browse/KOGITO-1163 and https://issues.redhat.com/browse/KOGITO-1166 is not solved
   @disabled
-  @cr
   @jobsservice
   Scenario Outline: Deploy timer-quarkus-example service with Jobs service
     Given Kogito Operator is deployed
-    And "CR" install Kogito Jobs Service with 1 replicas
-    And "CR" deploy quarkus example service "timer-quarkus-example" with native "<native>"
+    And Install Kogito Jobs Service with 1 replicas
+    And Deploy quarkus example service "timer-quarkus-example" with native <native>
     And Kogito application "timer-quarkus-example" has 1 pods running within <minutes> minutes
 
     When HTTP POST request on service "timer-quarkus-example" is successful within 2 minutes with path "timer" and body:
@@ -134,12 +107,12 @@ Feature: Deploy quarkus service
     And Kogito application "timer-quarkus-example" log contains text "After timer" within 1 minutes
 
     Examples: Non Native
-      | native | minutes |
-      | false  | 10      |
+      | native   | minutes |
+      | disabled | 10      |
 
     # Disabled as long as https://issues.redhat.com/browse/KOGITO-1179 is not solved
     @disabled
     @native
     Examples: Native
-      | native | minutes |
-      | true   | 20      |
+      | native  | minutes |
+      | enabled | 20      |
--- a/test/features/deploy_springboot_service.feature
+++ b/test/features/deploy_springboot_service.feature
@@ -4,33 +4,23 @@ Feature: Deploy spring boot service
   Background:
     Given Namespace is created
 
-  Scenario Outline: Deploy jbpm-springboot-example service without persistence
+  @smoke
+  Scenario: Deploy jbpm-springboot-example service without persistence
     Given Kogito Operator is deployed
     
-    When "<installer>" deploy spring boot example service "jbpm-springboot-example"
+    When Deploy spring boot example service "jbpm-springboot-example"
 
     Then Kogito application "jbpm-springboot-example" has 1 pods running within 10 minutes
     And HTTP GET request on service "jbpm-springboot-example" with path "orders" is successful within 2 minutes
-
-    @cr
-    Examples: CR
-      | installer |
-      | CR        |
-
-    @smoke
-    @cli
-    Examples: CLI
-      | installer |
-      | CLI       |
 
 #####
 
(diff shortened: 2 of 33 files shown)


test/features/deploy_quarkus_service.feature
```diff
@@ -7,40 +7,27 @@ Feature: Deploy quarkus service
...
 
     Then Kogito application "drools-quarkus-example" has 1 pods running within <minutes> minutes
     And HTTP GET request on service "drools-quarkus-example" with path "persons/all" is successful within 2 minutes
-    
-    @cr
-    Examples: CR Non Native
-      | installer | native | minutes |
-      | CR        | false  | 10      |
-
-    @cr
```
MarianMacik: Is it OK that we no longer test deployments via CRs?

spolti: I think the reason is to have it transparent, cli or cr, if there is no CLI set the cli option. is my understanding correct @radtriste ?

test/features/deploy_quarkus_service.feature
```diff
@@ -60,38 +47,24 @@ Feature: Deploy quarkus service
...
 
   # Disabled as long as https://issues.redhat.com/browse/KOGITO-1163 and https://issues.redhat.com/browse/KOGITO-1166 is not solved
   @disabled
-  @cr
   @jobsservice
   Scenario Outline: Deploy timer-quarkus-example service with Jobs service
     Given Kogito Operator is deployed
-    And "CR" install Kogito Jobs Service with 1 replicas
-    And "CR" deploy quarkus example service "timer-quarkus-example" with native "<native>"
+    And CR Install Kogito Jobs Service with 1 replicas
```
sutaakar: Is the CR needed in this scenario? I think it can be dropped too.

test/features/deploy_springboot_service.feature
```diff
@@ -50,25 +40,15 @@ Feature: Deploy spring boot service
...
 
   # Disabled as long as https://issues.redhat.com/browse/KOGITO-1163 and https://issues.redhat.com/browse/KOGITO-1166 is not solved
   @disabled
-  @cr
   @jobsservice
   Scenario: CR deploy timer-springboot-example service with Jobs service
     Given Kogito Operator is deployed
-    And "CR" install Kogito Jobs Service with 1 replicas
-    And "CR" deploy spring boot example service "timer-springboot-example"
+    And CR Install Kogito Jobs Service with 1 replicas
```
sutaakar: Same as above.

test/features/deploy_travel_agency.feature
```diff
@@ -1,13 +1,13 @@
...
 
   Background:
     Given Namespace is created
     And Kogito Operator is deployed with Infinispan and Kafka operators
-    And "CR" install Kogito Data Index with 1 replicas
-    And "CR" deploy service from example file "travelapp-kogito-travel-agency.yaml"
-    And "CR" deploy service from example file "travelapp-kogito-visas.yaml"
+    And CR Install Kogito Data Index with 1 replicas
+    And CR Deploy service from example file "travelapp-kogito-travel-agency.yaml"
+	And CR Deploy service from example file "travelapp-kogito-visas.yaml"
```
sutaakar:
```suggestion
    And CR Deploy service from example file "travelapp-kogito-visas.yaml"
```

test/features/install_dataindex.feature
```diff
@@ -18,25 +18,13 @@ Feature: Kogito Data Index
...
-
 #####
 
-  @cr
   @events
   @persistence
   Scenario Outline: Process instance events are stored in Data Index
-    Given "CR" install Kogito Data Index with 1 replicas
-    And "CR" deploy quarkus example service "jbpm-quarkus-example" with native "<native>" and persistence and events
+    Given CR Install Kogito Data Index with 1 replicas
```
sutaakar: Same as above.

test/framework/kogitoapp.go
```diff
@@ -94,7 +95,9 @@ func crDeployExample(namespace string, kogitoAppDeployment KogitoAppDeployment)
 
 	setupBuildImageStreams(kogitoApp)
 
-	// Add a resource request if native building
+	// Add a resource request if native building	// Add namespace for service discovery
+	// Can be removed once https://issues.redhat.com/browse/KOGITO-675 is done
+	appendNewEnvToKogitoApp(kogitoApp, "NAMESPACE", namespace)
```
sutaakar: Duplication of line 78 above. Please remove this.

test/features/deploy_quarkus_service.feature
radtriste: by default, if nothing is given, it will try to use the CLI to deploy.
CLI is anyway creating a CR to deploy.

`cr_deployment_only` is there to say to the tests that we want only to test via CR (because ,for example, CLI is not available)

going to update the description

test/main_test.go
```diff
@@ -60,12 +61,12 @@ func TestMain(m *testing.M) {
...
 	}
-	if framework.IsConfigShowScenarios() {
+	if config.IsShowScenarios() {
 		showScenarios(features)
 	}
 
-	if !framework.IsConfigDryRun() {
-		if matchingFeature(cliTag, features) {
+	if !config.IsDryRun() {
+		if !config.IsCrDeploymentOnly() && matchingFeature(cliTag, features) {
```
sutaakar: Is matchingFeature(cliTag, features) needed any more?
Do we still use cli tag anywhere?

radtriste: you are right. will remove that ...
we still however need to check that cli binary is there if we are not `cr_deployment_only`

test/steps/kogitoapp.go
```diff
@@ -30,13 +29,13 @@ var assist = assistdog.NewDefault()
...
 func registerKogitoAppSteps(s *godog.Suite, data *Data) {
 	// Deploy steps
-	s.Step(`^"([^"]*)" deploy quarkus example service "([^"]*)" with native "([^"]*)"$`, data.deployQuarkusExampleServiceWithNative)
-	s.Step(`^"([^"]*)" deploy quarkus example service "([^"]*)" with native "([^"]*)" and labels$`, data.deployQuarkusExampleServiceWithNativeAndLabels)
-	s.Step(`^"([^"]*)" deploy quarkus example service "([^"]*)" with native "([^"]*)" and persistence$`, data.deployQuarkusExampleServiceWithNativeAndPersistence)
-	s.Step(`^"([^"]*)" deploy quarkus example service "([^"]*)" with native "([^"]*)" and persistence and events$`, data.deployQuarkusExampleServiceWithNativeAndPersistenceAndEvents)
-	s.Step(`^"([^"]*)" deploy spring boot example service "([^"]*)"$`, data.deploySpringBootExampleService)
-	s.Step(`^"([^"]*)" deploy spring boot example service "([^"]*)" with persistence$`, data.deploySpringBootExampleServiceWithPersistence)
-	s.Step(`^"CR" deploy service from example file "([^"]*)"$`, data.deployServiceFromExampleFile)
+	s.Step(`^((?:CR|CLI)?)[\s]*Deploy quarkus example service "([^"]*)" with native ((?:enabled|disabled){1})$`, data.deployQuarkusExampleServiceWithNative)
```
sutaakar: What is the purpose of {1} and why do you use ?: to not capture enabled|disabled ?

radtriste: overthinking on my side ? ...

radtriste: alos not sure now that `^((?:CR|CLI)?)[\s]*` is needed anymore ...

sutaakar: Yes, the ((?:CR|CLI)?)[\s]* was my second question once we finish previous one :) 
I think that ((?:CR|CLI)?) is needed to express that CR|CLI is not mandatory and the result is captured in every case. AFAIK [\s]* could be dropped.

radtriste: as we don't mention CR or CLI anymore into feature files, I don't think we need to keep the regexp. The choice CR/CLI is done into the code itself.
Also `CR Deploy service from example file` could be directly `Deploy service from example file` as anyway it will be via CR.

sutaakar: Ok for me.

test/main_test.go
radtriste: in fact I just found that in features/olm, we have disabled test that are checking cli commands ...

radtriste: just set `||` instead of `&&` as we still have some cli tests

test/steps/kogitoapp.go
```diff
@@ -20,23 +20,21 @@ import (
...
-	s.Step(`^"([^"]*)" deploy quarkus example service "([^"]*)" with native "([^"]*)" and persistence and events$`, data.deployQuarkusExampleServiceWithNativeAndPersistenceAndEvents)
-	s.Step(`^"([^"]*)" deploy spring boot example service "([^"]*)"$`, data.deploySpringBootExampleService)
-	s.Step(`^"([^"]*)" deploy spring boot example service "([^"]*)" with persistence$`, data.deploySpringBootExampleServiceWithPersistence)
-	s.Step(`^"CR" deploy service from example file "([^"]*)"$`, data.deployServiceFromExampleFile)
+	s.Step(`^Deploy quarkus example service "([^"]*)" with native ((?:enabled|disabled))$`, data.deployQuarkusExampleServiceWithNative)
+	s.Step(`^Deploy quarkus example service "([^"]*)" with native ((?:enabled|disabled)) and labels$`, data.deployQuarkusExampleServiceWithNativeAndLabels)
+	s.Step(`^Deploy quarkus example service "([^"]*)" with native ((?:enabled|disabled)) and persistence$`, data.deployQuarkusExampleServiceWithNativeAndPersistence)
+	s.Step(`^Deploy quarkus example service "([^"]*)" with native ((?:enabled|disabled)) and persistence and events$`, data.deployQuarkusExampleServiceWithNativeAndPersistenceAndEvents)
```
MarianMacik: Do we need a non-capturing group inside the capturing one? What if it was just a one capturing group, i.e. `(enabled|disabled)`?

radtriste: yes we do in fact. just tested before without the double `(` and it does not catch the value.
seems that `(?:enabled|disabled)` is just the choice but does not catch the value to be used as method parameter
`((?:enabled|disabled))` does ;)

sutaakar: And why not to use `(enabled|disabled)` ?

MarianMacik: That's why I suggested `(enabled|disabled)` because `(?:enabled|disabled)` is not capturing so no wonder it doesn't capture anything :)

radtriste: yes better like that... will update

radtriste: updated

radtriste:
@MarianMacik @sutaakar can you have a final review ?

test/features/deploy_travel_agency.feature
```diff
@@ -1,13 +1,13 @@
...
 
   Background:
     Given Namespace is created
     And Kogito Operator is deployed with Infinispan and Kafka operators
-    And "CR" install Kogito Data Index with 1 replicas
-    And "CR" deploy service from example file "travelapp-kogito-travel-agency.yaml"
-    And "CR" deploy service from example file "travelapp-kogito-visas.yaml"
+    And Install Kogito Data Index with 1 replicas
+    And Deploy service from example file "travelapp-kogito-travel-agency.yaml"
+	And Deploy service from example file "travelapp-kogito-visas.yaml"
```
sutaakar:
```suggestion
    And Deploy service from example file "travelapp-kogito-visas.yaml"
```

PR #218, opened 2020-03-17
[KOGITO-1011] - Github Jenkins Integration for PR

msmagnanijr:
Signed-off-by: Mauricio Magnani <mmagnani@redhat.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,85 @@
+pipeline {
+    agent { label 'go'}
+    options {
+        buildDiscarder logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '10')
+        timeout(time: 90, unit: 'MINUTES')
+    }
+    environment {
+        WORKING_DIR = "/home/jenkins/go/src/github.com/kiegroup/kogito-cloud-operator/"
+        GOPATH = "/home/jenkins/go"
+        GOCACHE = "/home/jenkins/go/.cache/go-build"
+        GOROOT = "go env GOROOT"
+    }
+    stages {
+        stage('Clean Workspace') {
+            steps {
+                dir ("${WORKING_DIR}") {
+                    deleteDir()
+                }
+            }
+        }
+        stage('Clone repository') {
+            steps {
+              sh "mkdir -p ${WORKING_DIR}"
+              sh "git clone https://github.com/kiegroup/kogito-cloud-operator.git ${WORKING_DIR}"
+            }    
+        }
+        stage('Initialize') {
+            steps {
+                sh "set +x && oc login --token=\$(oc whoami -t) --server=${OPENSHIFT_API} --insecure-skip-tls-verify"
+            }
+        }
+        stage('Build Kogito Operator') {
+            steps {
+                dir ("${WORKING_DIR}") {
+                    sh """
+                        GO111MODULE=on 
+                        go get -u golang.org/x/lint/golint
+                        touch /etc/sub{u,g}id
+                        usermod --add-subuids 10000-75535 \$(whoami)
+                        usermod --add-subgids 10000-75535 \$(whoami)
+                        cat /etc/subuid
+                        cat /etc/subgid
+                        make image_builder=podman
+                    """
+                }
+            }
+            
+        }
+        stage('Build Kogito CLI') {
+            steps {
+                dir ("${WORKING_DIR}") {
+                    sh "make build-cli"
+                }
+            }
+        }
+        stage('Push Operator Image to Openshift Registry') {
+            steps {
+                dir ("${WORKING_DIR}") {
+                    sh """
+                        set +x && podman login -u jenkins -p \$(oc whoami -t) --tls-verify=false ${OPENSHIFT_REGISTRY}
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+                    sh """
+                        make run-smoke-tests operator_image=${OPENSHIFT_REGISTRY}/openshift/kogito-cloud-operator operator_tag=pr-\$(echo \${GIT_COMMIT} | cut -c1-7) maven_mirror=${MAVEN_MIRROR_REPOSITORY} concurrent=3
+                    """
+                }
+            }
+            post {
+                always {
+                    dir("${WORKING_DIR}") {
+                        archiveArtifacts artifacts: 'test/logs/**/*.log', allowEmptyArchive: true
+                        junit testResults: 'test/logs/**/junit.xml', allowEmptyResults: true
```
spolti: operator does not uses junit, can it be removed?

msmagnanijr: hey @radtriste what do you think? This was added to store those smoke test logs.

Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+            }
+        }
+        stage('Build Kogito Operator') {
+            steps {
+                dir ("${WORKING_DIR}") {
+                    sh """
+                        export GOPATH=/home/jenkins/go
+                        export GOCACHE=\${GOPATH}/.cache/go-build
+                        export GOROOT=`go env GOROOT`
+                        GO111MODULE=on
```
sutaakar: Is this still needed?
What GO version is used?

Jenkinsfile
sutaakar: the junit.xml file is generated by godog, I would keep it here

Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+                        podman push --tls-verify=false docker://${OPENSHIFT_REGISTRY}/openshift/kogito-cloud-operator:pr-\$(echo \${GIT_COMMIT} | cut -c1-7)
+                    """
+                }
+            }
+        }
+        stage('Running Smoke Testing') {
+            steps {
+                dir ("${WORKING_DIR}") {
+                    sh """
+                        make run-smoke-tests operator_image=${OPENSHIFT_REGISTRY}/openshift/kogito-cloud-operator operator_tag=pr-\$(echo \${GIT_COMMIT} | cut -c1-7) maven_mirror=${MAVEN_MIRROR_REPOSITORY} concurrent=3
```
sutaakar: From my personal experience there may be issues with running several concurrent tests with operator image pushed into internal OCP registry. If several tests started at the same time and pulled image from openshift namespace then sometimes the pull failed (with unauthorized error).
Have you experienced similar behaviour?

Jenkinsfile
ricardozanini: Right now we are on 1.13. This env was used to use the new go modules back in time, I guess we won't need it anymore since `mod` is default now, but we have to run some tests to make sure.

Jenkinsfile
ricardozanini: Related, I'm seeing some `cucumber-*` namespaces in the AWS instance. Side effect?

sutaakar: The namespaces should be deleted even in case of failure.
Maybe it is a leftover from manual testing?

Jenkinsfile
radtriste: environments variables may be put into `environment` section I think or directly into the running container.
For the gomodule, last time I checked, I needed it to install golint... worth to retry....

Jenkinsfile
msmagnanijr: It seems that sometimes the namespace is not removed ... I still haven't been able to identify what causes it.

radtriste: Could it be that you stop the pipeline manually ?

msmagnanijr: Make sense! I will test this.

msmagnanijr: @radtriste I guess you're right. If the pipeline is stopped manually, the namespace will not be removed.

Jenkinsfile
msmagnanijr: done!

Jenkinsfile
```diff
@@ -0,0 +1,92 @@
+pipeline {
+    agent { label 'go'}
+    options {
+        buildDiscarder logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '10')
+        timeout(time: 90, unit: 'MINUTES')
+    }
+    environment {
+        OPENSHIFT_API = "https://api.kogito.automation.rhmw.io:6443"
```
spolti: can we hide these info?
Can we create this envs elsewhere and reuse them like our internal jenkins does?

Jenkinsfile
radtriste: The `export` should not be needed as you set the variables as environment. They will be automatically available

msmagnanijr: ahh okay! I will fix that.

Jenkinsfile
msmagnanijr: Let me check that!

msmagnanijr: done!

Jenkinsfile
msmagnanijr: done!

Jenkinsfile
ricardozanini: please also make sure to clean up your commit so we won't have this history.

msmagnanijr: done!

PR #221, opened 2020-03-18
[KOGITO-1316] Travel Agency Openshift scenario

sutaakar:
Implemented scenario when Visa is not needed.
Other scenario will be delivered as another PR.

https://issues.redhat.com/browse/KOGITO-1316

### DIFF
--- a/test/features/deploy_travel_agency.feature
+++ b/test/features/deploy_travel_agency.feature
@@ -1,0 +1,44 @@
+@quarkus
+@cr
+Feature: Deploy Travel agency service and verify its functionality
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed with Infinispan and Kafka operators
+    And "CR" install Kogito Data Index with 1 replicas
+    And "CR" deploy service from example file "travelapp-kogito-travel-agency.yaml"
+    And Kogito application "kogito-travel-agency" has 1 pods running within 10 minutes
+    And HTTP GET request on service "kogito-travel-agency" with path "travels" is successful within 1 minutes
+
+  Scenario: Travel application without required Visa
+    When Start "travels" process on service "kogito-travel-agency" with body:
+      """json
+	{
+		"traveller" : {
+			"firstName" : "John6",
+			"lastName" : "Doe",
+			"email" : "john.doe@example.com",
+			"nationality" : "American",
+			"address" : {
+				"street" : "main street",
+				"city" : "Boston",
+				"zipCode" : "10005",
+				"country" : "US"
+			}
+		},
+		"trip" : {
+			"city" : "New York",
+			"country" : "US",
+			"begin" : "2019-12-10T00:00:00.000+02:00",
+			"end" : "2019-12-15T00:00:00.000+02:00"
+		}
+	}
+      """
+    And Service "kogito-travel-agency" contains 1 instance of process with name "travels"
+    And Service "kogito-travel-agency" contains 1 task of process with name "travels" and task name "ConfirmTravel"
+    And Complete "ConfirmTravel" task on service "kogito-travel-agency" and process with name "travels" with body:
+	  """json
+	  {}
+	  """
+
+    Then Service "kogito-travel-agency" contains 0 instances of process with name "travels"
--- a/test/framework/kogitoapp.go
+++ b/test/framework/kogitoapp.go
@@ -23,12 +23,15 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
+	"github.com/gobuffalo/packr/v2"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
 )
 
 const (
+	boxExamplesPath       = "../../deploy/examples"
 	mavenArgsAppendEnvVar = "MAVEN_ARGS_APPEND"
 	mavenMirrorURLEnvVar  = "MAVEN_MIRROR_URL"
 )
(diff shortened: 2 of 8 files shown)


test/framework/kogitoapp.go
```diff
@@ -137,6 +139,34 @@ func cliDeployExample(namespace string, kogitoAppDeployment KogitoAppDeployment)
...
+func DeployExampleServiceFromYaml(namespace, exampleServiceYaml string) error {
+	box := packr.New("examples", boxExamplesPath)
+	yamlContent, err := box.FindString(exampleServiceYaml)
+	if err != nil {
+		return fmt.Errorf("Error reading file %s: %v ", exampleServiceYaml, err)
+	}
+
+	// status.conditions cannot be omitted as conditions is an embedded struct, therefore it doesn't have default value
+	// As a workaround empty conditions have to be provided when creating Kogito CR.
+	kogitoApp := &v1alpha1.KogitoApp{
```
ricardozanini: I'd say you guys could have a util function to create those empty instances for you

ricardozanini: Or with some kind of parameter to customize the instance. I remember seeing this kind of code more than once.

sutaakar: Adjusted the implementation to use available KogitoApp stub.

test/features/deploy_travel_agency.feature
```diff
@@ -0,0 +1,44 @@
+@quarkus
+@cr
+Feature: Deploy Travel agency service and verify its functionality
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed with Infinispan and Kafka operators
+    And "CR" install Kogito Data Index with 1 replicas
+    And "CR" deploy "travelapp-kogito-travel-agency.yaml" example service
```
radtriste: maybe add `file` instead of `service` ? or add it ?

test/features/deploy_travel_agency.feature
```diff
@@ -0,0 +1,44 @@
...
+		},
+		"trip" : {
+			"city" : "New York",
+			"country" : "US",
+			"begin" : "2019-12-10T00:00:00.000+02:00",
+			"end" : "2019-12-15T00:00:00.000+02:00"
+		}
+	}
+      """
+    And Service "kogito-travel-agency" contains 1 instance of process with name "travels"
```
radtriste: just curiosity if we are clustering the kogito app, does it work automatically ? or should we wait a bit ?

test/steps/task.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// We assume that one process instance is available, can be refactored if more flexibility needed
+	foundProcessInstances, err := framework.GetProcessInstances(data.Namespace, routeURI, processName)
+	if err != nil {
+		return err
+	}
+	if len(foundProcessInstances) == 0 {
+		return fmt.Errorf("no process instance found, expected one instance")
+	}
+	if len(foundProcessInstances) > 1 {
```
radtriste: `else if` ?

test/steps/task.go
```diff
@@ -0,0 +1,103 @@
...
+	foundTasks, err := framework.GetTasks(data.Namespace, routeURI, processName, processInstanceID)
+	if err != nil {
+		return err
+	}
+	for _, foundTaskName := range foundTasks {
+		if taskName != foundTaskName {
+			return fmt.Errorf("found unexpected task name %s", foundTaskName)
+		}
+	}
+	if len(foundTasks) < numberOfTasks {
```
radtriste: could be done before the loop ?

test/steps/task.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// We assume that one process instance is available, can be refactored if more flexibility needed
+	foundProcessInstances, err := framework.GetProcessInstances(data.Namespace, routeURI, processName)
+	if err != nil {
+		return err
+	}
+	if len(foundProcessInstances) == 0 {
+		return fmt.Errorf("no process instance found, expected one instance")
+	}
+	if len(foundProcessInstances) > 1 {
```
radtriste: `else if`

test/steps/task.go
```diff
@@ -0,0 +1,103 @@
...
+	processInstanceID := foundProcessInstances[0]["id"].(string)
+	foundTasks, err := framework.GetTasks(data.Namespace, routeURI, processName, processInstanceID)
+	if err != nil {
+		return err
+	}
+
+	var taskID string
+	for foundTaskID, foundTaskName := range foundTasks {
+		if taskName == foundTaskName {
+			taskID = foundTaskID
```
radtriste: should we break the loop ? (just in case)

test/steps/task.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	err = framework.CompleteTask(data.Namespace, routeURI, processName, processInstanceID, taskName, taskID, body.ContentType, body.Content)
```
radtriste: what happens if no `taskID` found ?

radtriste:
test ran ok

test/framework/kogitoapp.go
```diff
@@ -24,11 +24,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/gobuffalo/packr/v2"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
 )
 
 const (
+	boxExamplesPath       = "../../deploy/examples"
```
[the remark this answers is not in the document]
sutaakar: Well, my idea is that the files in examples folder should be usable out of the box by users, that they should just simply work. And our tests should make sure that they works.

Maybe we can mark such tests, so they can be ran separately?

test/steps/task.go
sutaakar: Not needed, previous if returns from function.

test/steps/task.go
sutaakar: Not needed, previous if returns from function.

test/features/deploy_travel_agency.feature
sutaakar: Well, the purpose is to deploy service from yaml which is located in examples folder.

So, maybe:
"CR" deploy service from example file "travelapp-kogito-travel-agency.yaml"

What do you think about it?

test/features/deploy_travel_agency.feature
sutaakar: What do you mean by clustering the kogito app? Like having several process instances?

test/steps/task.go
sutaakar: Well, I prefer it this way.
Let me show an example - what will happen if we expect one task of name AAA, but task request return two tasks - AAA and BBB.
In case we would check number of tasks first then we get message saying that there are two tasks instead of one. But it doesn't tell me much about why there are two tasks. Is the expected task duplicated? Or some other tasks is created?
In current implementation you get an error saying that unexpected task named BBB was found, which AFAIK provides more information than previous approach.

Ideally we should compare expected and actual result and construct complex error showing what do we expect and what we got, but I wanted to keep things simple.

test/steps/task.go
sutaakar: sure

test/steps/task.go
sutaakar: good point, will adjust the implementation

test/steps/task.go
radtriste: right...

test/steps/task.go
radtriste: good point. you win :)

test/steps/task.go
radtriste: oki

test/features/deploy_travel_agency.feature
radtriste: yes, we have many instances of same app

test/features/deploy_travel_agency.feature
radtriste: I do like that one :)

test/framework/kogitoapp.go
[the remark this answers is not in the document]
radtriste: We can keep them as they are. but we need to make sure build & services (data-index, jobs-service) version are not set as we need to be able to override them in our tests. (only the git reference could be set).
if you take the `onboarding.yaml`, there is there `imageVersion` which is set. and we would have no way in our tests to override it ...
and I just noiticed that is the case as well in the file `travelapp-kogito-travel-agency.yaml`: https://github.com/sutaakar/kogito-cloud-operator/blob/cd3fdf18332030395d862563092f9251b7847e26/deploy/examples/travelapp-kogito-travel-agency.yaml#L19

test/features/deploy_travel_agency.feature
sutaakar: Well, I would expect the process starting to be a synchronized action (so when Kogito returns response then process flow reached "save point"). So all the process info should be stored in infinispan already.
So I would expect that informations should be available without waiting.

test/framework/kogitoapp.go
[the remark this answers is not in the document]
sutaakar: Done, now the image version is overwritten once the example is loaded.

sutaakar:
@radtriste Please have another look.

PR #223, opened 2020-03-18
[KOGITO-1499] Cucumber tests: migrate to godog 0.9.0

radtriste:
https://issues.redhat.com/browse/KOGITO-1499

For now we are with godog 0.9.0-rc2. There is a big change of Gherkin structure as they are migrating to cucumber Gherkin: https://github.com/cucumber/cucumber/tree/master/gherkin
We need to wait anyway that godog release before merging this one.

Not yet tested !!!!!!

- [x] Set new godog version & migrate Gherkin code
- [ ] ~~Refactor the way we parse feature into main_test.go, maybe not needed anymore~~


radtriste:
tests ran ok

README.md
```diff
@@ -1071,6 +1071,11 @@ $ make run-tests [key=value]*
 
 You can set those optional keys:
 
+<!--- tests configuration -->
```
Sgitario, line 1074: Good refactor !

test/feature.go
```diff
@@ -26,18 +26,24 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/cucumber/godog/gherkin"
+	"github.com/cucumber/gherkin-go/v11"
+	"github.com/cucumber/messages-go/v10"
 )
 
-func parseFeatures(filter string, paths []string) ([]*gherkin.Feature, error) {
+type feature struct {
```
Sgitario, line 33: Should the type name be capitalised? Since this type will be used in other scripts, I would expect to be named "Feature" instead.

test/feature.go
```diff
@@ -71,31 +77,38 @@ func parsePath(path string) ([]*gherkin.Feature, error) {
...
+func parseFeatureFile(path string, newIDFunc func() string) (*feature, error) {
 	reader, err := os.Open(path)
 	if err != nil {
 		return nil, err
 	}
 	defer reader.Close()
 
 	var buf bytes.Buffer
-	ft, err := gherkin.ParseFeature(io.TeeReader(reader, &buf))
+	ft, err := gherkin.ParseGherkinDocument(io.TeeReader(reader, &buf), newIDFunc)
```
Sgitario, line 97: What is the point of sending the newIDFunc here? Can we create this function variable within this method before calling to gherkin.ParseGherkingDocument and gherkin.Pickles?

As far I can understand, the "newIDFunc" function is a Gherkin thing, so if we isolate all the Gherkin stuff in this method, the better.

test/feature.go
```diff
@@ -50,17 +56,17 @@ func parseFeatures(filter string, paths []string) ([]*gherkin.Feature, error) {
...
-			features[ft.Name] = ft
+			features[ft.document.Feature.GetName()] = ft
 		}
 	}
 	return filterFeatures(filter, features), nil
 }
 
-func parsePath(path string) ([]*gherkin.Feature, error) {
-	var features []*gherkin.Feature
+func parsePath(path string) ([]*feature, error) {
```
Sgitario, line 68: Should this method be renamed to parseFeaturesFromPath to make clear the purpose?

test/feature.go
```diff
@@ -71,31 +77,38 @@ func parsePath(path string) ([]*gherkin.Feature, error) {
...
+	ft, err := parseFeatureFile(path, newIDFunc)
 	if err != nil {
 		return features, err
 	}
 
 	return append(features, ft), nil
 }
 
-func parseFeatureFile(path string) (*gherkin.Feature, error) {
+func parseFeatureFile(path string, newIDFunc func() string) (*feature, error) {
```
Sgitario, line 89: The path argument is a filename here, does it make sense to rename it as well?

test/feature.go
```diff
@@ -127,39 +140,22 @@ func filterFeatures(tags string, collected map[string]*gherkin.Feature) (feature
 	return features
 }
 
-func applyTagFilter(tags string, ft *gherkin.Feature) {
+func applyTagFilter(tags string, ft *feature) {
```
Sgitario, line 143: Since we're going to modify the ft instance, does it make sense to scope the method to the Feature type?

```go
func (ft *Feature) applyTagFilter(tags string)
```

Also, the "tags" argument makes me thing that this is a list of tag and I think this is instead a filter (a string containing && tokens). Also, the "filter" name matches with the below method "matchesTags". What do you think about?

test/feature.go
```diff
@@ -177,47 +173,13 @@ func matchesTags(filter string, tags []string) (ok bool) {
 	return
 }
 
-func hasTag(tags []string, tag string) bool {
+func hasTag(tags []*messages.Pickle_PickleTag, tag string) bool {
```
Sgitario, line 176: For consistency with the above methods, should we make the tags argument the last parameter?

```go
func hasTag(tag string, tags []*messages.Pickle_PickleTag) bool {
```

test/framework/util.go
```diff
@@ -103,24 +103,25 @@ func WaitFor(namespace, display string, timeout time.Duration, condition func()
...
-	for _ , examples := range outline.Examples {
-		examplesName := examples.Name
-		if len(examples.Name) == 0 {
-			examplesName = "Unnamed Examples"
-		}
-		names = append(names, examplesName)
-	}
-	return
-}
+// // GetScenarioName retrieves the name of the scenario
```
Sgitario, line 106: What is usually your approach about commented blocks like this? I normally remove the commented blocks as we already have the source repository to track changes.

test/main_test.go
```diff
@@ -175,28 +175,38 @@ func deleteNamespaceIfExists(namespace string) {
...
-		if len(getMatchingScenarios(tags, ft)) > 0 {
+		if matchesPickles(tags, ft.pickles) {
 			return true
 		}
 	}
 	return false
 }
 
-func showScenarios(features []*gherkin.Feature) {
+func matchesPickles(tags string, pickles []*messages.Pickle) bool {
```
Sgitario, line 187: According to my above comment, the tags argument should be renamed to "filter" instead?

test/feature.go
radtriste: this is used only in package

about `feature.go`, I just copied from godog source to have the possibility to retrieve the features which will be run.
I don't want to make toomuch changed yet as this will be deleted once https://github.com/cucumber/godog/issues/222 is done and released

test/feature.go
radtriste: as said, I just copied and did not much touch this code ...

test/feature.go
radtriste: as said, I just copied and did not much touch this code ...

test/feature.go
radtriste: as said, I just copied and did not much touch this code ...

test/framework/util.go
radtriste: that I should delete indeed

test/main_test.go
radtriste: I will change this

radtriste: modified

test/framework/util.go
radtriste: deleted

test/features/kogito_service_performance.feature
```diff
@@ -42,18 +42,13 @@ Feature: Kogito Service Performance
 #      | disabled | 10      | 320000   |
 
     @native
-    Examples: Native
```
sutaakar, line 45: Accidentally removed?

test/steps/data.go
```diff
@@ -93,26 +93,25 @@ func (data *Data) AfterScenario(s interface{}, err error) {
...
 
-	handleScenarioResult(data, s, err)
-	logScenarioDuration(data, s)
+	handleScenarioResult(data, pickle, err)
+	logScenarioDuration(data, pickle)
 	deleteTemporaryExamplesFolder(data)
 }
 
-func logScenarioDuration(data *Data, s interface{}) {
+func logScenarioDuration(data *Data, pickle *messages.Pickle) {
```
sutaakar, line 101: seems that pickle parameter can be removed here, it doesn't seem to be used.

test/framework/util.go
```diff
@@ -189,3 +170,17 @@ func MustParseEnabledDisabled(value string) bool {
 		panic(fmt.Errorf("Unknown value for enabled/disabled: %s", value))
 	}
 }
+
+// GetStringMapFromDatatable parse the given datatable to a key/value string map
+func GetStringMapFromDatatable(dt *messages.PickleStepArgument_PickleTable) (map[string]string, error) {
```
sutaakar, line 175: This method doesn't seem to be used.

sutaakar:
Few minor points, otherwise looks good.

test/features/kogito_service_performance.feature
radtriste: good catch. Might have been to fast to try to delete the `Native` string and deleted the rest ...

test/steps/data.go
radtriste: deleted

test/framework/util.go
radtriste: Nice catch. I did that before merging with latest changes from Marian. will delete it

radtriste:
I still need to make sure performance tests are ok

radtriste:
Perf tests ran ok

test/features/kogito_service_performance.feature
```diff
@@ -10,7 +10,7 @@ Feature: Kogito Service Performance
     Given Namespace is created
 
   @quarkus
-  Scenario Outline: Quarkus Kogito Service Performance without persistence
+  Scenario Outline: Quarkus Kogito Service Performance with requests <requests>, native <native> but without persistence
```
MarianMacik, line 13: Didn't know you can basically parameterize everything in the Scenario. It seems you can really use it everywhere :)

test/main_test.go
```diff
@@ -175,28 +175,38 @@ func deleteNamespaceIfExists(namespace string) {
...
 	mainLogger.Info("------------------ SHOW SCENARIOS ------------------")
 	for _, ft := range features {
-		if len(ft.ScenarioDefinitions) > 0 {
-			mainLogger.Infof("Feature: %s", ft.Name)
-			for _, scenario := range ft.ScenarioDefinitions {
-				scenarioMessage := framework.GetScenarioName(scenario)
-
-				if scenarioOutline, ok := scenario.(*gherkin.ScenarioOutline); ok {
-					scenarioMessage += fmt.Sprintf(" (Examples: %s)", strings.Join(framework.GetExamplesNames(scenarioOutline), ", "))
+		// Placeholders in names are now replace directly into names for each pickle
```
MarianMacik, line 200: Ah, so it is a new feature in the end :)

BTW `are now replaceD`

test/features/kogito_service_performance.feature
radtriste: Yeah it is with the new structure, they replace the placeholder(s) in title now also :)
Because each scenariooutline/example is now parsed as a pickle, whereas before it was still the feature structure

test/main_test.go
radtriste: thx. updated

PR #234, opened 2020-03-24
[KOGITO-1193] Mount Config Maps for Kogito Services and Runtimes application properties configuration

xiezhang7:
Jira: https://issues.redhat.com/browse/KOGITO-1193

Description: Quarkus and Springboot support application.properties override based on a directory in the file system. We could leverage from this to mount ConfigMaps to the deployed Kogito Services (either provided or built) for users to fine tune their configuration, not having to always rely on environment variables for that.

Implementation: Calculate the content hash of the `application.properties` of the config map in the reconcile loop. Store the content hash as an annotation of the deployment of the services. When the content hash changes, the deployment will rolling update.

Test: Only tested with quarkus for kogito runtime service and data index service. Springboot and Jobs service should work in the same way. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Table of Contents
          * [Binary Builds](#binary-builds)
          * [Cleaning up a Kogito service deployment](#cleaning-up-a-kogito-service-deployment)
          * [Native X JVM builds](#native-x-jvm-builds)
+         * [Kogito Runtimes properties configuration](#kogito-runtimes-properties-configuration)
          * [Troubleshooting Kogito Runtimes service deployment](#troubleshooting-kogito-runtimes-service-deployment)
             * [No builds are running](#no-builds-are-running)
       * [Kogito Data Index Service deployment](#kogito-data-index-service-deployment)
@@ -30,12 +31,14 @@ Table of Contents
             * [Installing the Kogito Data Index Service with the Operator Catalog (OLM)](#installing-the-kogito-data-index-service-with-the-operator-catalog-olm)
             * [Installing the Kogito Data Index Service with the oc client](#installing-the-kogito-data-index-service-with-the-oc-client)
          * [Kogito Data Index Integration with persistent Kogito Services](#kogito-data-index-integration-with-persistent-kogito-services)
+         * [Kogito Data Index Service properties configuration](#kogito-data-index-service-properties-configuration)
       * [Kogito Jobs Service deployment](#kogito-jobs-service-deployment)
          * [Kogito Jobs Service installation](#kogito-jobs-service-installation)
             * [Installing the Kogito Jobs Service with the Kogito CLI](#installing-the-kogito-jobs-service-with-the-kogito-cli)
             * [Installing the Kogito Jobs Service with the Operator Catalog (OLM)](#installing-the-kogito-jobs-service-with-the-operator-catalog-olm)
             * [Installing the Kogito Jobs Service with the oc client](#installing-the-kogito-jobs-service-with-the-oc-client)
          * [Enabling Persistence with Infinispan](#enabling-persistence-with-infinispan)
+         * [Kogito Jobs Service properties configuration](#kogito-jobs-service-properties-configuration)
       * [Kogito CLI](#kogito-cli)
          * [Kogito CLI requirements](#kogito-cli-requirements)
          * [Kogito CLI installation](#kogito-cli-installation)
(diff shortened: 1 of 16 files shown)


pkg/infrastructure/services/configmap.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return contentHash, configMap, nil
+}
+
+// GetAppPropConfigMapName generates the name of the config map for application.properties
+func GetAppPropConfigMapName(name string) string {
+	return name + appPropConfigMapSuffix
+}
+
+// ExcludeAppPropConfigMapFromResource excludes the applicatio properties config map from the resources
```
ricardozanini:
```suggestion
// ExcludeAppPropConfigMapFromResource excludes the application properties config map from the resources
```

pkg/infrastructure/services/deployment.go
```diff
@@ -26,9 +26,18 @@ const (
...
+	// AppPropVolumeName is the name of the volume for application.properties
+	AppPropVolumeName = "app-prop-config"
+
+	appPropDefaultMode = int32(420)
+	appPropFileName    = "application.properties"
+	appPropFilePath    = "/home/kogito/config"
 )
 
-func createRequiredDeployment(service v1alpha1.KogitoService, resolvedImage string, definition ServiceDefinition) *appsv1.Deployment {
+func createRequiredDeployment(service v1alpha1.KogitoService, resolvedImage string, appPropContentHash string, definition ServiceDefinition) *appsv1.Deployment {
```
ricardozanini: Please, instead of adding the configMap here, I'd say to leave it to the `resources.go` file to do it. This function supposed to return only the **raw** `Deployment` object and then [we use other functions to decorate it](https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/infrastructure/services/resources.go#L88), as we are doing when injecting Kafka or Infinispan variables. The support functions could go into `configmap.go` or in the decorator itself.

pkg/infrastructure/services/resources.go
```diff
@@ -46,12 +46,21 @@ func (s *serviceDeployer) createRequiredResources(instance v1alpha1.KogitoServic
 		resources[reflect.TypeOf(imgv1.ImageStream{})] = []resource.KubernetesResource{imageHandler.imageStream}
 	}
 
+	contentHash, configMap, err := GetAppPropConfigMapContentHash(instance.GetName(), instance.GetNamespace(), s.client)
```
ricardozanini: When you created the decorator, this will go after the Deployment creation, since is highly dependant on it. If we don't have a image ready yet, won't make sense creating the CM.

pkg/infrastructure/services/configmap.go
ricardozanini: Please also consider expanding this doc to explain that this function removes the configMap from kubernetes resources that need to be compared, ignoring the ConfigMap. I took sometime to figure that out. xD

xiezhang7:
@ricardozanini , I have made the changes you suggested. Please take a look. Thanks.

pkg/infrastructure/services/resources.go
```diff
@@ -185,7 +212,14 @@ func (s *serviceDeployer) getDeployedResources(instance v1alpha1.KogitoService)
...
 		objectTypes = append(objectTypes, &kafkabetav1.KafkaTopicList{})
 	}
-	return reader.ListAll(objectTypes...)
+
+	resources, err = reader.ListAll(objectTypes...)
+	if err != nil {
+		return
+	}
+
+	ExcludeAppPropConfigMapFromResource(instance.GetName(), resources)
```
ricardozanini: If `configMap` is not in the  `objectTypes` list we still need this function? I understand that if we have other kind of configMaps that we control we will need to remove only the properties one, which makes sense. But in this case that we don't have the `operator-utils` will fetch it either way because of the owner reference?

xiezhang7: The `configmaps` won't be fetched because it is not in the `objectTypes` list. Added this code here, just in case we will introduce other `configmaps` in future. With or without this line of code makes no difference for now. I can remove it if you think we should take care of it in future.

ricardozanini: Let's leave it there since we won't have a configMap in the list, thus the function will return before entering in the loop. Won't be a problem IMO.

radtriste:
Also do we plan to migrate the env variables to the application.properties ?

radtriste:
sorry, wrong manip ...

xiezhang7:
> See my comments on readme.
> 
> Cucumber tests went through

Thanks @radtriste , I have updated the readme as you suggested.

pkg/controller/kogitoapp/resource/deployed_resources.go
```diff
@@ -57,6 +58,8 @@ func GetDeployedResources(instance *v1alpha1.KogitoApp, client *client.Client) (
 		log.Warn("Failed to list deployed objects. ", err)
 		return nil, err
 	}
+
+	services.ExcludeAppPropConfigMapFromResource(instance.GetName(), resourceMap)
```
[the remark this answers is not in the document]
ricardozanini: see my comment above regarding this discussion

pkg/infrastructure/services/configmap.go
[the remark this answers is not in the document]
xiezhang7: It is already fixed.

README.md
```diff
@@ -406,6 +430,27 @@ property in your domain data, this data will be reflect automatically in the Dat
 _Please note that removed Kogito Services will remove the protobuf files associated to it as well. This means that you won't
 be able to see the data through the Data Index anymore, although the data still persisted in Infinispan._
 
+### Kogito Data Index Application Properties Configuration
```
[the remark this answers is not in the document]
xiezhang7: I think it is better to add a section for each of the affected services based on the current structure of the readme. It is easier to understand for the readers.

I agree we should remove the duplicated contents once we restructure the contents of the readme, for example we can create a parent section to describe the deployment of all the services and move the deployment of the individual services as child sections.

PR #239, opened 2020-03-26
[KOGITO-1370] Add Cucumber tests for binary build

sutaakar:
https://issues.redhat.com/browse/KOGITO-1370

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -47,6 +47,8 @@ import (
 
 	monv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	monclientv1 "github.com/coreos/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
+
+	apibuildv1 "github.com/openshift/api/build/v1"
 )
 
 const (
@@ -204,6 +206,7 @@ func newControllerCliOptions() controllercli.Options {
 	mapper.Add(olmapiv1alpha1.SchemeGroupVersion.WithKind(meta.KindSubscription.Name), &restScope{name: apimeta.RESTScopeNameNamespace})
 	mapper.Add(monv1.SchemeGroupVersion.WithKind(meta.KindPrometheus.Name), &restScope{name: apimeta.RESTScopeNameNamespace})
 	mapper.Add(corev1.SchemeGroupVersion.WithKind(meta.KindPod.Name), &restScope{name: apimeta.RESTScopeNameNamespace})
+	mapper.Add(apibuildv1.SchemeGroupVersion.WithKind(meta.KindBuildConfig.Name), &restScope{name: apimeta.RESTScopeNameNamespace})
 
 	// the kube client is having problems with plural: kogitodataindexs :(
 	mapper.AddSpecific(v1alpha1.SchemeGroupVersion.WithKind(meta.KindKogitoDataIndex.Name),
--- a/test/features/binary_build.feature
+++ b/test/features/binary_build.feature
@@ -1,0 +1,19 @@
+@binary
+Feature: Use binary build to deploy the service
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+
+  @springboot
+  # jbpm-springboot-example is used here as it is the quickest example to be built
+  Scenario: Deploy jbpm-springboot-example using binary build
+    Given Clone Kogito examples into local directory
+    And Local example service "jbpm-springboot-example" is built by Maven
+
+    When Create service "jbpm-springboot-example"
+    And BuildConfig "jbpm-springboot-example-binary" is created after 1 minutes
+    And Start build with name "jbpm-springboot-example-binary" from local example service path "jbpm-springboot-example/target"
+
+    Then Kogito application "jbpm-springboot-example" has 1 pods running within 5 minutes
+    And HTTP GET request on service "jbpm-springboot-example" with path "orders" is successful within 2 minutes
--- a/test/framework/command.go
+++ b/test/framework/command.go
@@ -1,0 +1,46 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import "os/exec"
+
+// ExecuteCommand Execute command in current directory
+func ExecuteCommand(commandName string, args ...string) (string, error) {
+	return ExecuteCommandInDirectory("", commandName, args...)
+}
+
+// ExecuteCommandInDirectory Execute command in defined directory
+func ExecuteCommandInDirectory(directory, commandName string, args ...string) (string, error) {
+	if len(directory) == 0 {
+		GetMainLogger().Infof("Execute command %s %v", commandName, args)
+	} else {
+		GetMainLogger().Infof("Execute command %s %v in directory %s", commandName, args, directory)
+	}
+
+	command := exec.Command(commandName, args...)
+	command.Dir = directory
+	out, err := command.Output()
+
+	if err != nil {
+		GetMainLogger().Errorf("output command: %s", string(out[:]))
+		if ee, ok := err.(*exec.ExitError); ok {
+			GetMainLogger().Errorf("error output command: %s", string(ee.Stderr))
+		}
+	} else {
+		GetMainLogger().Debugf("output command: %s", string(out[:]))
+	}
+
+	return string(out[:]), err
+}
--- a/test/framework/kogitoapp.go
+++ b/test/framework/kogitoapp.go
@@ -41,37 +41,47 @@ const (
 
 // KogitoAppDeployment defines the elements for the deployment of a kogito application
 type KogitoAppDeployment struct {
-	AppName     string
-	ContextDir  string
-	Runtime     v1alpha1.RuntimeType
-	Native      bool
-	Persistence bool
-	Events      bool
-	Labels      map[string]string
+	AppName            string
+	GitSourceURI       string
+	GitSourceReference string
+	ContextDir         string
+	Runtime            v1alpha1.RuntimeType
+	Native             bool
+	Persistence        bool
+	Events             bool
+	Labels             map[string]string
 }
 
-// DeployExample deploy a Kogito example
-func DeployExample(namespace string, installerType InstallerType, kogitoAppDeployment KogitoAppDeployment) error {
+// DeployService deploy a Kogito service
+func DeployService(namespace string, installerType InstallerType, kogitoAppDeployment KogitoAppDeployment) error {
 	GetLogger(namespace).Infof("%s deploy %s example %s with name %s, native %v, persistence %v, events %v and labels %v", installerType, kogitoAppDeployment.Runtime, kogitoAppDeployment.ContextDir, kogitoAppDeployment.AppName, kogitoAppDeployment.Native, kogitoAppDeployment.Persistence, kogitoAppDeployment.Events, kogitoAppDeployment.Labels)
 	switch installerType {
 	case CLIInstallerType:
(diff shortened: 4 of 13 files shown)


sutaakar:
To be able to run binary build test you need to reference local examples folder and have oc client and maven installed locally.

Makefile
```diff
@@ -99,6 +99,7 @@ build_runtime_image_tag=
 # examples repository
 examples_uri=
 examples_ref=
+examples_path=
```
radtriste: Can't we reuse the `examples_uri` and `examples_ref` and use Git from a golang library ?

sutaakar: Will take a look on it.

test/framework/command.go
```diff
@@ -0,0 +1,47 @@
...
+// ExecuteCommandInDirectory Execute command in defined directory
+func ExecuteCommandInDirectory(directory, commandName string, args ...string) (string, error) {
+	if len(directory) == 0 {
+		GetMainLogger().Infof("Execute command %s %v", commandName, args)
+	} else {
+		GetMainLogger().Infof("Execute command %s %v in directory %s", commandName, args, directory)
+	}
+
+	command := exec.Command(commandName, args...)
+	command.Dir = directory
```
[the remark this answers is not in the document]
sutaakar: well, that doesn't change the functionality and makes the code a bit more complex

test/framework/command.go
```diff
@@ -0,0 +1,47 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import "os/exec"
+
+// ExecuteCommand Execute command in current directory
```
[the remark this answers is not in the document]
sutaakar: Yes, I want to address it in separate PR to keep this PR smaller and focused just on new functionality.

test/framework/command.go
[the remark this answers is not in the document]
radtriste: oki

test/framework/command.go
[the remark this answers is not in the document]
radtriste: oki :)

ricardozanini:
I'll get back to review this one once all @radtriste's comments were addressed.

sutaakar:
@radtriste Please take another look, I hope I addressed all your comments.

test/steps/kogitoapp.go
```diff
@@ -68,77 +68,90 @@ func (data *Data) deployQuarkusExampleServiceWithNativeAndLabels(contextDir, nat
...
 }
 
-func (data *Data) deployServiceFromExampleFile(exampleFile string) error {
-	return framework.DeployServiceFromExampleFile(data.Namespace, exampleFile)
+func (data *Data) createService(serviceName string) error {
+	return framework.DeployService(data.Namespace, framework.GetDefaultInstallerType(),
+		framework.KogitoAppDeployment{
+			AppName: serviceName,
+			Runtime: v1alpha1.SpringbootRuntimeType,
+		})
```
MarianMacik: Just if I get the binary build right...This will just deploy the custom resource of the service without building it as it misses the git URL etc.? After that the operator will create a BuildConfig resource which will be bound with the actual start of build by the build name? Thanks!

test/features/binary_build.feature
```diff
@@ -0,0 +1,18 @@
+@binary
+Feature: Use binary build to deploy the service
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+
+  @springboot
+  Scenario: Deploy jbpm-springboot-example using binary build
```
MarianMacik: Do we support only springboot for binary build?

test/steps/kogitoapp.go
sutaakar: yes, exactly.

test/features/binary_build.feature
sutaakar: We support quarkus too.
For both quarkus and springboot the process will be same (upload built project to OpenShift), so I test it just on one example. I chose springboot because it is faster.

radtriste: maybe add a small comment why this choice so we don't have that question again in 6 months XD

test/framework/kogitoapp.go
```diff
@@ -41,37 +41,44 @@ const (
...
 	kogitoApp.Spec.Build.Native = kogitoAppDeployment.Native
-	kogitoApp.Spec.Build.GitSource.URI = config.GetExamplesRepositoryURI()
-	kogitoApp.Spec.Build.GitSource.ContextDir = kogitoAppDeployment.ContextDir
-	kogitoApp.Spec.Build.GitSource.Reference = config.GetExamplesRepositoryRef()
+
+	if gitSourceURI := kogitoAppDeployment.GitSourceURI; len(gitSourceURI) > 0 {
+		kogitoApp.Spec.Build.GitSource.URI = gitSourceURI
+		kogitoApp.Spec.Build.GitSource.ContextDir = kogitoAppDeployment.ContextDir
+
+		if ref := kogitoAppDeployment.GitSourceReference; len(ref) > 0 {
```
radtriste: `kogitoApp.Spec.Build.GitSource.Reference` is with `omitempty` configuration so don't need to do this check

test/framework/kogitoapp.go
```diff
@@ -109,17 +116,22 @@ func crDeployExample(namespace string, kogitoAppDeployment KogitoAppDeployment)
...
 }
 
-func cliDeployExample(namespace string, kogitoAppDeployment KogitoAppDeployment) error {
-	cmd := []string{"deploy-service", kogitoAppDeployment.AppName, config.GetExamplesRepositoryURI()}
+func cliDeployService(namespace string, kogitoAppDeployment KogitoAppDeployment) error {
+	cmd := []string{"deploy-service", kogitoAppDeployment.AppName}
+
+	if gitSourceURI := kogitoAppDeployment.GitSourceURI; len(gitSourceURI) > 0 {
+		cmd = append(cmd, gitSourceURI)
+		cmd = append(cmd, "-c", kogitoAppDeployment.ContextDir)
```
radtriste: do we suppose there is always a `ContextDir` ? I am just thinking about that now ...

test/steps/data.go
```diff
@@ -62,6 +69,14 @@ func getNamespaceName() string {
 	return framework.GenerateNamespaceName("cucumber")
 }
 
+func createTemporaryFolder() string {
```
radtriste: wouldn't be better to have that one in `util.go` from framework ?

test/features/binary_build.feature
```diff
@@ -0,0 +1,18 @@
...
+    Given Clone Kogito examples into temporary directory
```
radtriste: would use `local` instead of temporary as we have then step `Local example service` wdyt ?

test/steps/kogitoapp.go
```diff
@@ -51,15 +50,16 @@ func registerKogitoAppSteps(s *godog.Suite, data *Data) {
...
 		framework.KogitoAppDeployment{
-			AppName:     filepath.Base(contextDir),
-			ContextDir:  contextDir,
-			Runtime:     v1alpha1.QuarkusRuntimeType,
-			Native:      framework.MustParseEnabledDisabled(native),
-			Persistence: false,
-			Events:      false,
-			Labels:      nil,
+			AppName:      filepath.Base(contextDir),
+			GitSourceURI: config.GetExamplesRepositoryURI(),
```
radtriste: missing `GitSourceReference: config.GetExamplesRepositoryRef` or ?

test/steps/git.go
```diff
@@ -0,0 +1,61 @@
...
+	} else {
+		// Try cloning as branch reference
+		cloneOptions.ReferenceName = plumbing.NewBranchReferenceName(reference)
+		err = cloneExamples(data.KogitoExamplesLocation, cloneOptions)
+		if err == nil {
+			return err
+		}
+
+		// If branch cloning failed then try cloning as tag
+		cloneOptions.ReferenceName = plumbing.NewTagReferenceName(reference)
```
radtriste: we arrive until here ?
if the clone from branch fails, it will return an error that will be returned just before or ?

test/framework/kogitoapp.go
sutaakar: Actually I need to do the check.
Because reference can be empty, therefore default value is applied - empty string.

radtriste: weird... I thought it would be ignored

test/framework/kogitoapp.go
sutaakar: Good point, will ad an empty check there too.

test/steps/data.go
sutaakar: ok

test/steps/kogitoapp.go
sutaakar: yes, I forgot to add it there

test/steps/git.go
sutaakar: No, actually we return from branch clone just in case there is no exception, will add a comment there to make it clearer.

radtriste: ahaha. yeah. too much used to return err if not null XD

test/framework/kogitoapp.go
sutaakar: it can't be ignored as it is not a pointer but structure field
omitempty serves as a hint for marshaller to not marshall the field if it contains default value

sutaakar:
@radtriste Ready for another round of review.

test/framework/kogitoapp.go
radtriste: now I get it :)

ricardozanini: @sutaakar please implement a function on `KogitoAppSpec` like `IsGitSourceEmpty` and do the checks there. We can reuse it and the code will be clearer.

test/steps/kogitoapp.go
```diff
@@ -68,77 +69,90 @@ func (data *Data) deployQuarkusExampleServiceWithNativeAndLabels(contextDir, nat
...
+			Persistence:        false,
+			Events:             false,
+			Labels:             nil,
 		})
 }
 
 func (data *Data) deploySpringBootExampleServiceWithPersistence(contextDir string) error {
-	return framework.DeployExample(data.Namespace, framework.GetDefaultInstallerType(),
+	return framework.DeployService(data.Namespace, framework.GetDefaultInstallerType(),
 		framework.KogitoAppDeployment{
```
ricardozanini: You guys recreated the `KogitoApp` structure in the framework? Why?

sutaakar: That was mostly prepared by Tristan, but AFAIK it serves as a value holder for both CR and CLI deployments.

ricardozanini: Understood. Well, I think you could reconsider using the `KogitoApp` instead since it's basically the same thing. Could be a hell to maintain and to keep up with the API. On the other hand, I understand that a wrapper could help in such scenarios.

sutaakar: Hmm, I think it could have sense.
@radtriste What do you think about it?

radtriste: indeed ... we could use the KogitoApp struct directly.
I was trying to figure out if there would be anything extra we need, but we can still give as a parameter :)

sutaakar:
Considering enhancements based on https://github.com/kiegroup/kogito-cloud-operator/pull/239#discussion_r399238165 I think we can postpone merging of this PR after 0.9.0 release, not to block it.

This affects just tests anyway.

ricardozanini:
> Considering enhancements based on [#239 (comment)](https://github.com/kiegroup/kogito-cloud-operator/pull/239#discussion_r399238165) I think we can postpone merging of this PR after 0.9.0 release, not to block it.
> 
> This affects just tests anyway.

@sutaakar I'd say to merge and open a JIRA to follow up this refactoring

sutaakar:
Created https://issues.redhat.com/browse/KOGITO-1597 to track the refactoring

pkg/client/client.go
```diff
@@ -47,6 +47,8 @@ import (
 
 	monv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	monclientv1 "github.com/coreos/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
+
+	apibuildv1 "github.com/openshift/api/build/v1"
```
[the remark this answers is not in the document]
sutaakar: because buildv1 is already imported on https://github.com/kiegroup/kogito-cloud-operator/blob/11e7f8df37b6d74ebda655b6df1120403b24df70/pkg/client/client.go#L45

PR #247, opened 2020-03-28
[KOGITO-256] -Kogito CLI to upload source files from local machine to…

spolti:
… build s2i runtimes
See: https://issues.redhat.com/browse/KOGITO-256

Signed-off-by: spolti <fspolti@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Table of Contents
       * [Kogito Runtimes service deployment](#kogito-runtimes-service-deployment)
          * [Deploying a new service](#deploying-a-new-service)
          * [Binary Builds](#binary-builds)
+         * [Build From File](#build-from-file)
          * [Cleaning up a Kogito service deployment](#cleaning-up-a-kogito-service-deployment)
          * [Native X JVM builds](#native-x-jvm-builds)
          * [Kogito Runtimes properties configuration](#kogito-runtimes-properties-configuration)
@@ -145,7 +146,7 @@ kogitoapp.app.kiegroup.org/example-quarkus created
 Alternatively, you can use the [Kogito CLI](#kogito-cli) to deploy your services:
 
 ```bash
-$ kogito deploy-service example-quarkus https://github.com/kiegroup/kogito-examples/ --context-dir=drools-quarkus-example
+$ kogito deploy-service example-quarkus https://github.com/kiegroup/kogito-examples/ --context-dir=rules-quarkus-helloworld
 ```
 
 ### Binary Builds
@@ -196,6 +197,85 @@ $ oc start-build example-quarkus-binary --from-dir=target
 ``` 
 And that's it. In a couple minutes you should have your Kogito Service application up and running using the same binaries
 built locally.
+
+### Build From File
+
+Kogito Operator can configure a build for you that will pull and build your application through the source to image (s2i)
+feature (available only on OpenShift). This kind of build consists to upload a Kogito service file, e.g. a dmn or a bpmn file
+to the OpenShift Cluster and trigger a new Source to Image build automatically.
+
+You can provide a single dmn, drl, bpmn, bpmn2 and properties files, the standalone file can be locally on the filesystem or on a Web Site, like GitHub:
+
+```bash
+# from filesystem
+$ kogito deploy-service example-dmn-quarkus /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/"Traffic Violation.dmn"
+File found /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/Traffic Violation.dmn.
+...
+The requested file(s) was successfully uploaded to OpenShift, a build with this file(s) should now be running. To see the logs, run 'oc logs -f bc/example-dmn-quarkus-builder -n kogito'
+```
+
+While checking OpenShift build logs with this command **oc logs -f bc/example-dmn-quarkus-builder -n kogito** you can see on the beginning ot the build log:
+```
+Receiving source from STDIN as file TrafficViolation.dmn
+Using docker-registry.default.svc:5000/openshift/kogito-quarkus-ubi8-s2i@sha256:729e158710dedba50a49943ba188d8f31d09568634896de9b903838fc4e34e94 as the s2i builder image
+```
+
+```bash
+# from a git repository
+$ kogito deploy-service example-dmn-quarkus https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
+Asset found: TrafficViolation.dmn.
+...
+The requested file(s) was successfully uploaded to OpenShift, a build with this file(s) should now be running. To see the logs, run 'oc logs -f bc/example-dmn-quarkus-builder -n kogito'
+```
+
+If you check the logs, we can also see the dmn file being used to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. In that case, 
+all we need to do is to specify a whole directory. If any valid file is found in the given directory, the CLI will 
+compress these files and upload to the OpenShift cluster. To upload more than one file, put all the desired files in a 
+directory and specify it as parameter. 
+Note that if you have other unsupported files in this directory, they will not be copied. For such complex cases you can use
+either [build from source](#deploying-a-new-service) or a [binary build](#binary-builds).
+
+After a build is created and for some reason we need to update the Kogito Assets we can do it with the `oc start-build`
+command, but keep in mind that a s2i build is not able to identify what files has been changed and update only this one,
+All files must be provided again. Suppose we have create a Kogito Application with the following command:
+
+```bash
+$ kogito deploy-service example-dmn-quarkus-builder /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/"Traffic Violation.dmn"
+Uploading file "/tmp/kogito-examples/dmn-quarkus-example/src/main/resources/Traffic Violation.dmn" as binary input for the build ...
+.
+Uploading finished
+build.build.openshift.io/example-dmn-quarkus-builder-2 started
+
+```
+
+We can provide the files again, after we're done updating it, to do this use the following command:
+
+```bash
+$ oc start-build example-dmn-quarkus-builder --from-file /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/"Traffic Violation.dmn"
+```
+
+If a directory was provided, just update the `--from-file` flag to `--from-dir`.
+
+
+```bash
+kogito deploy-service test /home/user/development/kogito-test/
+```
+
+As output of the cli command above, we will have something similar to:
+
+```bash
+The provided source is a dir, packing files.
+File(s) found: [/home/user/development/kogito-test/Traffic Violation.dmn /home/user/development/kogito-test/application.properties].
+...
+The requested file(s) was successfully uploaded to OpenShift, a build with this file(s) should now be running. To see the logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If for some reason the build fails, pass the env [BUILD_LOGLEVEL](https://docs.openshift.com/container-platform/4.3/builds/basic-build-operations.html#builds-basic-access-build-verbosity_basic-build-operations) with the desired level, e.g. 9 as a build-env parameter:
+```bash
+kogito --verbose deploy-service test /home/user/development/kogito-test/  --build-env BUILD_LOGLEVEL=5
+```
 
 ### Cleaning up a Kogito service deployment
 
--- a/cmd/kogito/command/deploy/deploy_service.go
+++ b/cmd/kogito/command/deploy/deploy_service.go
@@ -17,23 +17,33 @@ package deploy
 import (
 	"fmt"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
+	kogitoerror "github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/error"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/message"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/shared"
+	buildutil "github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/util"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/openshift"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/util"
+	buildv1 "github.com/openshift/api/build/v1"
+	"io"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"net/http"
 	"net/url"
-
-	"github.com/kiegroup/kogito-cloud-operator/pkg/util"
+	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
 
 const defaultDeployRuntime = string(v1alpha1.QuarkusRuntimeType)
 
-var deployRuntimeValidEntries = []string{string(v1alpha1.QuarkusRuntimeType), string(v1alpha1.SpringbootRuntimeType)}
+var (
+	deployRuntimeValidEntries = []string{string(v1alpha1.QuarkusRuntimeType), string(v1alpha1.SpringbootRuntimeType)}
+)
 
 type deployFlags struct {
 	CommonFlags
@@ -82,12 +92,15 @@ func (i *deployCommand) RegisterHook() {
 		Short:   "Deploys a new Kogito Runtime Service into the given Project",
 		Aliases: []string{"deploy"},
 		Long: `deploy-service will create a new Kogito Runtime Service in the Project context. 
-        If the source is provided, the build will take place on the cluster, otherwise you must upload the application binaries via "oc start-build [NAME-binary] --from-dir=target". 
-		Project context is the namespace (Kubernetes) or project (OpenShift) where the Service will be deployed.
-		To know what's your context, use "kogito project". To set a new Project in the context use "kogito use-project NAME".
+	If the [SOURCE] is provided, the build will take place on the cluster.
+	If not, you can also provide a dmn/drl/bpmn/bpmn2 file or a directory containing one or more of those files, using the --from-file
+	Or you can also later upload directly the application binaries via "oc start-build [NAME-binary] --from-dir=target
+			
+	Project context is the namespace (Kubernetes) or project (OpenShift) where the Service will be deployed.
+	To know what's your context, use "kogito project". To set a new Project in the context use "kogito use-project NAME".
 
-		Please note that this command requires the Kogito Operator installed in the cluster.
-		For more information about the Kogito Operator installation please refer to https://github.com/kiegroup/kogito-cloud-operator#kogito-operator-installation.
+	Please note that this command requires the Kogito Operator installed in the cluster.
+	For more information about the Kogito Operator installation please refer to https://github.com/kiegroup/kogito-cloud-operator#kogito-operator-installation.
 		`,
 		RunE:    i.Exec,
 		PreRun:  i.CommonPreRun,
@@ -96,11 +109,6 @@ func (i *deployCommand) RegisterHook() {
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) > 2 {
 				return fmt.Errorf("requires 1 arg, received %v", len(args))
-			}
-			if len(args) > 1 {
-				if _, err := url.ParseRequestURI(args[1]); err != nil {
-					return fmt.Errorf("source is not a valid URL, received %s", args[1])
-				}
 			}
 			if err := util.ParseStringsForKeyPair(i.flags.buildEnv); err != nil {
 				return fmt.Errorf("build environment variables are in the wrong format. Valid are key pairs like 'env=value', received %s", i.flags.buildEnv)
@@ -157,9 +165,33 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
 	log := context.GetDefaultLogger()
 	i.flags.name = args[0]
 	hasSource := false
+	options := &buildv1.BinaryBuildRequestOptions{}
+	var userProvidedFile io.Reader
+
 	if len(args) > 1 {
-		i.flags.source = args[1]
-		hasSource = true
+		userProvidedFile, options, err = matchKogitoResources(args[1])
+		if err != nil {
+			switch t := err.(type) {
+			case *os.PathError:
+				return t
+
+			case *kogitoerror.KogitoAssetFileBuildError:
+				return t
+
+			case *url.Error:
+				return t
+
+			default:
+				log.Warnf(message.KogitoAppAssetNotSupported, args[1])
+			}
+
+		}
+		// if the parameter does not match the expected input to perform a build from file, try a build from source instead
+		if userProvidedFile == nil {
+			log.Info("Trying to perform a build from source...")
+			i.flags.source = args[1]
+			hasSource = true
+		}
 	}
 
 	if len(i.flags.mavenMirrorURL) > 0 {
@@ -249,6 +281,21 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
 	if hasSource {
 		log.Infof(message.KogitoAppViewDeploymentStatus, i.flags.name, i.flags.Project)
 		log.Infof(message.KogitoAppViewBuildStatus, i.flags.name, i.flags.Project)
+
+	} else if userProvidedFile != nil {
+
+		options.Name = kogitoApp.Name
+		cli, err := client.NewClientBuilder().WithBuildClient().Build()
+		if err != nil {
+			return err
+		}
+
+		_, err = openshift.BuildConfigC(cli).TriggerBuildFromFile(i.flags.Project, userProvidedFile, options)
+		if err != nil {
+			return err
+		}
+
+		log.Infof(message.KogitoAppSuccessfullyUploadedFile, i.flags.name, i.flags.Project)
 	} else {
 		log.Infof(message.KogitoAppUploadBinariesInstruction, i.flags.name, i.flags.Project)
 	}
@@ -264,4 +311,70 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	return nil
+}
+
+// matchKogitoResources will verify if the given parameter matches one of the allowed extensions
+// Accepted values are, supported file extensions as it is, directory or url
+// See supportedExtensions env
+// returns the file content as io.Reader if the provided file or url is valid.
+func matchKogitoResources(resource string) (io.Reader, *buildv1.BinaryBuildRequestOptions, error) {
+	log := context.GetDefaultLogger()
+	options := &buildv1.BinaryBuildRequestOptions{}
+	var fileName string
+	// from url, do only basic check, for example, if the url has the suffix that's match the allowed ones
+	switch {
+	case strings.HasPrefix(resource, "http"):
+		parsedURL, err := url.ParseRequestURI(resource)
+		if err != nil || parsedURL.Scheme == "" || parsedURL.Host == "" {
+			return nil, nil, &url.Error{
+				URL: resource,
+				Err: err,
+			}
+		}
+
+		ff := strings.Split(parsedURL.Path, "/")
+		fileName = strings.Join(strings.Fields(ff[len(ff)-1]), "")
+
+		if buildutil.IsSuffixSupported(fileName) {
+			response, err := http.Get(resource)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to download %s, error message: %s", resource, err.Error())
+			}
+			options.AsFile = fileName
+			log.Infof(message.KogitoAppFoundAsset, fileName)
+			return response.Body, options, nil
+		}
+		return nil, nil, nil
+
+	// handle single file, files on dir.
+	default:
+		fileInfo, err := os.Stat(resource)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if fileInfo.Mode().IsRegular() {
+			if buildutil.IsSuffixSupported(resource) {
+				log.Infof(message.KogitoAppFoundFile, resource)
+				ff := strings.Split(resource, "/")
+				fileName = strings.Join(strings.Fields(ff[len(ff)-1]), "")
+				fileReader, err := os.Open(resource)
+				if err != nil {
+					return nil, nil, err
+				}
+				options.AsFile = fileName
+				return fileReader, options, nil
+			}
+
+		} else if fileInfo.Mode().IsDir() {
+			log.Info(message.KogitoAppProvidedFileIsDir)
+
+			ioTgzR, err := buildutil.ProduceTGZfile(resource)
+			if err != nil {
+				return nil, nil, err
+			}
+			return ioTgzR, options, nil
+		}
+		return nil, nil, &kogitoerror.KogitoAssetFileBuildError{Asset: resource}
+	}
 }
--- a/cmd/kogito/command/deploy/deploy_service_test.go
+++ b/cmd/kogito/command/deploy/deploy_service_test.go
@@ -20,7 +20,11 @@ import (
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/test"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
+	"io/ioutil"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"log"
+	"net/url"
+	"os"
 	"sort"
 	"strings"
 	"testing"
@@ -176,6 +180,68 @@ func Test_DeployCmd_WrongGitURL(t *testing.T) {
 		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
 	_, _, err := test.ExecuteCli()
 	assert.Error(t, err)
+}
+
+func Test_DeployCmd_WithSingleKogitoResourceFileWithInvalidFile(t *testing.T) {
+	ns := t.Name()
+	cli := fmt.Sprintf("deploy-service example-from-file my-file.dmn -p %s", ns)
+	test.SetupCliTest(cli,
+		context.CommandFactory{BuildCommands: BuildCommands},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}},
+		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
+	lines, _, err := test.ExecuteCli()
+	assert.Error(t, err)
+	assert.Contains(t, lines, "Error: stat my-file.dmn: no such file or directory")
+}
+
+func Test_DeployCmd_WithSingleKogitoResourceFileWithValidFileFromUrl(t *testing.T) {
+	ns := t.Name()
+	fileLocation, _ := url.Parse("https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic Violation.dmn")
+	cli := fmt.Sprintf("deploy-service example-from-file %s -p %s", fileLocation, ns)
+	test.SetupCliTest(cli,
+		context.CommandFactory{BuildCommands: BuildCommands},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}},
+		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
+	lines, _, err := test.ExecuteCli()
+	assert.Contains(t, lines, "successfully deployed")
+	assert.Contains(t, lines, "successfully created")
+	// expect error at this point as there is no valid cluster to do a post request. (connect: no route to host)
+	assert.Error(t, err)
+	assert.NotContains(t, lines, "Error: stat TrafficViolation.dmn: no such file or directory")
+}
+
+func Test_DeployCmd_WithSingleKogitoResourceFileWithInvalidFileFromUrl(t *testing.T) {
+	ns := t.Name()
+	fileLocation, _ := url.Parse("https:ss//raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic Violation.dmnsd")
+	cli := fmt.Sprintf("deploy-service example-from-file %s -p %s", fileLocation, ns)
+	test.SetupCliTest(cli,
+		context.CommandFactory{BuildCommands: BuildCommands},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}},
+		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
+	lines, _, err := test.ExecuteCli()
+	assert.NotContains(t, lines, "successfully deployed")
+	assert.NotContains(t, lines, "successfully created")
+	// expect error at this point as there is no valid cluster to do a post request. (connect: no route to host)
+	assert.Error(t, err)
+	assert.NotContains(t, lines, "Error:  https:ss//raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic Violation.dmnsd")
+}
+
+func Test_DeployCmd_WithSingleKogitoResourceFileWithInvalidFileType(t *testing.T) {
+	ns := t.Name()
+	invalidFile, err := ioutil.TempFile(os.TempDir(), "invalidFile.txt")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer invalidFile.Close()
+	cli := fmt.Sprintf("deploy-service example-from-file %s -p %s", invalidFile.Name(), ns)
+	test.SetupCliTest(cli,
+		context.CommandFactory{BuildCommands: BuildCommands},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}},
+		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
+	lines, _, err := test.ExecuteCli()
+	assert.Error(t, err)
+	assert.Contains(t, lines, "unable to process the requested file "+invalidFile.Name())
+	os.Remove(invalidFile.Name())
 }
 
 //see: https://issues.redhat.com/browse/KOGITO-1431
--- a/cmd/kogito/command/error/kogitoapp_errors.go
+++ b/cmd/kogito/command/error/kogitoapp_errors.go
@@ -1,0 +1,28 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package error
+
+import (
+	"fmt"
+)
+
+// KogitoAssetFileBuildError Custom error for handling build from file errors.
+type KogitoAssetFileBuildError struct {
+	Asset string
+}
+
+func (k *KogitoAssetFileBuildError) Error() string {
+	return fmt.Sprintf("unable to process the requested file %s", k.Asset)
+}
--- a/cmd/kogito/command/message/kogitoapp.go
+++ b/cmd/kogito/command/message/kogitoapp.go
@@ -20,9 +20,13 @@ const (
 	KogitoAppViewDeploymentStatus      = "You can see the deployment status by using 'oc describe kogitoapp %s -n %s'"
 	KogitoAppViewBuildStatus           = "Your Kogito Runtime Service should be deploying. To see its logs, run 'oc logs -f bc/%s-builder -n %s'"
 	KogitoAppUploadBinariesInstruction = "Your Kogito Runtime Service needs the application binaries to proceed. To upload your binaries please run 'oc start-build %s-binary --from-dir=target -n %s' from your project's root"
-
-	KogitoAppNoMgmtConsole = `To more easily manage your Kogito Runtime Service install Data Index Service and Process Instance Management. 
+	KogitoAppNoMgmtConsole             = `To more easily manage your Kogito Runtime Service install Data Index Service and Process Instance Management. 
 For how to install see: https://github.com/kiegroup/kogito-cloud-operator#kogito-data-index-service-deployment
 and https://github.com/kiegroup/kogito-cloud-operator#kogito-management-console-install`
-	KogitoAppMgmtConsoleEndpoint = `You can manage your process using the management console: %s`
+	KogitoAppMgmtConsoleEndpoint      = `You can manage your process using the management console: %s`
+	KogitoAppSuccessfullyUploadedFile = "The requested file(s) was successfully uploaded to OpenShift, a build with this file(s) should now be running. To see the logs, run 'oc logs -f bc/%s-builder -n %s'"
+	KogitoAppAssetNotSupported        = "The provided resource is not a standalone Kogito asset."
+	KogitoAppFoundFile                = "File(s) found: %s."
+	KogitoAppFoundAsset               = "Asset found: %s."
+	KogitoAppProvidedFileIsDir        = "The provided source is a directory, packing files."
 )
--- a/cmd/kogito/command/util/build_from_file.go
+++ b/cmd/kogito/command/util/build_from_file.go
@@ -1,0 +1,100 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/message"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var supportedExtensions = []string{".dmn", ".drl", ".bpmn", ".bpmn2", ".properties"}
+
+// ProduceTGZfile produces a tgz file of the given directory files
+func ProduceTGZfile(resource string) (io.Reader, error) {
+	log := context.GetDefaultLogger()
+	var buf bytes.Buffer
+	var filesFound []string
+	var link string
+
+	gzipWriter := gzip.NewWriter(&buf)
+	defer gzipWriter.Close()
+
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
+
+	filepath.Walk(resource, func(absoluteFilePath string, fileInfo os.FileInfo, err error) error {
+		if IsSuffixSupported(fileInfo.Name()) {
+
+			fileToCompress, err := os.Open(absoluteFilePath)
+			if err != nil {
+				return err
+			}
+			defer fileToCompress.Close()
+
+			stat, err := fileToCompress.Stat()
+			if err != nil {
+				return err
+			}
+
+			if fileInfo.Mode()&os.ModeSymlink != 0 {
+				link, err = os.Readlink(absoluteFilePath)
+				if err != nil {
+					return err
+				}
+			}
+			header, err := tar.FileInfoHeader(fileInfo, link)
+			if err != nil {
+				return err
+			}
+			header.Name = filepath.ToSlash(fileInfo.Name())
+			header.Linkname = filepath.ToSlash(header.Linkname)
+			header.Format = tar.FormatPAX
+			header.Size = stat.Size()
+			header.Mode = int64(stat.Mode())
+			header.ModTime = stat.ModTime()
+			// write header
+			if err := tarWriter.WriteHeader(header); err != nil {
+				return err
+			}
+
+			if _, err := io.Copy(tarWriter, fileToCompress); err != nil {
+				return err
+			}
+
+			filesFound = append(filesFound, fileToCompress.Name())
+		}
+		return nil
+	})
+	log.Infof(message.KogitoAppFoundFile, filesFound)
+	return &buf, nil
+}
+
+// IsSuffixSupported checks if the given extension is supported
+// when performing a build from single file
+func IsSuffixSupported(value string) bool {
+	for _, ext := range supportedExtensions {
+		if strings.HasSuffix(value, ext) {
+			return true
+		}
+	}
+	return false
+}
--- a/cmd/kogito/command/util/build_from_file_test.go
+++ b/cmd/kogito/command/util/build_from_file_test.go
@@ -1,0 +1,123 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"encoding/hex"
+	"github.com/stretchr/testify/assert"
+	"io"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestProduceTGZfile(t *testing.T) {
+	tempDir := "/tmp/kogito-test/"
+	simpleContent := []byte("hello World!!")
+	files := []string{"file.bpmn", "file2.drl", "file3.bpmn2", "file4.dmn", "file5.properties", "file6.unsupported"}
+	_, err := os.Stat(tempDir)
+	if os.IsNotExist(err) {
+		errDir := os.MkdirAll(tempDir, 0755)
+		if errDir != nil {
+			panic(err)
+		}
+	}
+	for _, file := range files {
+		err := ioutil.WriteFile(tempDir+file, simpleContent, 0755)
+		fileInfo, err := os.Stat(tempDir + file)
+		// double check file exists
+		assert.Nil(t, err)
+		assert.True(t, fileInfo.Size() > 0)
+	}
+
+	ioR, err := ProduceTGZfile(tempDir)
+	assert.Nil(t, err)
+	fileToWrite, err1 := ioutil.TempFile(os.TempDir(), "compressed_kogito_resources_*.tgz")
+	assert.Nil(t, err1)
+
+	if _, err2 := io.Copy(fileToWrite, ioR); err2 != nil {
+		panic(err2)
+	}
+	defer fileToWrite.Close()
+
+	// test if file is tgz and if the tgz contains the right files.
+	// is Gzip?
+	testGzip, err := os.Open(fileToWrite.Name())
+	assert.Nil(t, err)
+	defer testGzip.Close()
+
+	// see https://mimesniff.spec.whatwg.org/
+	// gzip 3 first bytes should be 1F 8B 08
+	buff := make([]byte, 3)
+	_, err = testGzip.Read(buff)
+	assert.Nil(t, err)
+	assert.Equal(t, hex.Dump(buff), "00000000  1f 8b 08                                          |...|\n")
+
+	// extract file and see if it is tarball file
+	testTar, err := os.Open(fileToWrite.Name())
+	assert.Nil(t, err)
+
+	gzf, err := gzip.NewReader(testTar)
+	assert.Nil(t, err)
+
+	tarHeader := tar.NewReader(gzf)
+
+	var filesFromGzip []string
+	for true {
+		header, err := tarHeader.Next()
+		if err == io.EOF {
+			break
+		}
+		assert.Equal(t, header.Format, tar.FormatPAX)
+		assert.NotNil(t, header.Name)
+		filesFromGzip = append(filesFromGzip, header.Name)
+	}
+
+	// all files extracted from gzip must be on the initial slice of test files
+	for _, fileFromGzip := range filesFromGzip {
+		assert.True(t, contains(fileFromGzip, files))
+	}
+
+	// And filesFromGzip should not contain the unsupported file: file6.unsupported
+	assert.False(t, contains("file6.unsupported", filesFromGzip))
+
+	os.RemoveAll(tempDir)
+	os.Remove(fileToWrite.Name())
+}
+
+func TestIsSuffixSupported(t *testing.T) {
+	//test a few invalid extensions
+	invalidExtensions := []string{"test.dn", "test.st", "test.www", "test.bmn2", "test.prop"}
+	for _, ext := range invalidExtensions {
+		assert.False(t, IsSuffixSupported(ext))
+	}
+
+	//test a few valid extensions
+	validExtensions := []string{"test.drl", "test.dmn", "test.bpmn2", "test.properties", "test.bpmn"}
+	for _, ext := range validExtensions {
+		assert.True(t, IsSuffixSupported(ext))
+	}
+}
+
+func contains(item string, slice []string) bool {
+	for _, value := range slice {
+		if value == item {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/client/openshift/buildconfig.go
+++ b/pkg/client/openshift/buildconfig.go
@@ -17,15 +17,24 @@ package openshift
 import (
 	"fmt"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
+	"io"
+	"k8s.io/apimachinery/pkg/runtime"
 	"strings"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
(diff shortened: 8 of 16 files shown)


cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+			// produce gzip
+			if err := gzipWriter.Close(); err != nil {
+				return nil, nil, err
+			}
+			tarWriter.Flush()
+
+			var r io.Reader
+			br := bufio.NewReaderSize(&buf, 4096)
+			r = br
+			fileToWrite, err1 := os.OpenFile(filepath.Join("/tmp", "compressed_kogito_resources.tgz"), os.O_CREATE|os.O_RDWR, os.FileMode(777))
```
spolti: remind for later, use tmp dir from go env.

pkg/client/client.go
```diff
@@ -43,6 +43,7 @@ import (
 	ocappsv1 "github.com/openshift/api/apps/v1"
 	ocroutev1 "github.com/openshift/api/route/v1"
 	buildv1 "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
+	//ocbuildv1 "github.com/openshift/api/build/v1"
```
spolti: Leftover, need to remove.

pkg/client/openshift/buildconfig.go
```diff
@@ -38,6 +42,7 @@ type BuildState struct {
 type BuildConfigInterface interface {
 	EnsureImageBuild(bc *buildv1.BuildConfig, labelSelector, imageName string) (BuildState, error)
 	TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (bool, error)
+	TriggerFromFileBuild(namespace string, r io.Reader, options *buildv1.BinaryBuildRequestOptions) (*buildv1.Build, error)
```
spolti: Rename triggerBuildFromFile

pkg/client/openshift/buildconfig.go
```diff
@@ -101,6 +106,38 @@ func (b *buildConfig) TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (b
...
+// When called a new build will be triggered with the request kogito resource or a tgz file.
+func (b *buildConfig) TriggerFromFileBuild(namespace string, r io.Reader, options *buildv1.BinaryBuildRequestOptions) (*buildv1.Build, error) {
+
+	result := &buildv1.Build{}
+	buildName := fmt.Sprintf("%s%s", options.Name, "-builder")
+
+	// before upload the file, make sure that the build exist
+	err := b.waitForBuildConfig(5, 2*time.Second, func() (err error) {
+		if bc, err := b.client.BuildCli.BuildConfigs(namespace).Get(buildName, metav1.GetOptions{}); err != nil && errors.IsNotFound(err) {
+			log.Debugf("", bc.Namespace)
```
spolti: Remove this debug

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+# from a git repository
+$ kogito -service example-dmn-quarkus --from-file https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. In that case all we need to do
+is, instead specify a directory instead, if any valid file is found on the given directory, the cli will compress these
```
sutaakar: > if any valid file is found on the given directory

So, only files with specific suffix are copied?

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+files and upload to the OpenShift cluster. To start upload various files, put all the desired files in a directory and 
+specify it as parameter. Note that if you have more files on this directory, they wont be copied.
```
sutaakar: > Note that if you have more files on this directory, they wont be copied.

What does it mean? From previous sentences I understood that more files can be uploaded at once.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -165,15 +175,31 @@ func (i *deployCommand) InitHook() {
...
 	if len(args) > 1 {
-		i.flags.source = args[1]
-		hasSource = true
+		// if build is from file
+		if i.flags.fromFile {
+			userProvidedFile, options, err = matchKogitoResources(args[1])
+			if err != nil {
+				return err
+			}
+			// otherwise configure a source build
```
sutaakar:
```suggestion
		// otherwise configure a source build
```

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+
+		hasExtension := false
+		ff := strings.Split(parsedURL.Path, "/")
+		fileName = strings.Join(strings.Fields(ff[len(ff)-1]), "")
+
+		if isSuffixSupported(fileName) {
+			hasExtension = true
+		}
+
+		if hasExtension {
```
sutaakar: The hasExtension can be removed and content of this condition can be moved under `if isSuffixSupported(fileName) {`

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+			response, err := http.Get(resource)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to download %s, error message: %s", resource, err.Error())
+			}
+			options.AsFile = fileName
+			return response.Body, options, nil
```
sutaakar: Is the body reader closed anywhere?

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+
+			// produce tar
+			if err := tarWriter.Close(); err != nil {
+				return nil, nil, err
+			}
+			// produce gzip
+			if err := gzipWriter.Close(); err != nil {
+				return nil, nil, err
+			}
+			tarWriter.Flush()
```
sutaakar: Shouldn't flush be called before close?

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+
+			var r io.Reader
+			br := bufio.NewReaderSize(&buf, 4096)
+			r = br
```
sutaakar: What is the purpose of this assignment, why not to use br directly?

cmd/kogito/command/deploy/deploy_service.go
sutaakar: can you please apply the change in this PR so it won't be forgotten?
```suggestion
			fileToWrite, err1 := ioutil.TempFile("", "compressed_kogito_resources_*.tgz")
```

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+			if _, err2 := io.Copy(fileToWrite, r); err2 != nil {
+				panic(err2)
+			}
+			defer fileToWrite.Close()
+
+			f, err := os.Open(fileToWrite.Name())
+			if err != nil {
+				return nil, nil, err
+			}
+			os.Remove(fileToWrite.Name())
```
sutaakar: Is it safe to remove opened file?

cmd/kogito/command/deploy/deploy_service_test.go
```diff
@@ -178,6 +178,43 @@ func Test_DeployCmd_WrongGitURL(t *testing.T) {
...
+	test.SetupCliTest(cli,
+		context.CommandFactory{BuildCommands: BuildCommands},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}},
+		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
+	lines, _, err := test.ExecuteCli()
+	assert.Error(t, err)
+	assert.Contains(t, lines, "Error: stat my-file.dmn: no such file or directory")
+}
+
+//func Test_DeployCmd_WithSingleKogitoResourceFileWithValidFileFromUrl(t *testing.T) {
```
sutaakar: Why are the tests commented?

pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -72,6 +72,17 @@ func (k *KogitoAppSpec) IsGitURIEmpty() bool {
 	return len(k.Build.GitSource.URI) == 0
 }
 
+// HasStandaloneKogitoResource checks if user has provided a single Kogito file resource to be deployed.
+func (k *KogitoAppSpec) HasStandaloneKogitoResource() bool {
+	if k == nil {
+		return false
+	}
+	if k.Build == nil || &k.Build.HasStandaloneKogitoResource == nil {
```
sutaakar: Just a nitpick, IMHO the HasStandaloneKogitoResource can never be nil (because it is not a pointer, but value).

pkg/client/openshift/buildconfig.go
```diff
@@ -101,6 +106,38 @@ func (b *buildConfig) TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (b
 	return true, nil
 }
 
+// TriggerFromFileBuild wil be called by kogito-cli when a build from file is performed.
+// When called a new build will be triggered with the request kogito resource or a tgz file.
+func (b *buildConfig) TriggerFromFileBuild(namespace string, r io.Reader, options *buildv1.BinaryBuildRequestOptions) (*buildv1.Build, error) {
```
sutaakar: Can you please extend the name of `b`, `bc` and `r`? These shortcuts complicates reading (at least for me).

pkg/client/openshift/buildconfig.go
```diff
@@ -101,6 +106,38 @@ func (b *buildConfig) TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (b
...
+
+	result := &buildv1.Build{}
+	buildName := fmt.Sprintf("%s%s", options.Name, "-builder")
+
+	// before upload the file, make sure that the build exist
+	err := b.waitForBuildConfig(5, 2*time.Second, func() (err error) {
+		if bc, err := b.client.BuildCli.BuildConfigs(namespace).Get(buildName, metav1.GetOptions{}); err != nil && errors.IsNotFound(err) {
+			log.Debugf("", bc.Namespace)
+			return err
+		}
```
sutaakar: What if err is produced and `errors.IsNotFound(err) ` is false? This case doesn't seem to be handled, code will behave like there is no error.

pkg/client/openshift/buildconfig.go
```diff
@@ -160,3 +197,20 @@ func newBuildRequest(triggedby string, bc *buildv1.BuildConfig) buildv1.BuildReq
...
+
+		if i >= (retries - 1) {
+			break
+		}
+		time.Sleep(retryInterval)
+
+		log.Debug("retrying after error:", err)
+	}
+	return fmt.Errorf("after %d attempts, last error: %s", retries, err)
```
sutaakar:
```suggestion
	return fmt.Errorf("after %d attempts, last error: %v", retries, err)
```

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+				fileReader, err := os.Open(resource)
+				if err != nil {
+					return nil, nil, err
+				}
+				options.AsFile = fileName
+				return fileReader, options, nil
+			}
+		}
+
+		if fileInfo.Mode().IsDir() {
```
radtriste: shouldn't be `else if` here ?
If `IsRegular()` is true and `isSuffixSupported(resource)`, we will still check if it is a dir ...

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -278,3 +319,149 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+
+			var buf bytes.Buffer
+			gzipWriter := gzip.NewWriter(&buf)
+			tarWriter := tar.NewWriter(gzipWriter)
+
+			var link string
+			filepath.Walk(resource, func(absoluteFilePath string, fileInfo os.FileInfo, err error) error {
+				if isSuffixSupported(fileInfo.Name()) {
+
+					if fileInfo.Mode()&os.ModeSymlink != 0 {
```
radtriste: does it not happen also for regular files ?

README.md
spolti: yes, IMHO there's no point to allow user to push any kind of file since the archetypes can not work properly.

spolti: does it make sense?

README.md
spolti: it is not clear, updated.

cmd/kogito/command/deploy/deploy_service.go
spolti: it really didn't make any difference, it is a leftover from  test I was doing.

cmd/kogito/command/deploy/deploy_service.go
spolti: didn't work, if I return the buffer reader or the io.Reader it comes a empty file.
this was the only way I found so far to make it work.

Do you have any suggestion?

pkg/apis/app/v1alpha1/kogitoapp_types.go
spolti: correct, but & points to the memory address which can be null, no?

pkg/client/openshift/buildconfig.go
spolti: b is comming from the same class, buildCOnfig struct, same for all other methods.
will update the others.

pkg/client/openshift/buildconfig.go
spolti: few lines above the error is being handled, should I move it to somewhere else?

cmd/kogito/command/deploy/deploy_service.go
spolti: +1

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+
+While checking OpenShift build logs with this command **oc logs -f bc/test-builder -n kogito** you can see on the beginning ot the build log:
+```
+Receiving source from STDIN as file TrafficViolation.dmn
+Using docker-registry.default.svc:5000/spolti-1/kogito-quarkus-ubi8-s2i@sha256:729e158710dedba50a49943ba188d8f31d09568634896de9b903838fc4e34e94 as the s2i builder image
+```
+
+```bash
+# from a git repository
+$ kogito -service example-dmn-quarkus --from-file https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
```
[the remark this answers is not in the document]
spolti: LOL, again hahaha
dam ctrl c and v

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. In that case all we need to do
+is, instead specify a directory instead, if any valid file is found on the given directory, the cli will compress these
+files and upload to the OpenShift cluster. To start upload various files, put all the desired files in a directory and 
+specify it as parameter. Note that if you have more files on this directory, they wont be copied.
+
+```bash
+kogito --verbose deploy-service test --from-file /home/user/development/kogito-test/
```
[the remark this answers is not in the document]
spolti: no need, probably leftover.

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```
+
+If for some reason the build fails, pass the env BUILD_LOGLEVEL with the desired level, i.e. 9 as a build-env parameter, i.e.:
```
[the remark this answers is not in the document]
spolti: +1

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```bash
+kogito --verbose deploy-service test --from-file /home/user/development/kogito-test/  --build-env BUILD_LOGLEVEL=9
```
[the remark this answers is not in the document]
spolti: here it is intentional, the cli verbose can bring useful info.

README.md
```diff
@@ -195,6 +196,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+$ kogito -service example-dmn-quarkus --from-file https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. In that case all we need to do
+is, instead specify a directory instead, if any valid file is found on the given directory, the cli will compress these
+files and upload to the OpenShift cluster. To start upload various files, put all the desired files in a directory and
```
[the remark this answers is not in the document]
spolti: is 'To upload more than one file' better?

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -90,7 +103,8 @@ func (i *deployCommand) RegisterHook() {
 		Short:   "Deploys a new Kogito Runtime Service into the given Project",
 		Aliases: []string{"deploy"},
 		Long: `deploy-service will create a new Kogito Runtime Service in the Project context. 
-        If the source is provided, the build will take place on the cluster, otherwise you must upload the application binaries via "oc start-build [NAME-binary] --from-dir=target". 
+        If the source is provided, the build will take place on the cluster, otherwise you must upload the application binaries via "oc start-build [NAME-
```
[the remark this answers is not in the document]
spolti: +1

cmd/kogito/command/deploy/deploy_service.go
spolti: actually I am removing the file and returning the content read from it before deletion.

cmd/kogito/command/deploy/deploy_service_test.go
spolti: because it is not finished yet.

pkg/client/openshift/buildconfig.go
spolti: below*

cmd/kogito/command/deploy/deploy_service.go
spolti: I think that it doesn't matter since the process will be endded anyways when cli finish the steps, right?

cmd/kogito/command/deploy/deploy_service.go
spolti: not sure if I understand your point.

cmd/kogito/command/deploy/deploy_service.go
spolti: make sense.

cmd/kogito/command/deploy/deploy_service_test.go
spolti: done, tests finished.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -82,12 +98,15 @@ func (i *deployCommand) RegisterHook() {
 		Short:   "Deploys a new Kogito Runtime Service into the given Project",
 		Aliases: []string{"deploy"},
 		Long: `deploy-service will create a new Kogito Runtime Service in the Project context. 
-        If the source is provided, the build will take place on the cluster, otherwise you must upload the application binaries via "oc start-build [NAME-binary] --from-dir=target". 
-		Project context is the namespace (Kubernetes) or project (OpenShift) where the Service will be deployed.
-		To know what's your context, use "kogito project". To set a new Project in the context use "kogito use-project NAME".
+	If the [SOURCE] is provided, the build will take place on the cluster.
```
spolti, line 101: this will generate it:

```
deploy-service will create a new Kogito Runtime Service in the Project context. 
	If the [SOURCE] is provided, the build will take place on the cluster.
	If not, you can also provide a dmn/drl/bpmn/bpmn2 file or a directory containing one or more of those files, using the --from-file
	Or you can also later upload directly the application binaries via "oc start-build [NAME-binary] --from-dir=target
			
	Project context is the namespace (Kubernetes) or project (OpenShift) where the Service will be deployed.
	To know what's your context, use "kogito project". To set a new Project in the context use "kogito use-project NAME".

	Please note that this command requires the Kogito Operator installed in the cluster.
	For more information about the Kogito Operator installation please refer to https://github.com/kiegroup/kogito-cloud-operator#kogito-operator-installation.

```

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
 
+### Build From File
+
+Kogito Operator can configure a build for you that will pull and build your application through the source to image (s2i)
+feature (available only on OpenShift). This kind of build consists to upload a Kogito service file, e.g. a dmn or a bpmn file
+to the OpenShift Cluster and trigger a new Source to Image build automatically.
+
+You can provide a single dmn, drl, bpmn or bpmn2 files, the standalone file can be locally on the filesystem or on a Web Site, like Git Hub:
+
+
```
Sgitario, line 209: There is an extra empty line here.

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```bash
+# from filesystem
+$ kogito deploy-service example-dmn-quarkus --from-file TrafficViolation.dmn
```
Sgitario, line 212: I think it would be good to use the same source for filesystem and from git repository. For example, the TrafficViolation.dmn file does not exist, but "Traffic Violation.dmn".

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```bash
+# from a git repository
+$ kogito deploy-service example-dmn-quarkus --from-file https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. IIn that case,
```
Sgitario, line 233: Typo here: "IIn"

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+$ kogito deploy-service example-dmn-quarkus --from-file https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. IIn that case, 
+all we need to do is to specify a whole directory. If any valid file is found in the given directory, the cli will compress these
+files and upload to the OpenShift cluster. To upload more than one file, put all the desired files in a directory and
```
Sgitario, line 235: This line must be appended to the previous line.

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+The provided source is a dir, packing files.
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If for some reason the build fails, pass the env BUILD_LOGLEVEL with the desired level, e.g. 9 as a build-env parameter:
+```bash
+kogito --verbose deploy-service test --from-file /home/user/development/kogito-test/  --build-env BUILD_LOGLEVEL=9
+```
+
```
Sgitario, line 254: Extra blank line here.

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+As output of the cli command above, we will have something similar to:
+
+```bash
+The provided source is a dir, packing files.
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
+```
+
+If for some reason the build fails, pass the env BUILD_LOGLEVEL with the desired level, e.g. 9 as a build-env parameter:
```
Sgitario, line 250: Can we make the "BUILD_LOGLEVEL" a link? Maybe: https://docs.openshift.com/container-platform/3.5/dev_guide/builds/basic_build_operations.html. As far I could see, the level 9 does not exist, the range is between 1 to 5.

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. IIn that case, 
+all we need to do is to specify a whole directory. If any valid file is found in the given directory, the cli will compress these
+files and upload to the OpenShift cluster. To upload more than one file, put all the desired files in a directory and 
+specify it as parameter. Note that if you have other unsupported files in this directory, they will not be copied.
+
+```bash
+kogito deploy-service test --from-file /home/user/development/kogito-test/
```
Sgitario, line 239: I have a question about uploading folders. Do we support also Java sources like models? As far I can see below, we do not.

pkg/apis/app/v1alpha1/kogitoapp_types.go
sutaakar: & creates a pointer of a structure, true, but it points to the structure which it was created from :) so the structure exists.
I cannot image any case when `&k.Build.HasStandaloneKogitoResource` can be nil (if k.Build is not nil).

README.md
spolti: no, only bpmn, drl or dmn.

cmd/kogito/command/deploy/deploy_service.go
sutaakar: true

cmd/kogito/command/deploy/deploy_service.go
sutaakar: My local experiments show that when I use br or &buf on https://github.com/kiegroup/kogito-cloud-operator/pull/247/files#diff-7b68eb431df193ad354067385d51880dR426 then content of the file is returned.
Checked using test:
```
func Test_Qqq(t *testing.T) {
	fileToWrite, _ := ioutil.TempFile(os.TempDir(), "compressed_kogito_resourcesxxx_*.tgz")
	reader, _, _ := matchKogitoResources("/home/ksuta/Downloads/qqq")
	io.Copy(fileToWrite, reader)
	fileToWrite.Sync()
	fileToWrite.Close()
}
```
(need to be careful as the test results can be cached, skipping execution in such case).

sutaakar: When I think about it, what prevents us from returning directly &buf, skipping whole temp file handling?

pkg/client/openshift/buildconfig.go
sutaakar: I see now, thanks. I overlooked `(err error)` declaration, binding err to the output.

README.md
spolti: Maybe for windows users?
At least on Unix systems it is not common to use name files with spaces.

We can change it.

@radtriste @sutaakar what do you guys think?

README.md
spolti: surething :)

cmd/kogito/command/deploy/deploy_service.go
spolti: the content might be returned, but not on a correct tar.gz format, after build the buf only, without writing the file on filesyste, looks like it goes a empty file or not recognized one. I'll try again to double check it.

pkg/apis/app/v1alpha1/kogitoapp_types.go
spolti: probably :)

spolti: will update.

ricardozanini:
I won't join the trend right now. Once you guys finish all this discussions/threads, I'll start reviewing it.

README.md
ricardozanini: > I think it would be good to use the same source for filesystem and from git repository

I didn't understand this question. Doesn't matter if the file has space or not, the plugin will create the project anyway. What we can do is run a regexp in the file name and emit a warn or deny the upload based on this rule.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -151,15 +166,31 @@ func (i *deployCommand) InitHook() {
 	i.command.Flags().StringVar(&i.flags.imageVersion, "image-version", "", "Image version for standard Kogito build images. Ignored if a custom image is set for image-s2i or image-runtime.")
 	i.command.Flags().StringVar(&i.flags.mavenMirrorURL, "maven-mirror-url", "", "Internal Maven Mirror to be used during source-to-image builds to considerably increase build speed, e.g: https://my.internal.nexus/content/group/public")
 	i.command.Flags().BoolVar(&i.flags.enableIstio, "enable-istio", false, "Enable Istio integration by annotating the Kogito service pods with the right value for Istio controller to inject sidecars on it. Defaults to false")
+	i.command.Flags().BoolVarP(&i.flags.fromFile, "from-file", "f", false, "If set, a build from a single file, no pom.xml needed, will start, you can provide a entire directory, a single file from filesystem or url, the supported files are dmn, drl, bpmn or bpmn2.")
```
ricardozanini, line 169: Since this is a `bool`, I can write the command like:

```
kogito deploy test myprocess.bpmn --env ENV1=VALUE  --from-file --maven-mirror-url="https:/"
```
Do you think that it makes sense from a UX perspective? 

Why not a `string`?

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -249,6 +281,20 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+		options.Name = kogitoApp.Name
+		cli, buildCliErr := client.NewClientBuilder().WithBuildClient().Build()
+		if buildCliErr != nil {
+			return buildCliErr
+		}
+
+		_, err := openshift.BuildConfigC(cli).TriggerBuildFromFile(i.flags.Project, userProvidedFile, options)
+		if err != nil {
+			return err
+		}
```
ricardozanini, line 295:
```suggestion
		cli, err := client.NewClientBuilder().WithBuildClient().Build()
		if err != nil {
			return err
		}

		_, err = openshift.BuildConfigC(cli).TriggerBuildFromFile(i.flags.Project, userProvidedFile, options)
		if err != nil {
			return err
		}
```

I think it's clearer this way

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -249,6 +281,20 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
 	if hasSource {
```
ricardozanini, line 281: `fromFile` and `hasSource` it's almost the same thing semantically. Can we think on another naming/approach?

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -265,3 +311,141 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+				fileReader, err := os.Open(resource)
+				if err != nil {
+					return nil, nil, err
+				}
+				options.AsFile = fileName
+				return fileReader, options, nil
+			}
+
+		} else if fileInfo.Mode().IsDir() {
+			log.Info("The provided source is a dir, packing files.")
```
ricardozanini, line 365: Please add those messages to the `message` package.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -265,3 +311,141 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+
+					fileToCompress, err := os.Open(absoluteFilePath)
+					if err != nil {
+						return err
+					}
+					defer fileToCompress.Close()
+
+					if _, err := io.Copy(tarWriter, fileToCompress); err != nil {
+						return err
+					}
```
ricardozanini, line 405: Please move this compression feature to an `util` package.

cmd/kogito/command/deploy/deploy_service.go
ricardozanini: Like I mentioned in other comment, please move all this file handling/compression to an util/compression package and deal with it there, so we can create a bunch of unit tests to verify all those behaviors.

deploy/olm-catalog/kogito-operator/0.9.0/app.kiegroup.org_kogitoapps_crd.yaml
```diff
@@ -178,6 +178,11 @@ spec:
                   required:
                   - uri
                   type: object
+                hasStandaloneKogitoResource:
```
ricardozanini, line 181: This should not be in the 0.9.x  dir.

pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -130,6 +141,8 @@ type KogitoAppBuildObject struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 	// Internal Maven Mirror to be used during source-to-image builds to considerably increase build speed
 	MavenMirrorURL string `json:"mavenMirrorURL,omitempty"`
+	// Internal usage only, will be set by CLI when user create a kogito service from a Kogito resource, e.g. dmn/drl/bpmn standalone files
+	HasStandaloneKogitoResource bool `json:"hasStandaloneKogitoResource,omitempty"`
```
ricardozanini, line 145: I don't think we should expose an "internal" use only attribute in a public interface. We need to figure another way to have this feature in the CR.

pkg/client/openshift/buildconfig.go
```diff
@@ -101,6 +106,37 @@ func (b *buildConfig) TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (b
...
 
+// TriggerBuildFromFile wil be called by kogito-cli when a build from file is performed.
+// When called a new build will be triggered with the request kogito resource or a tgz file.
+func (b *buildConfig) TriggerBuildFromFile(namespace string, bodyPost io.Reader, options *buildv1.BinaryBuildRequestOptions) (*buildv1.Build, error) {
+
+	result := &buildv1.Build{}
+	buildName := fmt.Sprintf("%s%s", options.Name, "-builder")
+
+	// before upload the file, make sure that the build exist
+	err := b.waitForBuildConfig(5, 2*time.Second, func() (err error) {
```
ricardozanini, line 117: Can we get rid of these magic numbers with `const`s? :dancers:

pkg/controller/kogitoapp/resource/build_config_s2i.go
```diff
@@ -101,6 +113,16 @@ func setBCS2IStrategy(kogitoApp *v1alpha1.KogitoApp, buildConfig *buildv1.BuildC
 		envs = framework.EnvOverride(envs, corev1.EnvVar{Name: mavenMirrorURLEnvVar, Value: kogitoApp.Spec.Build.MavenMirrorURL})
 	}
 
+	// if user has provided a file, binary build should be used instead.
+	if kogitoApp.Spec.HasStandaloneKogitoResource() {
```
ricardozanini, line 117: Why not a new buildconfiguration instead of modifying this one if this if clause?

README.md
radtriste, line 212:
```suggestion
$ kogito deploy-service example-dmn-quarkus --from-file "Traffic Violation.dmn"
```

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+feature (available only on OpenShift). This kind of build consists to upload a Kogito service file, e.g. a dmn or a bpmn file
+to the OpenShift Cluster and trigger a new Source to Image build automatically.
+
+You can provide a single dmn, drl, bpmn or bpmn2 files, the standalone file can be locally on the filesystem or on a Web Site, like Git Hub:
+
+
+```bash
+# from filesystem
+$ kogito deploy-service example-dmn-quarkus --from-file TrafficViolation.dmn
+File found /tmp/kogito-test/src/main/resources/TrafficViolation.dmn.
```
radtriste, line 213:
```suggestion
File found /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn.
```

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```
+Receiving source from STDIN as file TrafficViolation.dmn
+Using docker-registry.default.svc:5000/spolti-1/kogito-quarkus-ubi8-s2i@sha256:729e158710dedba50a49943ba188d8f31d09568634896de9b903838fc4e34e94 as the s2i builder image
+```
+
+```bash
+# from a git repository
+$ kogito deploy-service example-dmn-quarkus --from-file https://raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic%20Violation.dmn
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
```
radtriste, line 228:
```suggestion
The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/example-dmn-quarkus-builder -n kogito'
```

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```
+
+If you check the logs, we can also see the dmn file being used to to build the Kogito service.
```
radtriste, line 231:
```suggestion
If you check the logs, we can also see the dmn file being used to build the Kogito service.
```

README.md
radtriste, line 233:
```suggestion
Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. In that case, 
```

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+
+Sometimes we also need to provide more than one Kogito resource file, e.g. a drl and a bpmn. IIn that case, 
+all we need to do is to specify a whole directory. If any valid file is found in the given directory, the cli will compress these
+files and upload to the OpenShift cluster. To upload more than one file, put all the desired files in a directory and 
+specify it as parameter. Note that if you have other unsupported files in this directory, they will not be copied.
```
radtriste, line 236:
```suggestion
specify it as parameter.  
Note that if you have other unsupported files in this directory, they will not be copied.
```

README.md
```diff
@@ -197,6 +198,61 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+```bash
+kogito deploy-service test --from-file /home/user/development/kogito-test/
+```
+
+As output of the cli command above, we will have something similar to:
+
+```bash
+The provided source is a dir, packing files.
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
```
radtriste, line 247:
```suggestion
The requested file(s) was successfully upload to OpenShift, a build with the file(s) should now be running. To see its logs, run 'oc logs -f bc/test-builder -n kogito'
```

pkg/controller/kogitoapp/resource/build_config_s2i.go
spolti: to avoid code replication, it is basically the same thing than a usual from source build.

ricardozanini: We could review it and reuse the same function and decorate it to address this new build.

Well, I'll change it anyway with the Kube support, so do as you think is more convenient for now.

README.md
ricardozanini: Oh you guys are talking about the README description of the example. Right, agreed.

spolti: using spaces from now.

cmd/kogito/command/deploy/deploy_service.go
spolti: @sutaakar apparently it worked now, changed some things.

README.md
spolti: it can be improved in the future if needed.
README updated.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -151,13 +164,24 @@ func (i *deployCommand) InitHook() {
 	i.command.Flags().StringVar(&i.flags.imageVersion, "image-version", "", "Image version for standard Kogito build images. Ignored if a custom image is set for image-s2i or image-runtime.")
 	i.command.Flags().StringVar(&i.flags.mavenMirrorURL, "maven-mirror-url", "", "Internal Maven Mirror to be used during source-to-image builds to considerably increase build speed, e.g: https://my.internal.nexus/content/group/public")
 	i.command.Flags().BoolVar(&i.flags.enableIstio, "enable-istio", false, "Enable Istio integration by annotating the Kogito service pods with the right value for Istio controller to inject sidecars on it. Defaults to false")
+	i.command.Flags().StringVarP(&i.flags.fromFile, "from-file", "f", "", "If set, a build from a single file, no pom.xml needed, will start, you can provide a entire directory, a single file from filesystem or url, the supported files are dmn, drl, bpmn or bpmn2. Porperties files is also allowed.")
```
Sgitario, line 167: I would change the description to "If set, a build from local sources will start (no pom.xml needed). You can provide a entire directory, a single file from filesystem or url, the supported files are dmn, drl, bpmn or bpmn2. Properties files is also allowed."

Also, note the typo in the properties word.

cmd/kogito/version/version.go
```diff
@@ -16,5 +16,5 @@ package version
 
 var (
 	// Version of Kogito CLI
-	Version = "0.11.0"
+	Version = "0.10.0"
```
Sgitario, line 19: Is this change ok?

version/version.go
```diff
@@ -16,5 +16,5 @@ package version
 
 var (
 	// Version of Kogito Operator
-	Version = "0.11.0"
+	Version = "0.10.0"
```
Sgitario, line 19: Is this ok?

cmd/kogito/version/version.go
radtriste: it shouldn't

cmd/kogito/command/deploy/deploy_service.go
radtriste, line 167: Why are we using an option and not the `[SOURCE]` argument itself for transmitting those source files ?
We could then leverage the `matchKogitoResources` on that source.
If match, then it is from source file.
If not then we consider a Git repository.
what do you think ?

pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -277,6 +277,8 @@ type Builds struct {
 	// Builds have been stopped from executing
 	// +listType=set
 	Cancelled []string `json:"cancelled,omitempty"`
+	// Flag to let operator know if the given build if from file, set by CLI when user create a kogito service from a Kogito resource, e.g. dmn/drl/bpmn standalone files
```
radtriste, line 280:
```suggestion
	// Flag to let operator know if the given build is from file, set by CLI when user create a kogito service from a Kogito resource, e.g. dmn/drl/bpmn standalone files
```

pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -277,6 +277,8 @@ type Builds struct {
...
+	BuildFromAsset bool `json:"buildFromAsset,omitempty"`
```
radtriste, line 281: shouldn't we have an array here ?
New builds could be later defined with Git source or we would have possibility to override a current CR with another source file ?

radtriste:
btw, couldn't we reuse that buildconfig creation from CLI for also binary builds ?
instead of having 2 commands to launch the binary build ...

cmd/kogito/version/version.go
spolti: left over :)

spolti:
> w, couldn't we reuse that buildconfig creation from CLI for also binary builds ?
> instead of having 2 commands to launch the binary build ...

We could, but I think that it is better to track in another issue.

pkg/apis/app/v1alpha1/kogitoapp_types.go
spolti: I don't think an array is necessary, all we need is just a bool flag to sinalize to operator that this build is from standalone kogito assets.

cmd/kogito/command/deploy/deploy_service.go
spolti: I could work, but IMHO, I think is better keep a separate flag/option for it. since we have plans to integrate with the binary builds later.

I would like to hear opnion from others about it.
@sutaakar @Sgitario @MarianMacik

spolti: cool, thanks.

version/version.go
spolti: letft over

README.md
```diff
@@ -197,6 +198,63 @@ $ oc start-build example-quarkus-binary --from-dir=target
...
+to the OpenShift Cluster and trigger a new Source to Image build automatically.
+
+You can provide a single dmn, drl, bpmn, bpmn2 and properties files, the standalone file can be locally on the filesystem or on a Web Site, like GitHub:
+
+```bash
+# from filesystem
+$ kogito deploy-service example-dmn-quarkus --from-file  /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/"Traffic Violation.dmn"
+File found /tmp/kogito-examples/dmn-quarkus-example/src/main/resources/Traffic Violation.dmn.
+...
+The requested file(s) was successfully upload to OpenShift, a build with this file(s) should now be running. To see the logs, run 'oc logs -f bc/example-dmn-quarkus-builder -n kogito'
```
evacchi, line 214: "upload**ed** to openshift". Should fix in docs, and source code.

cmd/kogito/command/message/kogitoapp.go
```diff
@@ -20,9 +20,12 @@ const (
...
 	KogitoAppUploadBinariesInstruction = "Your Kogito Runtime Service needs the application binaries to proceed. To upload your binaries please run 'oc start-build %s-binary --from-dir=target -n %s' from your project's root"
-
-	KogitoAppNoMgmtConsole = `To more easily manage your Kogito Runtime Service install Data Index Service and Process Instance Management. 
+	KogitoAppNoMgmtConsole             = `To more easily manage your Kogito Runtime Service install Data Index Service and Process Instance Management. 
 For how to install see: https://github.com/kiegroup/kogito-cloud-operator#kogito-data-index-service-deployment
 and https://github.com/kiegroup/kogito-cloud-operator#kogito-management-console-install`
-	KogitoAppMgmtConsoleEndpoint = `You can manage your process using the management console: %s`
+	KogitoAppMgmtConsoleEndpoint      = `You can manage your process using the management console: %s`
+	KogitoAppSuccessfullyUploadedFile = "The requested file(s) was successfully upload to OpenShift, a build with this file(s) should now be running. To see the logs, run 'oc logs -f bc/%s-builder -n %s'"
```
evacchi, line 27: upload**ed**

pkg/apis/app/v1alpha1/kogitoapp_types.go
spolti: I don't know if it is possible to update a source file, probably the best option wouldbe remove the service and create it again.

README.md
spolti: done.

pkg/apis/app/v1alpha1/kogitoapp_types.go
radtriste: @sutaakar wdyt ? should we have this field here or at the root of `Status` part ?

radtriste:
> > w, couldn't we reuse that buildconfig creation from CLI for also binary builds ?
> > instead of having 2 commands to launch the binary build ...
> 
> We could, but I think that it is better to track in another issue.

Agree => https://issues.redhat.com/browse/KOGITO-2113

cmd/kogito/command/deploy/deploy_service.go
Sgitario: Sorry in advance if I misunderstood something: as far I can see the difference is that the [SOURCE] argument points out to a folder that contains always a Kogito project (with pom.xml) and the new "--from-file" flag is to specify the supported source files and also a folder (with these files). 

Taking into account that the supported source files will likely include also Java files in the future, I don't see much difference between [SOURCE] or "--from-file" options. And having them separated might make more complicated the maintenance:

- what to do if we specify the two options at the same time? 
- both options have the same behaviour using the same sources?) 

Therefore, unless this is very complex, I think it makes sense to do it smart enough to adapt the [SOURCE] to behave accordingly to what is found in the folder (be a kogito project, a binary, spare files, etc).

radtriste: The source for now is a Github repository as we needed it for s2i builds.
For now, the source has precedence over the --from-file option.
That is why I was proposing to extend its purpose and avoid another option.

spolti:
> > > w, couldn't we reuse that buildconfig creation from CLI for also binary builds ?
> > > instead of having 2 commands to launch the binary build ...
> > 
> > 
> > We could, but I think that it is better to track in another issue.
> 
> Agree => https://issues.redhat.com/browse/KOGITO-2113

Thanks.

cmd/kogito/command/deploy/deploy_service.go
spolti: alright, I made some tweaks and looks like it is working, no flag needed.

cmd/kogito/command/error/kogitoapp_errors.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package error
+
+import (
+	"fmt"
+)
+
+// KogitoAssetNotFound Custom error for handling build from file errors.
+type KogitoAssetNotFound struct {
```
sutaakar, line 22: The struct name is misleading. From the name it looks like it marks just error related to missing asset, but in reality it is used for any asset build related error (for example when asset is not among supported types).
Can you please rename it, for example to something like KogitoAssetFileBuildError ?

cmd/kogito/command/deploy/deploy_service_test.go
```diff
@@ -178,6 +182,68 @@ func Test_DeployCmd_WrongGitURL(t *testing.T) {
...
+		&apiextensionsv1beta1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: v1alpha1.KogitoAppCRDName}})
+	lines, _, err := test.ExecuteCli()
+	assert.NotContains(t, lines, "successfully deployed")
+	assert.NotContains(t, lines, "successfully created")
+	// expect error at this point as there is no valid cluster to do a post request. (connect: no route to host)
+	assert.Error(t, err)
+	assert.NotContains(t, lines, "Error:  https:ss//raw.githubusercontent.com/kiegroup/kogito-examples/master/dmn-quarkus-example/src/main/resources/Traffic Violation.dmnsd")
+}
+
+func Test_DeployCmd_WithSingleKogitoResourceFileWithValidFile(t *testing.T) {
```
sutaakar, line 229:
```suggestion
func Test_DeployCmd_WithSingleKogitoResourceFileWithInvalidFileType(t *testing.T) {
```

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -157,9 +166,34 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+				return t
+
+			case *kogitoerror.KogitoAssetNotFound:
+				return t
+
+			case *url.Error:
+				return t
+
+			default:
+				fmt.Println(reflect.TypeOf(t))
```
sutaakar, line 186: Is this a debug leftover?

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -157,9 +166,34 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+				log.Warnf(message.KogitoAppAssetNotSupported)
```
sutaakar, line 187: It would be good to add the asset path as part of warning message.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -265,3 +317,69 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+
+			ioTgzR, err := buildutil.ProduceTGZfile(resource)
+			if err != nil {
+				return nil, nil, err
+			}
+			return ioTgzR, options, nil
+		}
+		return nil, nil, &kogitoerror.KogitoAssetNotFound{Asset: resource}
+	}
+	return nil, nil, nil
```
sutaakar, line 384:
```suggestion
		if buildutil.IsSuffixSupported(fileName) {
			response, err := http.Get(resource)
			if err != nil {
				return nil, nil, fmt.Errorf("failed to download %s, error message: %s", resource, err.Error())
			}
			options.AsFile = fileName
			log.Infof(message.KogitoAppFoundAsset, fileName)
			return response.Body, options, nil
		}
		return nil, nil, nil

	// handle single file, files on dir.
	default:
		fileInfo, err := os.Stat(resource)
		if err != nil {
			return nil, nil, err
		}

		if fileInfo.Mode().IsRegular() {
			if buildutil.IsSuffixSupported(resource) {
				log.Infof(message.KogitoAppFoundFile, resource)
				ff := strings.Split(resource, "/")
				fileName = strings.Join(strings.Fields(ff[len(ff)-1]), "")
				fileReader, err := os.Open(resource)
				if err != nil {
					return nil, nil, err
				}
				options.AsFile = fileName
				return fileReader, options, nil
			}

		} else if fileInfo.Mode().IsDir() {
			log.Info(message.KogitoAppProvidedFileIsDir)

			ioTgzR, err := buildutil.ProduceTGZfile(resource)
			if err != nil {
				return nil, nil, err
			}
			return ioTgzR, options, nil
		}
		return nil, nil, &kogitoerror.KogitoAssetNotFound{Asset: resource}
	}
```

sutaakar: The proposed change is to move `return nil, nil, nil` into line 352 to make switch case flow more readable (all other branches have own return statements so the only way how to reach the final return statement is from that 352 line).

cmd/kogito/command/util/build_from_file.go
```diff
@@ -0,0 +1,100 @@
...
+	"compress/gzip"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/message"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var supportedExtensions = []string{"dmn", "drl", "bpmn", "bpmn2", "properties"}
```
sutaakar, line 29: It may be good to check suffixes starting with dot, so the `IsSuffixSupported` function won't match any name ending with these strings (for example something like `https://github.com/<username>/my-first-bpmn`)

cmd/kogito/command/util/build_from_file_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestProduceTGZfile(t *testing.T) {
+	tempDir := "/tmp/kogito-test/"
+	simpleContent := []byte("hello World!!")
+	files := []string{"file.bpmn", "file2.drl", "file3.bpmn2", "file4.dmn", "file5.properties", "file6.unsupported"}
+	_, err := os.Stat("test")
```
sutaakar, line 32:
```suggestion
	_, err := os.Stat(tempDir)
```

pkg/client/openshift/buildconfig.go
```diff
@@ -101,6 +111,37 @@ func (b *buildConfig) TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (b
...
+	// before upload the file, make sure that the build exist
+	err := b.waitForBuildConfig(checkBcRetries, checkBcRetriesInterval, func() (err error) {
+		if targetBuildConfig, err := b.client.BuildCli.BuildConfigs(namespace).Get(buildName, metav1.GetOptions{}); err != nil && errors.IsNotFound(err) {
+			log.Debugf("BuildConfig %s not found in the %s namespace", targetBuildConfig.Name, targetBuildConfig.Namespace)
+			return err
+		}
+		return
+	})
+	if err != nil {
+		log.Error(err)
```
sutaakar, line 130: What is the reason to log an error here and continue?
Should build be triggered even when `waitForBuildConfig` ends in error?

pkg/controller/kogitoapp/resource/requested_resources.go
```diff
@@ -62,20 +62,21 @@ func GetRequestedResources(kogitoApp *v1alpha1.KogitoApp, client *client.Client)
...
 		}
-
 		chain.Resources.BuildConfigS2I = &bc
 	}
 	return chain
 }
 
 func buildConfigRuntimeBuilder(chain *builderChain) *builderChain {
-	if !chain.KogitoApp.Spec.IsGitURIEmpty() {
+	// if there is gitURI or standalone kogito resource file, create a s2i build config.
```
sutaakar, line 78:
```suggestion
	// if there is gitURI or standalone kogito resource file, create a runtime build config.
```

pkg/apis/app/v1alpha1/kogitoapp_types.go
sutaakar: I think it is ok to keep it here as it is related to builds.

sutaakar:
@spolti The JIRA mentions capability to reupload source file to redeploy the service:
```
1. You do not have a service deployed and you want to give it a go to a BPMN process you just built. You do:

    kogito deploy-service my-service /path/to/bpmn/domain/classes/files

2. Your service is up and running.

3. You made changes to your files and want to see them again in the cluster. You repeat the command:

    kogito deploy-service my-service /path/to/bpmn/domain/classes/files
```

However when I try it then I get `Error: Looks like a Kogito App with the name 'example-dmn-quarkus' already exists in this context/namespace. Please try another name`

cmd/kogito/command/error/kogitoapp_errors.go
spolti: make sense, thanks,.

cmd/kogito/command/deploy/deploy_service.go
spolti: yes, it is :)

cmd/kogito/command/deploy/deploy_service.go
spolti: good point.

cmd/kogito/command/util/build_from_file.go
spolti: adjusted.

cmd/kogito/command/util/build_from_file_test.go
spolti: good catch.

pkg/client/openshift/buildconfig.go
spolti: +1

spolti:
> @spolti The JIRA mentions capability to reupload source file to redeploy the service:
> 
> ```
> 1. You do not have a service deployed and you want to give it a go to a BPMN process you just built. You do:
> 
>     kogito deploy-service my-service /path/to/bpmn/domain/classes/files
> 
> 2. Your service is up and running.
> 
> 3. You made changes to your files and want to see them again in the cluster. You repeat the command:
> 
>     kogito deploy-service my-service /path/to/bpmn/domain/classes/files
> ```
> 
> However when I try it then I get `Error: Looks like a Kogito App with the name 'example-dmn-quarkus' already exists in this context/namespace. Please try another name`

We could discuss it, but shouldn't this behaviour be enabled by a different flag on only be available if some kind of development flag is enabled?
We have plans to enable the remote Quarkus dev mode, which will fix this behavior, at least for Quarkus based apps.

Otherwise, if we want to be able to update the build config, all files will have to be uploaded again, will be better at least add a new flag to make it explicit to the user.
We could also add a extra check here https://github.com/kiegroup/kogito-cloud-operator/blob/master/cmd/kogito/command/deploy/deploy_service.go#L181
just to trigger a new build with the current file.


What do you guys thing?

cmd/kogito/command/deploy/deploy_service.go
spolti: Make sense, will apply and test it tomorrow morning.

radtriste:
I think that, like we would update a CR and apply it, we should give the possibility to redeploy a service via CLI without having to delete it. 
It will just replace/update the current existing. And we would display a message that is has been updated.

And that should be for any `deploy-service` cases (source file, binary file, git source).

I can see the "safe" side that you want to take by adding an option but I think it should not be the role of the CLI.

spolti:
@radtriste I would agree with you, but maybe, if we will change the behaviour for other ooptions, probably track/discuss it in a new issue would be better.

pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -277,6 +277,8 @@ type Builds struct {
 	// Builds have been stopped from executing
 	// +listType=set
 	Cancelled []string `json:"cancelled,omitempty"`
+	// Flag to let operator know if the given build is from file, set by CLI when user create a kogito service from a Kogito resource, e.g. dmn/drl/bpmn standalone files
+	BuildFromAsset bool `json:"buildFromAsset,omitempty"`
```
xiezhang7, line 281: I think this flag should be part of the CR spec [KogitoAppBuildObject](https://github.com/kiegroup/kogito-cloud-operator/blob/0f38a87706e3f16665a56c28d71205a3516926d7/pkg/apis/app/v1alpha1/kogitoapp_types.go#L95), not in the CR [status](https://github.com/kiegroup/kogito-cloud-operator/blob/0f38a87706e3f16665a56c28d71205a3516926d7/pkg/apis/app/v1alpha1/kogitoapp_types.go#L226).

spolti: It was initially there, @ricardozanini suggested remove it from there.

pkg/apis/app/v1alpha1/kogitoapp_types.go
spolti: @xiezhang7 see here.

pkg/apis/app/v1alpha1/kogitoapp_types.go
xiezhang7: I think we can find a way to stop it from appearing in the CRD, I'll take a look and see if we can do it by using some annotations. That's way we can completely avoid it from being exposed externally. Putting it in the `status` doesn't make much sense, and users still have access to the flag by setting a value in the `status`.

xiezhang7: @ricardozanini @spolti It seems it is not possible NOT to expose this flag, because the CLI is an external tool. If the flag is not exposed, CLI won't work.  On the other hand, if CLI can use the flag, users can also use this flag with `kubectl` or `oc` client. Actually, I think users are able to submit the files and trigger a Build manually without using the CLI anyway, although it is not documented. In this case, it makes no difference to put the flag in `status`.

spolti: @xiezhang7 not sure if I understand your comment, what are you suggesting?

xiezhang7: I think we should put the flag in the `spec` instead of `status`. Because the flag is not supposed to be hidden, it has to be exposed externally so that the CLI can use it. But exposing it in `status` doesn't make sense according to this [definition by kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#object-spec-and-status). So I think we should put it in the `spec` and make proper documentation that it is for the CLI.

spolti: flag removed.

pkg/controller/kogitoapp/resource/requested_resources_test.go
```diff
@@ -154,7 +160,34 @@ func Test_buildConfigS2IBuilder(t *testing.T) {
...
 	assert.NotNil(t, chain)
-	assert.Nil(t, chain.Resources.BuildConfigS2I)
+	assert.NotNil(t, chain.Resources.BuildConfigS2I)
+
+	chain = buildConfigS2IBuilder(gitURLGiven)
+	assert.NotNil(t, chain)
+	assert.NotNil(t, chain.Resources.BuildConfigS2I)
+}
+
+func Test_buildConfigS2IBuilderWithoutGitUrl(t *testing.T) {
```
sutaakar, line 170: It seems that this test is same as Test_buildConfigS2IBuilderWithGitUrl

pkg/controller/kogitoapp/resource/requested_resources.go
```diff
@@ -62,30 +62,43 @@ func GetRequestedResources(kogitoApp *v1alpha1.KogitoApp, client *client.Client)
...
+		chain.KogitoApp,
+		chain.Resources.BuildConfigS2I,
+	)
+	if err != nil {
+		chain.Error = err
+		return chain
 	}
+
+	// if gitURI is empty, instantiate a s2i binary build
+	if chain.KogitoApp.Spec.IsGitURIEmpty() {
```
sutaakar, line 97: Is this needed?
IMHO the binary build is initialized in every case in `buildConfigRuntimeBinaryBuilder` right below.

sutaakar:
Few minor comments, otherwise looks good.

sutaakar:
Now it is possible to use `oc start-build` to reupload content for source build, may be worth mentioning in readme.

cmd/kogito/command/deploy/deploy_service.go
```diff
@@ -157,9 +165,33 @@ func (i *deployCommand) Exec(cmd *cobra.Command, args []string) (err error) {
...
+				return t
+
+			default:
+				log.Warnf(message.KogitoAppAssetNotSupported, args[1])
+			}
+
+		}
+		// if the parameter does not match the expected input to perform a build from file, try a build from source instead
+		if userProvidedFile == nil {
+			log.Infof("Trying to perform an build from source...")
```
radtriste, line 191:
```suggestion
			log.Infof("Trying to perform a build from source...")
```

pkg/controller/kogitoapp/resource/requested_resources.go
spolti: for now I would say yes, I prefer change the binary build behaviour on the other jira that is opened to do the uploading thing on the binary builds as well.

pkg/controller/kogitoapp/resource/requested_resources_test.go
spolti: basically yes, but one has no giturl defined. so, both scenarios, must have the same buildConfigs getting created.

spolti:
> Now it is possible to use `oc start-build` to reupload content for source build, may be worth mentioning in readme.

Good point, let me add it.

pkg/client/openshift/buildconfig.go
```diff
@@ -101,6 +111,37 @@ func (b *buildConfig) TriggerBuild(bc *buildv1.BuildConfig, triggedBy string) (b
 	return true, nil
 }
 
+// TriggerBuildFromFile wil be called by kogito-cli when a build from file is performed.
+// When called a new build will be triggered with the request kogito resource or a tgz file.
+func (b *buildConfig) TriggerBuildFromFile(namespace string, bodyPost io.Reader, options *buildv1.BinaryBuildRequestOptions) (*buildv1.Build, error) {
+
+	result := &buildv1.Build{}
+	buildName := fmt.Sprintf("%s%s", options.Name, "-builder")
```
radtriste, line 119: shouldn't we be waiting for the `-binary` buildconfig ?

spolti: No, binary builds does not expect any s2i build.
In this case, build form asset we do expect a s2i build to build the kogito app using the kogito archetype.

radtriste: right

PR #253, opened 2020-03-30
KOGITO-1624 - Print also Examples for Scenario Outline when Show Scen…

MarianMacik:
…arios is selected

https://issues.redhat.com/browse/KOGITO-1624

* I have also removed unnecessary format configuration as it's already in the options.

Many thanks for submiting your Pull Request :heart: :racing_car: ! 

### DIFF
--- a/test/main_test.go
+++ b/test/main_test.go
@@ -121,7 +121,6 @@ func configureTestOutput() {
 		panic(fmt.Errorf("Error creating junit file: %v", err))
 	}
 
-	opt.Format = "junit"
 	opt.Output = io.MultiWriter(opt.Output, mainLogFile)
 }
 
@@ -182,7 +181,12 @@ func showScenarios(features []*gherkin.Feature) {
 		if len(ft.ScenarioDefinitions) > 0 {
 			mainLogger.Infof("Feature: %s", ft.Name)
 			for _, scenario := range ft.ScenarioDefinitions {
-				mainLogger.Info(framework.GetScenarioName(scenario))
+				scenarioMessage := framework.GetScenarioName(scenario)
+
+				if scenarioOutline, ok := scenario.(*gherkin.ScenarioOutline); ok {
+					scenarioMessage += fmt.Sprintf(" (Examples: %s)", strings.Join(framework.GetExamplesNames(scenarioOutline), ", "))
+				}
+				mainLogger.Info(scenarioMessage)
 			}
 		}
 	}
--- a/test/framework/util.go
+++ b/test/framework/util.go
@@ -111,6 +111,17 @@ func GetScenarioName(s interface{}) string {
 	return s.(*gherkin.ScenarioOutline).Name
 }
 
+func GetExamplesNames(outline *gherkin.ScenarioOutline) (names []string) {
+	for _ , examples := range outline.Examples {
+		examplesName := examples.Name
+		if len(examples.Name) == 0 {
+			examplesName = "Unnamed Examples"
+		}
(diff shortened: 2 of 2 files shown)


MarianMacik:
@sutaakar @radtriste can you have a look?
@spolti I cannot request reviews for PRs, can you probably add at least a `Triage` role to me? That should be enough. Also bear in mind that in other repositories all members of kiegroup/qe have the `Triage` role. Maybe would be good to align it.

spolti:
@MarianMacik you should have received a invite for now.

test/main_test.go
```diff
@@ -183,6 +182,10 @@ func showScenarios(features []*gherkin.Feature) {
 			mainLogger.Infof("Feature: %s", ft.Name)
 			for _, scenario := range ft.ScenarioDefinitions {
 				mainLogger.Info(framework.GetScenarioName(scenario))
+
+				if scenarioOutline, ok := scenario.(*gherkin.ScenarioOutline); ok {
+					mainLogger.Info(strings.Join(framework.GetExamplesNames(scenarioOutline), ", "))
```
sutaakar: Wouldn't it have more sense to put the examples information into same line as the scenario? For example into the brackets? When they are placed on new line it may be harder to recognize that it is not a new scenario but rather example related to previous scenario.

test/main_test.go
```diff
@@ -121,7 +121,6 @@ func configureTestOutput() {
 		panic(fmt.Errorf("Error creating junit file: %v", err))
 	}
 
-	opt.Format = "junit"
```
sutaakar: Why did you remove this line?

test/main_test.go
MarianMacik: I was considering it, but I thought that for some reason only feature file has the prefix `Feature` specified and others are left without any further information, so I decided to leave it that way.
Then I would also add the prefix `Scenario` on the same line as the current output is something like:

```
"------------------ SHOW SCENARIOS ------------------"}
{"Feature: Kogito Service Performance"}
{"Spring Boot Kogito Service Performance without persistence"}
{"Unnamed Examples"}
{"Spring Boot Kogito Service Performance with persistence"}
{"Unnamed Examples"}
```

As you can see, even now you cannot tell that the second one is a Scenario as only the feature file has the prefix. I would maybe go for prefixing all of that? On the other hand you may know that if it is not prefixed with the `Feature`, it's Scenario or Scenario Outline :)

Or I can do something like: `Spring Boot Kogito Service Performance without persistence (Examples: <Examples...>)`

test/main_test.go
MarianMacik: I noted it in the PR description, but it's true that it may have got lost among other things:

`I have also removed unnecessary format configuration as it's already in the options.`

The thing is that first I tried to change it to pretty in options struct on the top of the file, but then found out that this setting was overriding it. So why have 2 places where it is set up when we have it already in the options struct on the top.

sutaakar: oh, ok, I see.
Fine for me.

test/main_test.go
sutaakar: `Spring Boot Kogito Service Performance without persistence (Examples: <Examples...>)` would be nice to have, IMHO more readable than current implementation.

MarianMacik: done :wink:

ricardozanini:
What's the impact for 0.9.0? @radtriste @sutaakar ?

radtriste:
@ricardozanini no impact on 0.9.0. It is a dev functionality.

PR #257, opened 2020-04-01
[master][KOGITO-1044] - Kogito runtime performance and load tests

MarianMacik:
So guys,

it's finally here. A couple of things to be noted about the reporting:

* The metric which I am measuring now (create a given number of processes) is measured in the `executePostRequestsWithOptionalReportingUsingThreadsOnServiceWithPathAndBody` step function which calls `ExecuteHTTPRequestsInThreads` from the framework. At first, I tried to measure just the loop where requests are made inside the `ExecuteHTTPRequestsInThreads` function, but that would mean to push the `report` parameter and also the name of the metric into that function, so it wouldn't be as general as it is now. And when I measured it, the difference when measuring just requests and requests also with other nearby code was less than 1 ms, which is something I think we can live with as results are in seconds. Also the step for execution of many requests is available without reporting so if somebody in the future wants to create for example 10 process instances but doesn't care about the duration of it, they might find it helpful. Right now it's just me and I use the step with reporting.

* Anyway, I left included a log message with time measurement of the threads in the `ExecuteHTTPRequestsInThreads` so even if it is not used for performance reason, it is there as an additional information. Or maybe I should just print that threads have finished but duration won't be displayed? Wdyt?

* The metric name contains also the namespace and the scenario name so I know which scenario I am measuring. The only exception is that I cannot now distinguish between Native and Non Native (this is the name of the Examples table) as in the beforeScenario method Godog is passing whole ScenarioOutline with all examples tables. But so far it's OK as I can determine it by different times and also by order.

Currently, anything above 80k processes and with persistence is failing for 2 reasons, so these scenarios are commented out:

1. If I try more than 80k processes without persistence, I run out of heap as by default it is 1/4 of available memory, which is unrestricted for a project/pod so it seems that I am topping 4 GB as a node memory is 16 GB, which makes sense. So I will need to implement a way to specify it when deploying a service.

1. If I run anything with 80k processes or more with persistence, Infinispan runs out of heap as its heap is desperately low at 200 MB. These are the JVM options it is starting with: 
`JVM arguments = -Xmx200M -Xms200M -XX:MaxRAM=420M -Dsun.zip.disableMemoryMapping=true -XX:+UseSerialGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:+ExitOnOutOfMemoryError -XX:MetaspaceSize=32m -XX:MaxMetaspaceSize=96m -Djava.net.preferIPv4Stack=true -Djava.awt.headless=true -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dinfinispan.server.home.path=/opt/infinispan`

    So we have a couple of options:

    1. Leave it as it is because it is a default and I will somehow deploy my own Infinispan which has `extraJvmOpts` under `spec.container` in their CR. This will append -Xmx custom setting and since the last one wins, it will override the default. Already tried it manually on OpenShift.
    2. Increase the default deployed by Kogito Operator (KogitoInfra) to 1 GB for example. If it is sufficient for the performance tests I don't have to manually increase it afterwards.
    3. Provide a super cool way for configuring it via our Kogito Operator, so it will be configurable the same way we now configure for example the build images of services or the services themselves.

    I would personally go the way of the point ii. or iii. as I really believe 200 MB is too small for everything other than an example. Moreover, if we want to have our getting started experience as good as possible, it would be good to have at least 1 GB for storage by default so users can also try some real use cases out of the box and not just running 1 process instance.

1. Human tasks bulk completion step is currently commented as well. I have this on my list, it is there just as a note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ Table of Contents
             * [Running BDD tests with current branch](#running-bdd-tests-with-current-branch)
             * [Running BDD tests with custom Kogito Build images' version](#running-bdd-tests-with-custom-kogito-build-images-version)
             * [Running smoke tests](#running-smoke-tests)
+            * [Running performance tests](#running-performance-tests)
          * [Running the Kogito Operator locally](#running-the-kogito-operator-locally)
       * [Contributing to the Kogito Operator](#contributing-to-the-kogito-operator)
 
@@ -1147,6 +1148,20 @@ $ make run-smoke-tests [key=value]*
 
 It will run only tests tagged with `@smoke`.
 All options from BDD tests do also apply here.
+
+#### Running performance tests
+
+The BDD tests also provide performance tests. These tests are ignored unless you
+specifically provide the `@performance` tag or run:
+
+```bash
+$ make run-performance-tests [key=value]*
+```
+
+It will run only tests tagged with `@performance`. 
+All options from BDD tests do also apply here.
+
+**NOTE:** Performance tests should be run without concurrency.
 
 ### Running the Kogito Operator locally
 
--- a/test/features/kogito_service_performance.feature
+++ b/test/features/kogito_service_performance.feature
@@ -1,0 +1,152 @@
+# Commented code will be addressed by further enhancements:
+# https://issues.redhat.com/browse/KOGITO-1699
+# https://issues.redhat.com/browse/KOGITO-1700
+# https://issues.redhat.com/browse/KOGITO-1701
+
+@performance
+Feature: Kogito Service Performance
+
+  Background:
+    Given Namespace is created
+
+  @quarkus
+  Scenario Outline: Quarkus Kogito Service Performance without persistence
+    Given Kogito Operator is deployed
+    And Deploy quarkus example service "jbpm-quarkus-example" with native <native>
+    And Kogito application "jbpm-quarkus-example" has 1 pods running within <minutes> minutes
+    And HTTP GET request on service "jbpm-quarkus-example" with path "orders" is successful within 3 minutes
+
+    When <requests> HTTP POST requests with report using 100 threads on service "jbpm-quarkus-example" with path "orders" and body:
+      """json
+      {
+        "approver" : "john",
+        "order" : {
+          "orderNumber" : "12345",
+          "shipped" : false
+        }
+      }
+      """
+
+    Then HTTP GET request on service "jbpm-quarkus-example" with path "orders" should return an array of size <requests> within 1 minutes
+    And HTTP GET request on service "jbpm-quarkus-example" with path "orderItems" should return an array of size <requests> within 1 minutes
+    #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
+
+
+    Examples: Non Native
+      | native   | minutes | requests |
+      | disabled | 10      | 40000    |
+      | disabled | 10      | 80000    |
+#      | disabled | 10      | 160000   |
+#      | disabled | 10      | 320000   |
+
+    @native
+    Examples: Native
+      | native  | minutes | requests |
+      | enabled | 20      | 40000    |
+      | enabled | 20      | 80000    |
+#      | enabled | 20      | 160000   |
+#      | enabled | 20      | 320000   |
+
+#####
+
+  @quarkus
+  @persistence
+  Scenario Outline: Quarkus Kogito Service Performance with persistence
+    Given Kogito Operator is deployed with Infinispan operator
+    And Deploy quarkus example service "jbpm-quarkus-example" with native <native> and persistence
+    And Kogito application "jbpm-quarkus-example" has 1 pods running within <minutes> minutes
+    And HTTP GET request on service "jbpm-quarkus-example" with path "orders" is successful within 3 minutes
+
+    When <requests> HTTP POST requests with report using 100 threads on service "jbpm-quarkus-example" with path "orders" and body:
+      """json
+      {
+        "approver" : "john",
+        "order" : {
+          "orderNumber" : "12345",
+          "shipped" : false
+        }
+      }
+      """
+
+    Then HTTP GET request on service "jbpm-quarkus-example" with path "orders" should return an array of size <requests> within 1 minutes
+    And HTTP GET request on service "jbpm-quarkus-example" with path "orderItems" should return an array of size <requests> within 1 minutes
+    #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
+
+
+    Examples: Non Native
+      | native   | minutes | requests |
+      | disabled | 10      | 40000    |
+      | disabled | 10      | 80000    |
+#      | disabled | 10      | 160000   |
+#      | disabled | 10      | 320000   |
+
+    @native
+    Examples: Native
+      | native  | minutes | requests |
+      | enabled | 20      | 40000    |
+      | enabled | 20      | 80000    |
+#      | enabled | 20      | 160000   |
+#      | enabled | 20      | 320000   |
+
+#####
+
+  @springboot
+  Scenario Outline: Spring Boot Kogito Service Performance without persistence
+    Given Kogito Operator is deployed
+    And Deploy spring boot example service "jbpm-springboot-example"
+    And Kogito application "jbpm-springboot-example" has 1 pods running within <minutes> minutes
+    And HTTP GET request on service "jbpm-springboot-example" with path "orders" is successful within 3 minutes
+
+    When <requests> HTTP POST requests with report using 100 threads on service "jbpm-springboot-example" with path "orders" and body:
+      """json
+      {
+        "approver" : "john",
+        "order" : {
+          "orderNumber" : "12345",
+          "shipped" : false
+        }
+      }
+      """
+
+    Then HTTP GET request on service "jbpm-springboot-example" with path "orders" should return an array of size <requests> within 1 minutes
+    And HTTP GET request on service "jbpm-springboot-example" with path "orderItems" should return an array of size <requests> within 1 minutes
+    #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
+
+    Examples:
+      | minutes | requests |
+      | 10      | 40000    |
+      | 10      | 80000    |
+#      | 10      | 160000   |
+#      | 10      | 320000   |
+
+#####
+
+  @springboot
+  @persistence
+  Scenario Outline: Spring Boot Kogito Service Performance with persistence
+    Given Kogito Operator is deployed with Infinispan operator
+    And Deploy spring boot example service "jbpm-springboot-example" with persistence
+    And Kogito application "jbpm-springboot-example" has 1 pods running within <minutes> minutes
+    And HTTP GET request on service "jbpm-springboot-example" with path "orders" is successful within 3 minutes
+
+    When <requests> HTTP POST requests with report using 100 threads on service "jbpm-springboot-example" with path "orders" and body:
+      """json
+      {
+        "approver" : "john",
+        "order" : {
+          "orderNumber" : "12345",
+          "shipped" : false
+        }
+      }
+      """
+
+    Then HTTP GET request on service "jbpm-springboot-example" with path "orders" should return an array of size <requests> within 1 minutes
+    And HTTP GET request on service "jbpm-springboot-example" with path "orderItems" should return an array of size <requests> within 1 minutes
+    #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
+
+    Examples:
+      | minutes | requests |
+      | 10      | 40000    |
+      | 10      | 80000    |
+#      | 10      | 160000   |
+#      | 10      | 320000   |
--- a/test/framework/http.go
+++ b/test/framework/http.go
@@ -18,8 +18,19 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
+	"sync"
+	"time"
+)
+
+type HTTPRequestResult string
+
+const (
+	HTTPRequestResultSuccess HTTPRequestResult = "success"
+	HTTPRequestResultError   HTTPRequestResult = "error"
 )
 
 // WaitForSuccessfulHTTPRequest waits for an HTTP request to be successful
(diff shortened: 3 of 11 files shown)


ricardozanini:
> I would personally go the way of the point ii. or iii. as I really believe 200 MB is too small for everything other than an example. Moreover, if we want to have our getting started experience as good as possible, it would be good to have at least 1 GB for storage by default so users can also try some real use cases out of the box and not just running 1 process instance.

I believe that we shouldn't ever touch in those configurations. From the Kogito perspective, this is a "third party" infrastructure matter, I cannot say that 200MB is enough or too low for anything without proper scenario knowledge. On CRC environments for example, 200MB is more than enough. Which scenario we need to cover? Better to cover none, give the basic infra and the rest should be taken care of by the architecture/infra owner.

sutaakar:
> Do we want to always run those tests? Won't be better to have a flag/annotation in the run command to avoid running this over an over?

I think we should separate performance tests execution, being excluded by default from whole test suite run and activated by tag. There may be a separate command for this, similar to smoke tests.

test/features/kogito_service_performance.feature
```diff
@@ -0,0 +1,146 @@
+@performance
+Feature: Kogito Service Performance
+
+  Background:
+    Given Namespace is created
```
sutaakar, line 5: A new line below would increase readability. Right now background and scenario blocks are hard to distinguish on first look.

sutaakar:
```suggestion
    Given Namespace is created

```

radtriste:
> The metric name contains also the namespace and the scenario name so I know which scenario I am measuring. The only exception is that I cannot now distinguish between Native and Non Native (this is the name of the Examples table) as in the beforeScenario method Godog is passing whole ScenarioOutline with all examples tables. But so far it's OK as I can determine it by different times and also by order.

This will be changed with migration to new godog version. In the future, we will be able use placeholder in scenario outline name that will be replaced when compiled by the framework.
See: https://github.com/kiegroup/kogito-cloud-operator/pull/223
Still in dev but will have more time to finish after 0.9.1 and will update your code with new structure anyway

> I think we should separate performance tests execution, being excluded by default from whole test suite run and activated by tag. There may be a separate command for this, similar to smoke tests.

I do agree with this, could disable by default and add extra-param to the tests for those to be ran.
Can easily be handled there: https://github.com/kiegroup/kogito-cloud-operator/blob/5803e69a0b4442d10917cbe886f1804c4b938ff3/test/main_test.go#L90

test/framework/http.go
```diff
@@ -88,13 +96,73 @@ func ExecuteHTTPRequestWithUnmarshalledResponse(namespace, httpMethod, uri, path
...
 	if err != nil {
 		return false, err
 	}
+	io.Copy(ioutil.Discard, response.Body)  // Just read the response to be able to close the connection properly
+	defer response.Body.Close()
 	return checkHTTPResponseSuccessful(namespace, response), nil
 }
 
+// ExecuteHTTPRequestsInThreads executes given number of requests using given number of threads (Go routines).
+// Returns []string with the outcome of each thread ("success" or "error").
```
sutaakar, line 114: An idea for improvement, what about defining string type `HTTPResult` with two constants - `Success` and `Error`, that can replace these strings?

radtriste:
For those king of performance, we will need also to make sure scenario are not running in parallel (not yet available but is coming, so we could be preventive on that :)

test/framework/http.go
```diff
@@ -88,13 +96,73 @@ func ExecuteHTTPRequestWithUnmarshalledResponse(namespace, httpMethod, uri, path
...
+func ExecuteHTTPRequestsInThreads(namespace, httpMethod string, requestCount, threadCount int, routeURI, path, bodyFormat, bodyContent string) []string {
+	requestPerThread := requestCount / threadCount
+	results := make([]string, threadCount)
+	wg := &sync.WaitGroup{}
+	wg.Add(threadCount)
+
+	GetLogger(namespace).Info("Starting request threads")
+	startTime := time.Now()
+
+	for i := 0; i < threadCount; i++ {
```
radtriste, line 125: maybe change `i` to `threadId` or `threadIndex` ?
again for readiness

test/framework/http.go
```diff
@@ -88,13 +96,73 @@ func ExecuteHTTPRequestWithUnmarshalledResponse(namespace, httpMethod, uri, path
...
 	return checkHTTPResponseSuccessful(namespace, response), nil
 }
 
+// ExecuteHTTPRequestsInThreads executes given number of requests using given number of threads (Go routines).
+// Returns []string with the outcome of each thread ("success" or "error").
+// Useful for performance testing.
+func ExecuteHTTPRequestsInThreads(namespace, httpMethod string, requestCount, threadCount int, routeURI, path, bodyFormat, bodyContent string) []string {
+	requestPerThread := requestCount / threadCount
+	results := make([]string, threadCount)
+	wg := &sync.WaitGroup{}
```
radtriste, line 119: can we call it simply `waitGroup` or `threadWaitGroup` ? 
for more readiness of the code

test/framework/http.go
radtriste, line 114: Can we put `"sucess"` and `"error"` in public variables like `HTTP_REQUEST_SUCCESS_RESULT` and `HTTP_REQUEST_ERROR_RESULT` ?
so we can reuse when analyzing the `[]string`

sutaakar:
Apart from some minor comments looks good to me.
Good job.

ricardozanini:
Once we have this tests apart from the default run, I'll come back to review the rest. Many thanks @MarianMacik! Really good job indeed. We will need this more and more in the upcoming versions.

test/features/kogito_service_performance.feature
```diff
@@ -0,0 +1,146 @@
...
+        "order" : {
+          "orderNumber" : "12345",
+          "shipped" : false
+        }
+      }
+      """
+
+    Then HTTP GET request on service "jbpm-quarkus-example" with path "orders" should return an array of size <requests> within 1 minutes
+    And HTTP GET request on service "jbpm-quarkus-example" with path "orderItems" should return an array of size <requests> within 1 minutes
+    #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
```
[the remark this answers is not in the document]
MarianMacik: I would prefer to keep them if you don'ŧ mind as this is something as @disabled for now so I know I have to implement this in the near future.

test/features/kogito_service_performance.feature
```diff
@@ -0,0 +1,146 @@
...
+
+
+    Examples: Non Native
+      | native   | minutes | requests |
+      | disabled | 10      | 40000    |
+      | disabled | 10      | 80000    |
+#      | disabled | 10      | 160000   |
+#      | disabled | 10      | 320000   |
```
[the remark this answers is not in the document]
MarianMacik: Same here.

MarianMacik:
Everything should be addressed.

1. Regarding the Infinispan - OK, I understand, will modify it myself.

2. About not running performance tests by default - I was thinking about it, but first thought that we would just exclude it in the pipeline etc. Now it is ignored unless you provide `--performance` flag or specify `@performance` tag explicitly. I think it is better than reacting purely on the presence of `@performance` tag in the tags string as you can run for example `make run-performance-tests tags=springboot` which will run only springboot performance tests. But if that is too much, I can just check the passed tags string and act based on that.

3. Introduced new HTTPRequestResult type with 2 constants.

4. Added documentation.

README.md
```diff
@@ -1148,6 +1148,20 @@ $ make run-smoke-tests [key=value]*
 It will run only tests tagged with `@smoke`.
 All options from BDD tests do also apply here.
 
+#### Running performance tests
```
sutaakar, line 1151: Please add the chapter to the index in the beginning of the readme.

MarianMacik: Done.

sutaakar:
> And are we ok with a log message about the duration of those requests or should I delete it completely?

It is ok for me.

README.md
```diff
@@ -1148,6 +1149,20 @@ $ make run-smoke-tests [key=value]*
...
+specifically provide the `@performance` tag or run:
+
+```bash
+$ make run-performance-tests [key=value]*
+```
+
+It will run only tests tagged with `@performance`. 
+All options from BDD tests do also apply here.
+
+**NOTE:** Performance tests should be run without concurrency.
```
radtriste, line 1164: could we just disable concurrency into the code directly if performance is enabled ?
would just need to update `opt.Concurrency` to 1 into main_test.

test/framework/http.go
```diff
@@ -88,13 +103,73 @@ func ExecuteHTTPRequestWithUnmarshalledResponse(namespace, httpMethod, uri, path
...
+	io.Copy(ioutil.Discard, response.Body)  // Just read the response to be able to close the connection properly
+	defer response.Body.Close()
 	return checkHTTPResponseSuccessful(namespace, response), nil
 }
 
+// ExecuteHTTPRequestsInThreads executes given number of requests using given number of threads (Go routines).
+// Returns []HTTPRequestResult with the outcome of each thread (HTTPRequestResultSuccess or HTTPRequestResultError).
+// Useful for performance testing.
+func ExecuteHTTPRequestsInThreads(namespace, httpMethod string, requestCount, threadCount int, routeURI, path, bodyFormat, bodyContent string) []HTTPRequestResult {
+	requestPerThread := requestCount / threadCount
```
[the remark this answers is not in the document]
MarianMacik: Yes, considered it, but I would assume that nobody sane would specify 13 threads running 1000 requests. Anyway, in that case I would have to consider rounding it which means the metrics won't be accurate anyway. So I though we can use common sense here :)
But if you want, I can put there a panic in case the remainder is non-zero.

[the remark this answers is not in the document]
radtriste: ahah indeed.
a small panic or return error would be good so we know why it fails.
if not, someone may spend some time in the future just to figure that out (worst-case scenario)XD

test/main_test.go
```diff
@@ -89,22 +90,36 @@ func TestMain(m *testing.M) {
...
-		}
+		opt.Tags = appendAndIfNeeded(opt.Tags)
 		// Filter with smoke tag
 		opt.Tags += smokeTag
 	}
 
-	if !strings.Contains(opt.Tags, disabledTag) {
-		if len(opt.Tags) > 0 {
-			opt.Tags += " && "
+	if !strings.Contains(opt.Tags, performanceTag) {
```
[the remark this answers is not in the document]
MarianMacik: Can be. I was trying to keep it as simple as possible at first, but makes sense to guard it this way.

test/main_test.go
```diff
@@ -89,22 +90,36 @@ func TestMain(m *testing.M) {
 
 func configureTags() {
 	if config.IsSmokeTests() {
-		if len(opt.Tags) > 0 {
-			opt.Tags += " && "
-		}
+		opt.Tags = appendAndIfNeeded(opt.Tags)
```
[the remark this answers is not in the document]
MarianMacik: Yep, even nicer.

README.md
MarianMacik: Well, if we keep concurrent 1 in Makefile like it is now I think it's OK. Wdyt?

test/framework/http.go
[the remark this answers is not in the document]
MarianMacik: OK, done.

test/features/kogito_service_performance.feature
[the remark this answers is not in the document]
MarianMacik: Good idea, I originally wanted to create it afterwards but now the explanation is on the top of the file.

MarianMacik:
What about now? :)

README.md
radtriste: I think we should force it in Go code or ?
so we avoid any problem, whatever the user put as concurrency, it stays at 1.

test/framework/http.go
```diff
@@ -88,13 +103,76 @@ func ExecuteHTTPRequestWithUnmarshalledResponse(namespace, httpMethod, uri, path
...
 	}
+	io.Copy(ioutil.Discard, response.Body)  // Just read the response to be able to close the connection properly
+	defer response.Body.Close()
 	return checkHTTPResponseSuccessful(namespace, response), nil
 }
 
+// ExecuteHTTPRequestsInThreads executes given number of requests using given number of threads (Go routines).
+// Returns []HTTPRequestResult with the outcome of each thread (HTTPRequestResultSuccess or HTTPRequestResultError).
+// Useful for performance testing.
+func ExecuteHTTPRequestsInThreads(namespace, httpMethod string, requestCount, threadCount int, routeURI, path, bodyFormat, bodyContent string) []HTTPRequestResult {
```
radtriste, line 123: other `ExecuteHTTP[...]` are returning an error. it could be the same here instead of panic. and you can deal with it then into the steps

README.md
MarianMacik: To be honest I am also fine with the current way. I anyway want to have it as a Jenkins input parameter so I know that for 100% I am passing there 1.

radtriste: oki

test/framework/http.go
MarianMacik: No problem, it's not bad to have it unified. I have put the check after the measurement is done to not influence results if it is not needed.

test/framework/http.go
```diff
@@ -88,13 +103,77 @@ func ExecuteHTTPRequestWithUnmarshalledResponse(namespace, httpMethod, uri, path
...
+	results := make([]HTTPRequestResult, threadCount)
+	waitGroup := &sync.WaitGroup{}
+	waitGroup.Add(threadCount)
+
+	GetLogger(namespace).Info("Starting request threads")
+	startTime := time.Now()
+
+	for threadId := 0; threadId < threadCount; threadId++ {
+		client := createCustomClient()
+		go runRequestRoutine(threadId, waitGroup, client, namespace, httpMethod, requestPerThread, routeURI, path, bodyFormat, bodyContent, results)
```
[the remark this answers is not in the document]
ricardozanini: The same for all other functions that you're receiving all those parameters to execute a request. :stuck_out_tongue_closed_eyes:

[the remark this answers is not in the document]
MarianMacik: Well, that would need a complete overhaul of the framework because if I did it just for this method, then it would be inconsistent with the rest of the methods.
The only thing I am adding in my PR is the number of requests and number of threads. Shouldn't we address this outside of this PR?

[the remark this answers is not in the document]
sutaakar: +1 for me to address this in another PR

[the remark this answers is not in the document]
radtriste: +1 also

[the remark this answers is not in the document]
ricardozanini: np, please open a follow up JIRA then to arrange all those methods parameters.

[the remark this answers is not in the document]
MarianMacik: https://issues.redhat.com/browse/KOGITO-1729 opened, merging this one.

PR #259, opened 2020-04-01
[KOGITO-1670] - Bump to 0.9.1 version

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

See:
https://issues.redhat.com/browse/KOGITO-1670

Many thanks for submiting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.MD#sending-a-pull-request)
- [x] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket
- [ ] Your feature/bug fix has a unit test that verifies it
- [ ] You've tested the new feature/bug fix in an actual OpenShift cluster

PR #260, opened 2020-04-01
[master][KOGITO-1663] - Support for Kogito runtimes images micro version

ricardozanini:

See:
https://issues.redhat.com/browse/KOGITO-1663

~We depend on: https://github.com/kiegroup/kogito-cloud-operator/pull/259
(after merging, I'll rebase this one)~

We also need `0.9.x-latest` images in Quay to be able to run proper tests. (cc @spolti )


deploy/examples/jbpm-quarkus-example-binary-builds.yaml
```diff
@@ -3,7 +3,5 @@ kind: KogitoApp
 metadata:
   name: example-quarkus-flash
 spec:
-  # you only need to inform the correct image tag if you wish to build from a custom image
-  build:
-    imageVersion: 0.9.1 # will use it as a reference to create the imagestreams that the Operator uses
+  build: {}
```
sutaakar, line 6: IMHO this can be removed, keeping just `spec{}`

sutaakar:
@ricardozanini There is still image version defined in https://github.com/ricardozanini/kogito-cloud-operator/blob/kogito-1663/deploy/examples/jbpm-quarkus-example.yaml#L14 and https://github.com/ricardozanini/kogito-cloud-operator/blob/kogito-1663/deploy/examples/onboarding-example.yaml . Should it be removed too?

deploy/examples/jbpm-quarkus-example-binary-builds.yaml
ricardozanini: We can't, there's a YAML validation test that enforces this

ricardozanini:
> code seems good. will have a look once x-latest is available (maybe a rc one for an example, that we will remove just after tests ?)

code is not ok. I'll change the `GetRuntimeImageVersion` function to be more unit test friendly :)

ricardozanini:
On hold waiting for https://issues.redhat.com/browse/KOGITO-1673

radtriste:
For reference https://github.com/kiegroup/kogito-images/pull/124

ricardozanini:
@radtriste the tag is there, but with the wrong naming:
https://quay.io/repository/kiegroup/kogito-quarkus-ubi8-s2i?tab=tags

@spolti the correct tag name is `0.9.x-latest`, unless you guys changed your mind to `0.9-latest`. Just let me know if I need to update the PR or you will fix the tag naming.

radtriste:
we will adopt the `M.m` tag

radtriste:
JIRA updated

ricardozanini:
Updated and rebased, please take another look @radtriste

ricardozanini:
> seems there is a problem with the CLI as it has default values to `FullTag` for the runtime services
> 
> ok when applying CRs

It's not a problem. It's the default behavior since the first implementation, I'll change it.

pkg/infrastructure/mgmt_console.go
```diff
@@ -24,8 +24,8 @@ const (
 	DefaultMgmtConsoleName = "management-console"
 	// DefaultMgmtConsoleImageName ...
 	DefaultMgmtConsoleImageName = "kogito-management-console"
-	// DefaultMgmtConsoleImageFullTag ...
-	DefaultMgmtConsoleImageFullTag = "quay.io/kiegroup/" + DefaultMgmtConsoleImageName + ":latest"
+	// DefaultMgmtConsoleImageNoVersion is the default Management Console image name without version definition
+	DefaultMgmtConsoleImageNoVersion = "quay.io/kiegroup/" + DefaultMgmtConsoleImageName + ":"
```
Sgitario, line 28: I think it makes no sense to have this field like this any longer since we cannot use it by its own (we require the tag version always).
I would suggest to have a method instead:

```go
GetMgmtConsoleImageVersion(tag string) {
    return "quay.io/kiegroup/" + DefaultMgmtConsoleImageName + ":" + tag;
}
```

This is only a suggestion! The same for the rest of services.

ricardozanini: I'd say that `quay.io/kiegroup/` could be in an internal constant in the `infrasctructure` package as well. Let's keep this way for now and refactor in the feature since we're targeting 0.9.x and I've done too much changes already :sweat_smile:

ricardozanini:
@radtriste could you please take another look/test the CLI? Thanks!

pkg/infrastructure/mgmt_console.go
Sgitario: Okis!

PR #262, opened 2020-04-02
[master][KOGITO-1628] Cucumber tests: Merge the way we handle services into th…

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1628
Description: 
Since the merge of handling the services (data-index, jobs-service, management-console) as a generic implementation on operator's side, we could do it as well on test framework.

Idea would be to merge files:

framework/kogitodataindex.go
framework/kogitojobsservice.go
framework/kogitomgmtconsole.go

### DIFF
--- a/test/framework/kogitodataindex.go
+++ b/test/framework/kogitodataindex.go
@@ -15,94 +15,35 @@
 package framework
 
 import (
-	"fmt"
-	"strconv"
-
-	apps "k8s.io/api/apps/v1"
-
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
-	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
-	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/kiegroup/kogito-cloud-operator/test/config"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// InstallKogitoDataIndexService deploy the Kogito Data Index service
+// InstallKogitoDataIndexService install the Kogito Data Index service
 func InstallKogitoDataIndexService(namespace string, installerType InstallerType, replicas int) error {
-	GetLogger(namespace).Infof("%s install Kogito Data Index with %d replicas", installerType, replicas)
-	switch installerType {
-	case CLIInstallerType:
-		return cliInstallKogitoDataIndex(namespace, replicas)
-	case CRInstallerType:
-		return crInstallKogitoDataIndex(namespace, int32(replicas))
-	default:
-		panic(fmt.Errorf("Unknown installer type %s", installerType))
-	}
+	resource := newKogitoDataIndexResource(namespace, replicas)
+	// Persistence is already configured internally by the Data Index service, so we don't need to add any additional persistence step here.
+	return InstallServiceWithoutCliFlags(resource, installerType, "data-index")
 }
 
-func crInstallKogitoDataIndex(namespace string, replicas int32) error {
-	kogitoDataIndex := &v1alpha1.KogitoDataIndex{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      infrastructure.DefaultDataIndexName,
-			Namespace: namespace,
-		},
+// WaitForKogitoDataIndexService wait for Kogito Data Index to be deployed
+func WaitForKogitoDataIndexService(namespace string, replicas int, timeoutInMin int) error {
+	return WaitForService(namespace, getDataIndexServiceName(), replicas, timeoutInMin)
+}
+
+func getDataIndexServiceName() string {
+	return infrastructure.DefaultDataIndexName
+}
+
+func newKogitoDataIndexResource(namespace string, replicas int) *v1alpha1.KogitoDataIndex {
+	return &v1alpha1.KogitoDataIndex{
+		ObjectMeta: NewObjectMetadata(namespace, getDataIndexServiceName()),
 		Spec: v1alpha1.KogitoDataIndexSpec{
-			KogitoServiceSpec: v1alpha1.KogitoServiceSpec{
-				Replicas: &replicas,
-				Image:    getDataIndexImage(),
-			},
+			KogitoServiceSpec: NewKogitoServiceSpec(int32(replicas), config.GetDataIndexImageTag(), infrastructure.DefaultDataIndexImageNoVersion+infrastructure.GetRuntimeImageVersion()),
 		},
 		Status: v1alpha1.KogitoDataIndexStatus{
-			KogitoServiceStatus: v1alpha1.KogitoServiceStatus{
-				ConditionsMeta: v1alpha1.ConditionsMeta{
-					Conditions: []v1alpha1.Condition{},
-				},
-			},
+			KogitoServiceStatus: NewKogitoServiceStatus(),
 		},
 	}
-
-	if _, err := kubernetes.ResourceC(kubeClient).CreateIfNotExists(kogitoDataIndex); err != nil {
-		return fmt.Errorf("Error creating Kogito Data Index service: %v", err)
-	}
-	return nil
-}
-
-func cliInstallKogitoDataIndex(namespace string, replicas int) error {
-	cmd := []string{"install", "data-index"}
-	cmd = append(cmd, "--image", framework.ConvertImageToImageTag(getDataIndexImage()))
-	cmd = append(cmd, "--replicas", strconv.Itoa(replicas))
-
-	_, err := ExecuteCliCommandInNamespace(namespace, cmd...)
-	return err
-}
-
-func getDataIndexImage() v1alpha1.Image {
-	if len(config.GetDataIndexImageTag()) > 0 {
-		return framework.ConvertImageTagToImage(config.GetDataIndexImageTag())
-	}
-
-	image := framework.ConvertImageTagToImage(infrastructure.DefaultDataIndexImageNoVersion + infrastructure.GetRuntimeImageVersion())
-	image.Tag = config.GetServicesImageVersion()
-	return image
-}
-
-// GetKogitoDataIndexDeployment retrieves the running data index deployment
-func GetKogitoDataIndexDeployment(namespace string) (*apps.Deployment, error) {
-	return GetDeployment(namespace, infrastructure.DefaultDataIndexName)
-}
-
-// WaitForKogitoDataIndex waits that the jobs service has a certain number of replicas
-func WaitForKogitoDataIndex(namespace string, replicas, timeoutInMin int) error {
-	return WaitForOnOpenshift(namespace, "Kogito data index running", timeoutInMin,
-		func() (bool, error) {
-			dataIndex, err := GetKogitoDataIndexDeployment(namespace)
-			if err != nil {
-				return false, err
-			}
-			if dataIndex == nil {
-				return false, nil
-			}
-			return dataIndex.Status.Replicas == int32(replicas) && dataIndex.Status.AvailableReplicas == int32(replicas), nil
-		})
 }
--- a/test/framework/kogitojobsservice.go
+++ b/test/framework/kogitojobsservice.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Red Hat, Inc. and/or its affiliates
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
 //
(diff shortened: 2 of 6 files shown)


test/framework/kogitojobsservice.go
```diff
@@ -16,135 +16,82 @@ package framework
...
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/kiegroup/kogito-cloud-operator/test/config"
-	apps "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// InstallKogitoJobsService deploy the Kogito Jobs service
+// InstallKogitoJobsService returns the installation service for the Kogito Jobs Service
```
radtriste, line 29: This should be changed. it does not return

test/framework/kogitodataindex.go
```diff
@@ -15,94 +15,50 @@
...
-	switch installerType {
-	case CLIInstallerType:
-		return cliInstallKogitoDataIndex(namespace, replicas)
-	case CRInstallerType:
-		return crInstallKogitoDataIndex(namespace, int32(replicas))
-	default:
-		panic(fmt.Errorf("Unknown installer type %s", installerType))
-	}
+// InstallKogitoDataIndex returns the installation service for the Kogito Data Index
+func InstallKogitoDataIndex(namespace string, installerType InstallerType, replicas int) error {
```
radtriste, line 25: keep `Service in name`

test/framework/kogitodataindex.go
```diff
@@ -15,94 +15,50 @@
...
-	GetLogger(namespace).Infof("%s install Kogito Data Index with %d replicas", installerType, replicas)
-	switch installerType {
-	case CLIInstallerType:
-		return cliInstallKogitoDataIndex(namespace, replicas)
-	case CRInstallerType:
-		return crInstallKogitoDataIndex(namespace, int32(replicas))
-	default:
-		panic(fmt.Errorf("Unknown installer type %s", installerType))
-	}
+// InstallKogitoDataIndex returns the installation service for the Kogito Data Index
```
radtriste, line 24: it does not return

test/framework/kogitodataindex.go
```diff
@@ -15,94 +15,50 @@
...
+func InstallKogitoDataIndex(namespace string, installerType InstallerType, replicas int) error {
+	serviceName := getDataIndexServiceName()
+	imageTag := getDataIndexImageTag()
+	cliName := "data-index"
+	crBuilder := getDataIndexCrBuilder()
+	persistence := false
```
radtriste, line 30: can you put all of this in a structure ?
it would lower the number of arguments into `InstallService` method and would be more readable

radtriste, line 30: persistence is needed by data-index and will be set automatically
better to put true to avoid confusion

test/framework/kogitoserviceutils.go
```diff
@@ -0,0 +1,113 @@
...
+	case CRInstallerType:
+		return crInstall(namespace, serviceName, int32(replicas), image, crBuilder, persistence)
+	default:
+		panic(fmt.Errorf("Unknown installer type %s", installerType))
+	}
+}
+
+func buildImage(imageTag string) v1alpha1.Image {
+	image := framework.ConvertImageTagToImage(imageTag)
+	if len(image.Tag) == 0 || image.Tag == "latest" {
```
radtriste, line 69: this won't work. you set the `services-image-version` if it is not empty and if there is no `data-index-image-tag` or `jobs-service-image-tag` or `management-console-image-tag` defined (depending on the service you want or install)

The default image full tag has anyway a Tag, so it would not be overriden here
If you defined a specific service image tag, then it could be `latest` tag in it and you would still override it whereas you should not.

radtriste: we could have a `imageFullTag` (empty if no service image tag defined) and a `defautImageFullTag`

test/framework/kogitoserviceutils.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+			if deployment == nil {
+				return false, nil
+			}
+			return deployment.Status.Replicas == int32(replicas) && deployment.Status.AvailableReplicas == int32(replicas), nil
+		})
+}
+
+// CustomResourceBuilder helps to configure a CR resource used by the InstallService method.
+type CustomResourceBuilder struct {
```
radtriste, line 45: Can't we reuse v1alpha1.KogitoServcie interface ?
https://github.com/kiegroup/kogito-cloud-operator/blob/4d9ec51b70ea1abdfb7ffc65c9be124e706fb34c/pkg/apis/app/v1alpha1/kogitoservices_types.go#L25

test/framework/kogitomgmtconsole.go
```diff
@@ -15,100 +15,50 @@
...
-	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
-	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/kiegroup/kogito-cloud-operator/test/config"
-	apps "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// InstallKogitoManagementConsole deploy the Kogito Management Console
+// InstallKogitoManagementConsole returns the installation service for the Kogito Management Console
```
radtriste, line 24: this method does return only an error ;)

test/framework/kogitodataindex.go
Sgitario: @sutaakar what do you think about this? I made use of structures in the first version, so this would imply to have one more type with these five fields.

sutaakar: What about using KogitoDataIndex structure here? That way we can fill the values directly into the structure used to create the CR (in case of CR deployer).

Sgitario: Do you mean the structure "KogitoDataIndex" in kogitodataindex_types.go? I think this type is intended to represent an object in Openshift and this method is used to actually create the object. Therefore, as the KogitoDataIndex structure is coming from Openshift (with objects, Specs and so on), I don't really like to create one instance (like a stub or dummy instance). Morever, we would still need other values like the cli name or the cr types (or builder).

sutaakar: > Do you mean the structure "KogitoDataIndex" in kogitodataindex_types.go?

Yes.

> I think this type is intended to represent an object in Openshift and this method is used to actually create the object.

Yes, this type is intended to represent that resource in OpenShift. But this type is also used to create this resource. You use it too in this PR, see https://github.com/kiegroup/kogito-cloud-operator/pull/262/files#diff-080625abc1eacce2f2c206f2a6c3d4a0R53 . It serves both for the resource creation and for resource representation in OpenShift.

> Morever, we would still need other values like the cli name or the cr types (or builder).

The CLI name can be still passed as it is now, as it is the only information which cannot be retrieved from CR. cr types (Installer type?) doesn't need to be passed, it is retrieved from config. Builder doesn't have to be passed, as you pass already built resource.

test/framework/kogitojobsservice.go
Sgitario: Changed

test/framework/kogitodataindex.go
Sgitario: Renamed

test/framework/kogitodataindex.go
Sgitario: Changed

test/framework/kogitomgmtconsole.go
Sgitario: Changed

test/framework/kogitoserviceutils.go
Sgitario: I reverted this change.

test/framework/kogitoserviceutils.go
Sgitario: Done. Thanks for the tip!

test/framework/kogitodataindex.go
Sgitario: If we pass true, then the CLI command will have --enable-persistence which is not valid for Data Index.

Sgitario: Refactored. Let me know what you think about the new changes.

test/framework/kogitodataindex.go
```diff
@@ -15,94 +15,31 @@
...
-
-func getDataIndexImage() v1alpha1.Image {
-	if len(config.GetDataIndexImageTag()) > 0 {
-		return framework.ConvertImageTagToImage(config.GetDataIndexImageTag())
-	}
-
-	image := framework.ConvertImageTagToImage(infrastructure.DefaultDataIndexImageFullTag)
-	image.Tag = config.GetServicesImageVersion()
-	return image
+	return InstallServiceWithoutPersistence(resource, installerType, "data-index")
```
radtriste, line 35: can you comment why you install without persistence ?

test/framework/kogitodataindex.go
```diff
@@ -15,94 +15,31 @@
...
-			Namespace: namespace,
-		},
+	resource := &v1alpha1.KogitoDataIndex{
+		ObjectMeta: NewObjectMetadata(namespace, getDataIndexServiceName()),
 		Spec: v1alpha1.KogitoDataIndexSpec{
-			KogitoServiceSpec: v1alpha1.KogitoServiceSpec{
-				Replicas: &replicas,
-				Image:    getDataIndexImage(),
-			},
+			KogitoServiceSpec: NewKogitoServiceSpec(int32(replicas), config.GetDataIndexImageTag(), infrastructure.DefaultDataIndexImageFullTag),
```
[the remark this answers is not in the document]
radtriste: or could we call those methods `Create[...]` ?

i find it confusing with `New`, thought at first it was a specific functionality of Go, but maybe I am too much Java biased XD

test/framework/kogitodataindex.go
Sgitario: I don't know what you mean. This should be used only when the services do not support persistence (the enable-persistence in CLI for example).

test/framework/kogitojobsservice.go
```diff
@@ -16,94 +16,34 @@ package framework
...
+			KogitoServiceStatus: NewKogitoServiceStatus(),
+		},
 	}
-	return service, nil
-}
 
-// GetKogitoJobsServiceDeployment retrieves the running jobs service deployment
-func GetKogitoJobsServiceDeployment(namespace string) (*apps.Deployment, error) {
-	return GetDeployment(namespace, infrastructure.DefaultJobsServiceName)
+	setKogitoJobsServicePersistance(resource, persistence)
```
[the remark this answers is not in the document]
Sgitario: oh my! I had to re-read your comment a few times to realise the typo!

test/framework/kogitodataindex.go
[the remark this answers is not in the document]
Sgitario: I still thinks too much on Java way... need to learn more about Go! 
Will rename the methods to Create if this name convention fits better in Go.

test/framework/kogitodataindex.go
radtriste: just add a small comment because it can be confusing. data-index needs persistence per definition. but here we install it without.
just add a comment like `CLI for data-index does not support extra persistence installation` or similar

test/framework/kogitojobsservice.go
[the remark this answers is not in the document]
Sgitario: Fixed

test/framework/kogitodataindex.go
Sgitario: Added

test/framework/kogitodataindex.go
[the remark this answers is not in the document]
Sgitario: According to [Effective Go](https://golang.org/doc/effective_go.html#allocation_new), it's preferible to have NewXXX when we are creating new objects in Go. This is only about name convention.
However, I will try to do this block more readable.

spolti:
@Sgitario can you please rebase?

Sgitario:
Rebase done

spolti:
Is the circleci test still running for you guys as well?

radtriste:
indeed it stays in `Pending` state

spolti:
re-started the job, let's see if it gets fixed now.

PR #263, opened 2020-04-02
[master][KOGITO-1597] Refactor Cucumber tests to get rid of KogitoAppDeployment

sutaakar:

https://issues.redhat.com/browse/KOGITO-1597

### DIFF
--- a/pkg/apis/app/v1alpha1/kogitoapp_types.go
+++ b/pkg/apis/app/v1alpha1/kogitoapp_types.go
@@ -132,6 +132,16 @@ type KogitoAppBuildObject struct {
 	MavenMirrorURL string `json:"mavenMirrorURL,omitempty"`
 }
 
+// AddEnvironmentVariable adds new environment variable to build environment variables
+func (k *KogitoAppBuildObject) AddEnvironmentVariable(name, value string) {
+	env := corev1.EnvVar{
+		Name:  name,
+		Value: value,
+	}
+	k.Envs = append(k.Envs, env)
+	return
+}
+
 // KogitoAppServiceObject Data to define the service of the Kogito application
 // +k8s:openapi-gen=true
 type KogitoAppServiceObject struct {
--- a/pkg/apis/app/v1alpha1/kogitoservices_types.go
+++ b/pkg/apis/app/v1alpha1/kogitoservices_types.go
@@ -157,3 +157,13 @@ func (k *KogitoServiceSpec) GetResources() corev1.ResourceRequirements { return 
 func (k *KogitoServiceSpec) SetResources(resources corev1.ResourceRequirements) {
 	k.Resources = resources
 }
+
+// AddEnvironmentVariable adds new environment variable to service environment variables
+func (k *KogitoServiceSpec) AddEnvironmentVariable(name, value string) {
+	env := corev1.EnvVar{
+		Name:  name,
+		Value: value,
+	}
+	k.Envs = append(k.Envs, env)
+	return
+}
--- a/test/framework/kogitoapp.go
+++ b/test/framework/kogitoapp.go
@@ -18,11 +18,7 @@ import (
 	"fmt"
 	"strings"
 
-	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
-
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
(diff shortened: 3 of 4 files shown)


pkg/apis/app/v1alpha1/kogitoapp_types.go
```diff
@@ -132,6 +132,20 @@ type KogitoAppBuildObject struct {
 	MavenMirrorURL string `json:"mavenMirrorURL,omitempty"`
 }
 
+// AddEnvironmentVariable adds new environment variable to build environment variables
+func (k *KogitoAppBuildObject) AddEnvironmentVariable(name, value string) {
+	if k == nil {
```
Sgitario, line 137: Does it make sense this check? I mean using this method by:

```go
k.AddEnvironmentVariable where k is null
```

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -157,3 +157,17 @@ func (k *KogitoServiceSpec) GetResources() corev1.ResourceRequirements { return
 func (k *KogitoServiceSpec) SetResources(resources corev1.ResourceRequirements) {
 	k.Resources = resources
 }
+
+// AddEnvironmentVariable adds new environment variable to service environment variables
+func (k *KogitoServiceSpec) AddEnvironmentVariable(name, value string) {
+	if k == nil {
```
Sgitario, line 163: Same question as above.

pkg/apis/app/v1alpha1/kogitoapp_types.go
sutaakar: Good question, I have put it there as it is done i other methods.
Will try if it is usable.

sutaakar: Actually, the nil check has sense. For example this is a "valid" code:
```
	var spec *KogitoServiceSpec
	spec.AddEnvironmentVariable("qq", "ww")
```
Function `AddEnvironmentVariable` is called with `spec = nil`

sutaakar:
Cucumber tests passed (apart from Jobs service which is broken now)
Ready for review.

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -157,3 +157,17 @@ func (k *KogitoServiceSpec) GetResources() corev1.ResourceRequirements { return
 func (k *KogitoServiceSpec) SetResources(resources corev1.ResourceRequirements) {
 	k.Resources = resources
```
Sgitario, line 158: I think we should make a decision about this. Either we check null pointers everywhere, or nowhere or we change the approach about adding functions vs methods.
What do you think?

sutaakar: I think we can use this approach:
Preferred way of declaring such methods is to declare them on struct, not pointer.
If method is declared on pointer then it should handle case when pointer is nil.

@ricardozanini @radtriste  What do you think about it?

radtriste: good question. for me it should be checked before the method is called and not when it is called...

test/framework/kogitoapp.go
```diff
@@ -34,144 +30,92 @@ import (
...
-	Labels             map[string]string
-}
-
 // DeployService deploy a Kogito service
-func DeployService(namespace string, installerType InstallerType, kogitoAppDeployment KogitoAppDeployment) error {
-	GetLogger(namespace).Infof("%s deploy %s example %s with name %s, native %v, persistence %v, events %v and labels %v", installerType, kogitoAppDeployment.Runtime, kogitoAppDeployment.ContextDir, kogitoAppDeployment.AppName, kogitoAppDeployment.Native, kogitoAppDeployment.Persistence, kogitoAppDeployment.Events, kogitoAppDeployment.Labels)
+func DeployService(namespace string, installerType InstallerType, kogitoApp *v1alpha1.KogitoApp) error {
+	GetLogger(namespace).Infof("%s deploy %s example %s with name %s, native %v, persistence %v, events %v and labels %v", installerType, kogitoApp.Spec.Runtime, kogitoApp.Spec.Build.GitSource.ContextDir, kogitoApp.Name, kogitoApp.Spec.Build.Native, kogitoApp.Spec.EnablePersistence, kogitoApp.Spec.EnableEvents, kogitoApp.Spec.Service.Labels)
+
+	setupBuildImageStreams(kogitoApp)
```
radtriste, line 40: should be called in `crDeployService` or in `cliDeployService` you should check against this kogitoApp and not the config

test/steps/kogitoapp.go
```diff
@@ -173,3 +134,49 @@ func (data *Data) scaleKogitoApplicationToPodsWithinMinutes(name string, nbPods,
...
+	kogitoApp.Spec.Build.GitSource.URI = config.GetExamplesRepositoryURI()
+	kogitoApp.Spec.Build.GitSource.ContextDir = contextDir
+
+	if ref := config.GetExamplesRepositoryRef(); len(ref) > 0 {
+		kogitoApp.Spec.Build.GitSource.Reference = ref
+	}
+
+	// Add namespace for service discovery
+	// Can be removed once https://issues.redhat.com/browse/KOGITO-675 is done
+	kogitoApp.Spec.KogitoServiceSpec.AddEnvironmentVariable("NAMESPACE", namespace)
```
radtriste, line 153: don't we want to put that into the `framework.GetKogitoAppStub` ?
I mean it impacts all kogitoapp as soon as they use the service discovery... but we don't need to make the distinction

test/framework/kogitoapp.go
```diff
@@ -34,144 +30,92 @@ import (
...
+
+	if kogitoApp.Spec.Build.Native {
 		cmd = append(cmd, "--native")
 	}
 
-	// Add namespace for service discovery
-	// Can be removed once https://issues.redhat.com/browse/KOGITO-675 is done
-	cmd = append(cmd, "-e", fmt.Sprintf("NAMESPACE=%s", namespace))
-
 	if mavenMirrorURL := config.GetMavenMirrorURL(); len(mavenMirrorURL) > 0 {
```
radtriste, line 80: check against kogitoApp CR

test/framework/kogitoapp.go
```diff
@@ -34,144 +30,92 @@ import (
...
-	cmd = append(cmd, "--image-version", config.GetBuildImageVersion())
+	if buildImageVersion := config.GetBuildImageVersion(); len(buildImageVersion) > 0 {
+		cmd = append(cmd, "--image-version", buildImageVersion)
+	}
+	if buildS2iImageStreamTag := config.GetBuildS2IImageStreamTag(); len(buildS2iImageStreamTag) > 0 {
+		cmd = append(cmd, "--image-s2i", buildS2iImageStreamTag)
+	}
+	if buildRuntimeImageStreamTag := config.GetBuildRuntimeImageStreamTag(); len(buildRuntimeImageStreamTag) > 0 {
+		cmd = append(cmd, "--image-runtime", buildRuntimeImageStreamTag)
+	}
```
radtriste, line 112: check against kogitoApp

test/framework/kogitoapp.go
```diff
@@ -187,14 +131,13 @@ func DeployServiceFromExampleFile(namespace, exampleFile string) error {
...
-	kogitoApp := getKogitoAppStub(namespace, "name-should-be overwritten-from-yaml", nil)
+	kogitoApp := GetKogitoAppStub(namespace, "name-should-be overwritten-from-yaml")
 
 	// Apply content from yaml file
 	if err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(yamlContent), len([]byte(yamlContent))).Decode(kogitoApp); err != nil {
 		return fmt.Errorf("Error while unmarshalling file: %v ", err)
 	}
 
-	// Apply custom image configuration
 	setupBuildImageStreams(kogitoApp)
```
radtriste, line 141: shouldn't we put that in `GetKogitoAppStub` in fact ?

sutaakar: The point is that some yaml examples can declare image versions.

What should we do in such case? Test with images declared in yaml or test with images declared in test config?

sutaakar:
@radtriste setupBuildImageStreams and namespace declaration moved to GetKogitoAppStub, all values in cliDeployService are taken from KogitoApp.

test/framework/kogitoapp.go
radtriste: Ok I see your point. yeah should be let like that then :)

radtriste:
@sutaakar ahah about `setupBuildImageStreams` I was just thinking about letting it like it was :S 

that will impact (for now) the travel agency and it has been moved to kogito-examples (or still in progress but soon enough) and will then follow the different versions of Kogito for now.
so better to override with the test version.

sutaakar:
@Sgitario nil checks removed

PR #270, opened 2020-04-03
[master] [KOGITO-1627] Load default test config

radtriste:
https://issues.redhat.com/browse/KOGITO-1627\

This will allow PR and nightlies to load against predefined version (case of cr builds).
This can still be overloaded when calling the script if needed.

### DIFF
--- a/deploy/examples/jobs-service.yaml
+++ b/deploy/examples/jobs-service.yaml
@@ -1,0 +1,22 @@
+apiVersion: app.kiegroup.org/v1alpha1
+kind: KogitoJobsService
+metadata:
+  name: kogito-jobs-service
+spec:
+  # environment variables to set in the runtime container. Example: JAVA_OPTIONS: "-Dquarkus.log.level=DEBUG"
+  #env:
+    # - name: JAVA_OPTIONS
+    #    value: "-Dquarkus.log.level=DEBUG"
+  # number of pods to be deployed
+  replicas: 1
+  # Limits and requests for the Data Index pod
+  #memoryLimit: ""
+  #memoryRequest: ""
+  #cpuLimit: ""
+  #cpuRequest: ""
+  # details about the kafka connection
+  kafka:
+    useKogitoInfra: false
+  infinispan:
+    useKogitoInfra: false
+
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,7 +64,7 @@ pipeline {
             steps {
                 dir ("${WORKING_DIR}") {
                     sh """
-                        make run-smoke-tests operator_image=${OPENSHIFT_REGISTRY}/openshift/kogito-cloud-operator operator_tag=pr-\$(echo \${GIT_COMMIT} | cut -c1-7) maven_mirror=${MAVEN_MIRROR_REPOSITORY} concurrent=3
+                        make run-smoke-tests load_default_config=true operator_image=${OPENSHIFT_REGISTRY}/openshift/kogito-cloud-operator operator_tag=pr-\$(echo \${GIT_COMMIT} | cut -c1-7) maven_mirror=${MAVEN_MIRROR_REPOSITORY} concurrent=3
                     """
                 }
             }
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -64,7 +64,7 @@ pipeline {
             steps {
                 dir ("${WORKING_DIR}") {
                     sh """
-                        make run-tests operator_image=${OPENSHIFT_REGISTRY}/openshift/kogito-cloud-operator operator_tag=pr-\$(echo \${GIT_COMMIT} | cut -c1-7) maven_mirror=${MAVEN_MIRROR_REPOSITORY} concurrent=3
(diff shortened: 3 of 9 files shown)


Sgitario:
If I understood correctly, we can now have a file where to set the default config (".default_test_parameters"). This configuration was being set via parameters so far. 

With the parameter --ignore_default_config, we can ignore this file and load the parameters as we were doing so far. 

I wanted to add a suggestion here. Do we need this new parameter "ignore_default_config"? In order to avoid to add another parameter, what about this behaviour:
- We load **always** the .default_test_parameters file.
- We overwrite the parameters within the file by the ones set via the parameters

For example, if the default test parameters file contains:

```properties
tests.build-image-version=0.9.0-rc2
tests.services-image-version=0.9.0-rc2
tests.examples-uri=https://github.com/kiegroup/kogito-examples.git
tests.examples-ref=master
```

And, as an example, run the tests with the parameter --build-image-version=0.10.0

The final run will be executed using the parameters:

```properties
tests.build-image-version=0.10.0
tests.services-image-version=0.9.0-rc2
tests.examples-uri=https://github.com/kiegroup/kogito-examples.git
tests.examples-ref=master
```

I think it could be more intuitive this way and get rid of an extra parameter. However, this is only one suggestion!

test/.default_test_parameters
```diff
@@ -0,0 +1,4 @@
+tests.build-image-version=0.9.0-rc2
+tests.services-image-version=0.9.0-rc2
```
sutaakar, line 2: Is there a specific reason for using rc2 as 0.9.0 is released already?

radtriste: jobs-service 0.9.0 does not work...

radtriste: I could set directly the jobs-service 0.9.0-rc2 and let the others to 0.9.0

sutaakar: could be good, also please add a comment explaining this setup and when it can be removed

radtriste:
@MarianMacik I could set the default into Makefile indeed but I am not sure it should be in there to IMO (we could also remove the other parameters already set or migrate to config file)
@Sgitario I see your point but what if we want to test the default case, means no extra configuration (do not set the images version and examples repository is on branch `stable`)

for me that file is mainly for the pipeline because:
- the branches (master & 0.9.x) are already 0.10 or 0.9.1 and images to that version are not existing.
- also for the examples repository to branch 0.9.x and master for the tests

radtriste:
Or I set just a pipeline config and have a `--load_default_config` parameter instead that would be used by the pipeline.

Sgitario:
The problem I see is how the parameters and the values in the file coexist altogether. For example, if I don't know about the ignore parameter and I'm trying to overwrite the versions tag, it won't work. Therefore, I think having a "--load_default_config" as you suggested is more intuitive than having the "--ignore_default_config". But again, this is my point of view.

radtriste:
>  For example, if I don't know about the ignore parameter and I'm trying to overwrite the versions tag, it won't work.

It will. Default config parameters are set only if not existing already
https://github.com/radtriste/kogito-cloud-operator/blob/12bfd0a69f4236d2a8002e7de31d54b337566b0d/hack/run-tests.sh#L294

but yeah the load_default_config is maybe better.

test/.default_test_parameters
radtriste: updated

radtriste:
@sutaakar @MarianMacik @Sgitario can you have another look please ?
Changed to `load_default_config`

Makefile => Contains a default configuration for tests parameters, but no version information
.default_test_config => Contains version information. Need to use `load_default_config=true`. Can still be overridden via other parameters from the `make` command

deploy/examples/jobs-service.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: app.kiegroup.org/v1alpha1
+kind: KogitoJobsService
```
MarianMacik, line 2: Is this related to the issue?

radtriste: nope but it was missing. Can put it in another PR is needed.

ricardozanini:
@radtriste could you please take a look at `hack/bump-version.sh` to make sure that your newly added file with the versions is covered?

radtriste:
quite tricky as we could have very specific configuration but will have a look

deploy/examples/jobs-service.yaml
MarianMacik: Sure, np.

radtriste:
@ricardozanini updated `bump-versions.sh`.
for a new version, it will override the configuration with brand new values :)
if anything to keep, we will be able to see into the PR anyway but I doubt it when there is a new PR.
one thing will be that the pipeline will be green only if images are already existing in the defined version and the branch for kogito-examples exist

deploy/examples/jobs-service.yaml
ricardozanini: Actually we don't need this since we have: https://github.com/kiegroup/kogito-cloud-operator/blob/master/deploy/crds/app.kiegroup.org_v1alpha1_kogitojobsservice_cr.yaml

But since we have examples for all services, I don't see any problems having it.

ricardozanini:
Ready?

radtriste:
not yet

radtriste:
Final test passed

PR #291, opened 2020-04-09
KOGITO-1699 - Unify configuration for KogitoApp deployments in BDD tests

MarianMacik:
Configuration table for KogitoApp:
|first column|second column|third column|
|------|------|------|
|config|native|enabled/disabled|
|config|persistence|enabled/disabled|
|config|events|enabled/disabled|
|build-env|\<name>|\<value>|
|runtime-env|\<name>|\<value>|
|label|\<name>|\<value>|
|build-request|\<name>|\<value>|
|build-limit|\<name>|\<value>|
|runtime-request|\<name>|\<value>|
|runtime-limit|\<name>|\<value>|

Steps added by @Sgitario which need to check requests and limits will use the subset of that table.

By default, if `JAVA_OPTIONS` env variable is not set, it will be set to `-Xmx2G` so we have more stable resources assignment to test with.

Let me know what you think about it.

### DIFF
--- a/test/steps/kogitoapp.go
+++ b/test/steps/kogitoapp.go
@@ -15,8 +15,8 @@
 package steps
 
 import (
+	"fmt"
 	"path/filepath"
-	"strings"
 
 	"github.com/cucumber/godog"
 	"github.com/cucumber/godog/gherkin"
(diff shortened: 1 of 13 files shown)


test/steps/kogitoapp.go
```diff
@@ -60,82 +71,43 @@ func registerKogitoAppSteps(s *godog.Suite, data *Data) {
...
-}
-
-func (data *Data) deployQuarkusExampleServiceWithBuildResources(contextDir string, dt *gherkin.DataTable) error {
-	resources, err := assist.ParseMap(dt)
+func (data *Data) deploySpringBootExampleServiceWithConfiguration(contextDir string, table *gherkin.DataTable) error {
+	kogitoApp, err := getKogitoAppExamplesStub(data.Namespace, contextDir, table)
 	if err != nil {
 		return err
 	}
+	if kogitoApp.Spec.Build.Native {
```
radtriste, line 90: that is not ignored by the operator ?

test/steps/kogitoapp.go
```diff
@@ -185,31 +157,145 @@ func getKogitoAppExamplesStub(namespace, contextDir string, native, persistence,
...
-		profiles = append(profiles, "persistence")
-		kogitoApp.Spec.EnablePersistence = true
+	if err := configureKogitoAppFromTable(table, kogitoApp); err != nil {
+		return nil, err
+	}
+
+	return kogitoApp, nil
+}
+
+func configureKogitoAppFromTable(table *gherkin.DataTable, kogitoApp *v1alpha1.KogitoApp) error {
```
radtriste, line 167: can we move the parsing of datatable to kogitoapp and resources to another file ?

spolti:
Is this PR ready to merge?

test/steps/kogitoapp.go
```diff
@@ -185,31 +157,145 @@ func getKogitoAppExamplesStub(namespace, contextDir string, native, persistence,
...
+	return row.Cells[1].Value
+}
+
+func getThirdColumn(row *gherkin.TableRow) string {
+	return row.Cells[2].Value
+}
+
+// parseResourceRequirementsTable is useful for steps that check resource requirements, table is a subset of KogitoApp
+// configuration table
+func parseResourceRequirementsTable(table *gherkin.DataTable) (build, runtime *v1.ResourceRequirements, err error) {
```
Sgitario, line 276: I think we're duplicating how we parse the resources in three places:
- In the Build type (for build resources)
- In the Spec type (for runtime resources)
- In this method (to assert resources in tests)

What about only using this method to set the resources either in build and spec types, and also in the asserts?

MarianMacik: Yes, I am aware of that. The reason why I split it was to not parse the table multiple times. Wdyt? If I reused this part I would have to parse the table again after already reading it :) In my opinion, if we reuse this, there will be less duplication but the final code will be less readable than now.

test/steps/kogitoapp.go
MarianMacik: Hmm, good question. But anyway it would be good to inform a user of a framework that this configuration is illegal as it can happen that somebody will by mistake use `| config | native | true |` with Spring Boot. Wdyt?

radtriste: fair enough.
anyway once Spring will have native compilation (not today bu may come), this can be easily  removed ;)

test/steps/kogitoapp.go
MarianMacik: We can. Not sure how we want to have the layers/responsibilities split in the framework, but I understood it that everything about parsing GoDog etc. is right now in the steps package and the framework package is "unaware" of GoDog so it receives only things like URLs, KogitoApp etc.

So I can create utils file in the steps package as well, wdyt? There will be utils needed just for steps, so it will be nicely separated.

radtriste: I am good with it.
another possibility is to rename "step" files with suffix `_steps.go` and call this one `kogitoapp_config.go` ?
@sutaakar @Sgitario wdyt ?

Sgitario: Anything to split up the steps and the datatable config would help +1

test/steps/kogitoapp.go
Sgitario: I'm leaving this up to you, so ignore this if you disagree. From my point of view, if we only use this method in this file, I would prefer only having it in this file.

MarianMacik: So you are suggesting that this shouldn't go to the `utils` file with the configureKogitoAppFromTable function?

Sgitario: I'm suggesting to keep this method only and reuse it.

test/steps/kogitoapp.go
sutaakar: @radtriste Sounds good

test/steps/kogitoapp.go
MarianMacik: I know, but this confused me: 
> From my point of view, if we only use this method in this file, I would prefer only having it in this file.

Not sure what you meant by that.

Sgitario: I meant only using this method.

test/steps/kogitoapp.go
MarianMacik: Created utils file in steps directory as well where we can put anything for transformation of data from godog to the framework.

MarianMacik:
All comments should be addressed now.

PR #298, opened 2020-04-14
KOGITO-1415: Cucumber tests: enable scenario 'Discovery with onboardi…

Sgitario:
…ng' with native enabled

JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1415
Description:
With the changes in examples in https://issues.redhat.com/browse/KOGITO-1361, we made the onboarding build worked. However, the test still fails because it takes too much time. Increasing the CPU and memory made the test finished in around 10-15 min.

### DIFF
--- a/test/features/service_discovery.feature
+++ b/test/features/service_discovery.feature
@@ -20,8 +20,11 @@ Feature: Discovery with onboarding
       | label  | payments/date  | process  |
 
     And Deploy quarkus example service "onboarding-example/onboarding" with configuration:
-      | config | native     | <native> |
-      | label  | onboarding | process  |
+      | config        | native     | <native>                |
+      | label         | onboarding | process                 |
+      | build-request | cpu        | <build-request-cpu>     |
+      | build-request | memory     | <build-request-memory>  |
+      | build-limit   | cpu        | <build-limit-cpu>       |
 
     And Kogito application "hr" has 1 pods running within <minutes> minutes
     And Kogito application "payroll" has 1 pods running within <minutes> minutes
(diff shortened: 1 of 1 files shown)


test/features/service_discovery.feature
```diff
@@ -20,8 +20,11 @@ Feature: Discovery with onboarding
       | label  | payments/date  | process  |
 
     And Deploy quarkus example service "onboarding-example/onboarding" with configuration:
-      | config | native     | <native> |
-      | label  | onboarding | process  |
+      | config        | native     | <native> |
+      | label         | onboarding | process  |
+      | build-request | cpu        | 4        |
+      | build-request | memory     | 10Gi     |
+      | build-limit   | cpu        | 8        |
```
MarianMacik, line 27: Is this build that demanding? Just asking out of curiosity.

Sgitario: I think there are a couple of issues here:
- Why the onboarding project requires so amount of memory to build? Unlikely other quarkus apps which only require 4 Gi (the problem could be because the cloud-events dependency). However, I tried many tweaks and none made the native build require less memory :(
- The another issue is that we're triggering a few times the same build because of the bug [I raised last week ](https://issues.redhat.com/browse/KOGITO-1771)which make the resources more busy

sutaakar: @Sgitario I think it would have sense to parametrize the resources too (not just native flag) as non native build doesn't have so large requirements - can be set much lower.
That can save resources.

Sgitario: it makes totally sense this. Will do.

Sgitario: Done

PR #302, opened 2020-04-16
KOGITO-1831: BDD tests: Test binary build via CLI is working for a Qu…

Sgitario:
…arkus application

JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1831
Description: Add the quarkus version for binary builds (spring boot was already done)
As part of the ticked, we updated all the steps that use "spring boot" to "springboot".

### DIFF
--- a/test/features/binary_build.feature
+++ b/test/features/binary_build.feature
@@ -5,15 +5,36 @@ Feature: Use binary build to deploy the service
     Given Namespace is created
     And Kogito Operator is deployed
 
+  @smoke
   @springboot
-  # process-springboot-example is used here as it is the quickest example to be built
   Scenario: Deploy process-springboot-example using binary build
     Given Clone Kogito examples into local directory
     And Local example service "process-springboot-example" is built by Maven
 
-    When Create service "process-springboot-example"
+    When Create springboot service "process-springboot-example"
     And BuildConfig "process-springboot-example-binary" is created after 1 minutes
     And Start build with name "process-springboot-example-binary" from local example service path "process-springboot-example/target"
 
     Then Kogito application "process-springboot-example" has 1 pods running within 5 minutes
     And HTTP GET request on service "process-springboot-example" with path "orders" is successful within 2 minutes
+
+#####
+
+  @smoke
+  @quarkus
+  Scenario Outline: Deploy process-quarkus-example using binary build
+    Given Clone Kogito examples into local directory
+    And Local example service "process-quarkus-example" is built by Maven using profile "<profile>"
+
+    When Create quarkus service "process-quarkus-example" with configuration:
+      | config | native | <native> |
+    And BuildConfig "process-quarkus-example-binary" is created after 1 minutes
+    And Start build with name "process-quarkus-example-binary" from local example service path "process-quarkus-example/target"
+
+    Then Kogito application "process-quarkus-example" has 1 pods running within 5 minutes
+    And HTTP GET request on service "process-quarkus-example" with path "orders" is successful within 2 minutes
+
+  Examples:
+      | profile | native   |
+      | default | disabled |
+      | native  | enabled  |
--- a/test/steps/kogitoapp.go
+++ b/test/steps/kogitoapp.go
@@ -35,10 +35,9 @@ var assist = assistdog.NewDefault()
 
 func registerKogitoAppSteps(s *godog.Suite, data *Data) {
 	// Deploy steps
-	s.Step(`^Deploy quarkus example service "([^"]*)" with configuration:$`, data.deployQuarkusExampleServiceWithConfiguration)
-	s.Step(`^Deploy spring boot example service "([^"]*)" with configuration:$`, data.deploySpringBootExampleServiceWithConfiguration)
-	s.Step(`^Create service "([^"]*)"$`, data.createService)
-	s.Step(`^Create service "([^"]*)" with configuration:$`, data.createServiceWithConfiguration)
+	s.Step(`^Deploy (quarkus|springboot) example service "([^"]*)" with configuration:$`, data.deployExampleServiceWithConfiguration)
+	s.Step(`^Create (quarkus|springboot) service "([^"]*)"$`, data.createService)
+	s.Step(`^Create (quarkus|springboot) service "([^"]*)" with configuration:$`, data.createServiceWithConfiguration)
 	s.Step(`^Deploy service from example file "([^"]*)"$`, data.deployServiceFromExampleFile)
 
 	// DeploymentConfig steps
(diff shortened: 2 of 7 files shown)


test/features/binary_build.feature
```diff
@@ -17,3 +17,17 @@ Feature: Use binary build to deploy the service
```
radtriste, line 17: `# process-springboot-example is used here as it is the quickest example to be built`
that comment could be removed and we could put this test as a smoke test. wdyt ?

test/features/binary_build.feature
```diff
@@ -17,3 +17,17 @@ Feature: Use binary build to deploy the service
 
     Then Kogito application "process-springboot-example" has 1 pods running within 5 minutes
     And HTTP GET request on service "process-springboot-example" with path "orders" is successful within 2 minutes
+
+#####
+
+  @quarkus
+  Scenario: Deploy process-quarkus-example using binary build
```
radtriste, line 24: What about native mode ? :)

Sgitario: Ops, sorry. Moving back the ticket to In Progress.

test/features/binary_build.feature
Sgitario: deleted and added @smoke

test/features/binary_build.feature
Sgitario: Updated with native support

test/steps/kogitoapp.go
```diff
@@ -37,8 +37,10 @@ func registerKogitoAppSteps(s *godog.Suite, data *Data) {
 	// Deploy steps
 	s.Step(`^Deploy quarkus example service "([^"]*)" with configuration:$`, data.deployQuarkusExampleServiceWithConfiguration)
 	s.Step(`^Deploy spring boot example service "([^"]*)" with configuration:$`, data.deploySpringBootExampleServiceWithConfiguration)
-	s.Step(`^Create service "([^"]*)"$`, data.createService)
-	s.Step(`^Create service "([^"]*)" with configuration:$`, data.createServiceWithConfiguration)
+	s.Step(`^Create quarkus service "([^"]*)"$`, data.createQuarkusService)
+	s.Step(`^Create quarkus service "([^"]*)" with configuration:$`, data.createQuarkusServiceWithConfiguration)
+	s.Step(`^Create spring boot service "([^"]*)"$`, data.createSpringBootService)
+	s.Step(`^Create spring boot service "([^"]*)" with configuration:$`, data.createQuarkusServiceWithConfiguration)
```
Sgitario, line 43: We need to change the Spec.Runtime if Quarkus or Spring Boot. However, I'm not sure if having different steps for each is too much. We could either:
a) Get the runtime from the name (we're following the approach of using quarkus or springboot naming, but this could not work in all the cases... )
b) Maybe only having one step with configuration, but don't really like to have the config - runtime there, since it's a bit verbose.

What do you think?

test/features/binary_build.feature
```diff
@@ -5,15 +5,36 @@ Feature: Use binary build to deploy the service
...
+      | config | native | <native> |
+    And BuildConfig "process-quarkus-example-binary" is created after 1 minutes
+    And Start build with name "process-quarkus-example-binary" from local example service path "process-quarkus-example/target"
+
+    Then Kogito application "process-quarkus-example" has 1 pods running within 5 minutes
+    And HTTP GET request on service "process-quarkus-example" with path "orders" is successful within 2 minutes
+
+  Examples:
+      | profile   | native   |
+      | "default" | disabled |
```
Sgitario, line 39: Just a note here, the "default" maven profile does not exist but works fine and build the artifacts without errors. I think this way we could reuse the whole scenario to implement non native and native.

radtriste: please put the `"` into the step and not into the examples:
```
And Local example service "process-quarkus-example" is built by Maven using "<profile>"
[...]
 | profile   | native   |
 | default | disabled |
```

Sgitario:
Update about testing:
- It was failing using the 0.9.0 images because [this issue](https://issues.redhat.com/browse/KOGITO-1636) affecting to native build reported by Karel. So, this is expected.

- The "deploy_service_resources.feature" scenario was updated as well, so I did check that it was working fine.
- The "binary_build.feature" scenario has been tested using the cucumber automation using CLI and CR installers, and Spring Boot and Quarkus.

test/steps/kogitoapp.go
radtriste: I find it more readable to have the runtime into the text.

Solution `a` could be an idea. Check https://github.com/kiegroup/kogito-cloud-operator/blob/f67468dbe663617f2db0f00c6fb81db099191ac4/test/steps/kogitoapp.go#L32
same as `enabled|disabled`, we could allow only `spring boot|quarkus`
wdyt ?

Sgitario: I thought about it. The problem is that the v1alpha1.RuntimeType expects either quarkus or springboot strings (not "spring boot"). Perhaps we could remove the spaces before doing the:

```
v1alpha.RuntimeType(runtime)
```

Or maybe we could just use an if statement instead of the v1alpha.RuntimeType(xxx) method. 

What do you think?

radtriste: we could use `springboot` directly into the step. Still readable to me.
@sutaakar @MarianMacik wdyt ?

test/features/binary_build.feature
Sgitario: done

test/steps/kogitoapp.go
Sgitario: Other steps use "spring boot", don't really like to have some steps with "springboot" and others with "spring boot". Waiting for your feedback.

sutaakar: > we could use springboot directly into the step. Still readable to me.

Sounds good to me, we can unify it for all steps.

Sgitario: I've updated the steps using the optional string as Tristan suggested. Let me know your thoughts about this line:

```
kogitoApp.Spec.Runtime = v1alpha1.RuntimeType(strings.ReplaceAll(runtimeType, " ", ""))
```

Sgitario: Using springboot in all the steps sounds fair enough to me as well.

sutaakar: `springboot` seems better to me, as we reflect exact syntax to be used in CR

Sgitario: Done using optional string (quarkus|springboot) and updated the rest of steps from "spring boot" to "springboot".

test/steps/maven.go
```diff
@@ -22,6 +22,7 @@ import (
 // registerMavenSteps register all existing Maven steps
 func registerMavenSteps(s *godog.Suite, data *Data) {
 	s.Step(`^Local example service "([^"]*)" is built by Maven$`, data.localServiceBuiltByMaven)
+	s.Step(`^Local example service "([^"]*)" is built by Maven using "([^"]*)"$`, data.localServiceBuiltByMavenWithProfile)
```
radtriste, line 25:
```suggestion
	s.Step(`^Local example service "([^"]*)" is built by Maven using profile "([^"]*)"$`, data.localServiceBuiltByMavenWithProfile)
```
one last thing I did not noticed. I think it is cleared and it will be understandable when looking at the logs

Sgitario: added

test/steps/kogitoapp.go
```diff
@@ -36,9 +36,9 @@ var assist = assistdog.NewDefault()
 func registerKogitoAppSteps(s *godog.Suite, data *Data) {
 	// Deploy steps
 	s.Step(`^Deploy quarkus example service "([^"]*)" with configuration:$`, data.deployQuarkusExampleServiceWithConfiguration)
-	s.Step(`^Deploy spring boot example service "([^"]*)" with configuration:$`, data.deploySpringBootExampleServiceWithConfiguration)
-	s.Step(`^Create service "([^"]*)"$`, data.createService)
-	s.Step(`^Create service "([^"]*)" with configuration:$`, data.createServiceWithConfiguration)
+	s.Step(`^Deploy springboot example service "([^"]*)" with configuration:$`, data.deploySpringBootExampleServiceWithConfiguration)
```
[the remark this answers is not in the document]
Sgitario: Well spotted. Done.

test/steps/kogitoapp.go
```diff
@@ -53,42 +52,32 @@ func registerKogitoAppSteps(s *godog.Suite, data *Data) {
...
 
-func (data *Data) deploySpringBootExampleServiceWithConfiguration(contextDir string, table *gherkin.DataTable) error {
+func (data *Data) deployExampleServiceWithConfiguration(runtimeType, contextDir string, table *gherkin.DataTable) error {
 	kogitoApp, err := getKogitoAppExamplesStub(data.Namespace, contextDir, table)
 	if err != nil {
 		return err
 	}
-	if kogitoApp.Spec.Build.Native {
-		return fmt.Errorf("Spring Boot does not support native build")
+	if runtimeType != "quarkus" && kogitoApp.Spec.Build.Native {
```
radtriste, line 61: I would parse first and then do the check, this will avoid the string

Sgitario: changed

radtriste:
@Sgitario good to merge ?

Sgitario:
I've been running the tests with every change I've been adding and they worked. Go for it.

PR #310, opened 2020-04-21
KOGITO-1877: Cucumber tests should allow to change registry/namespace…

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1877
Description: 
Added new attributes to allow share the images when running the pipelines:

- build_image_registry
- build_image_namespace
- services_image_registry
- services_image_namespace

We will remove the defaults in the services and images parameters (registry, namespace and version). If not set, the operator will deal with it using the defaults.

### DIFF
--- a/pkg/infrastructure/kogito_images.go
+++ b/pkg/infrastructure/kogito_images.go
@@ -16,14 +16,19 @@ package infrastructure
 
 import (
 	"fmt"
-	"github.com/kiegroup/kogito-cloud-operator/version"
 	"strings"
+
+	"github.com/kiegroup/kogito-cloud-operator/version"
 )
 
 const (
 	versionSeparator = "."
 	// LatestTag the default name for latest image tag
 	LatestTag = "latest"
+	// DefaultImageRegistry the default services image repository
+	DefaultImageRegistry = "quay.io"
+	// DefaultImageNamespace the default services image namespace
+	DefaultImageNamespace = "kiegroup"
 )
 
 // GetRuntimeImageVersion gets the Kogito Runtime latest micro version based on the Operator current version
--- a/pkg/infrastructure/services/deployment_test.go
+++ b/pkg/infrastructure/services/deployment_test.go
@@ -15,14 +15,15 @@
 package services
 
 import (
+	"testing"
+
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/stretchr/testify/assert"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
-	"testing"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const defaultDataIndexImageFullTag = infrastructure.DefaultDataIndexImageNoVersion + "latest"
+var defaultDataIndexImageFullTag = infrastructure.GetRuntimeImageVersion() + ":latest"
 
 func Test_createRequiredDeployment_CheckQuarkusProbe(t *testing.T) {
 	kogitoService := &v1alpha1.KogitoDataIndex{
--- a/test/config/config.go
+++ b/test/config/config.go
@@ -18,8 +18,6 @@ import (
 	"flag"
 	"path/filepath"
 
-	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
-
 	"github.com/kiegroup/kogito-cloud-operator/version"
 )
 
@@ -42,14 +40,18 @@ type TestConfig struct {
 	cliPath           string
 
 	// runtime
-	servicesImageVersion string
-	dataIndexImageTag    string
-	jobsServiceImageTag  string
-	mgmtConsoleImageTag  string
+	servicesImageVersion   string
+	servicesImageNamespace string
+	servicesImageRegistry  string
+	dataIndexImageTag      string
+	jobsServiceImageTag    string
+	mgmtConsoleImageTag    string
 
 	// build
 	mavenMirrorURL       string
 	buildImageVersion    string
+	buildImageNamespace  string
+	buildImageRegistry   string
 	buildS2iImageTag     string
 	buildRuntimeImageTag string
 
@@ -104,14 +106,18 @@ func BindFlags(set *flag.FlagSet) {
 	set.StringVar(&env.cliPath, prefix+"cli-path", defaultCliPath, "Path to built CLI to test")
 
 	// runtime
-	set.StringVar(&env.servicesImageVersion, prefix+"services-image-version", infrastructure.GetRuntimeImageVersion(), "Set the services (jobs-service, data-index) image version")
+	set.StringVar(&env.servicesImageVersion, prefix+"services-image-version", "", "Set the services (jobs-service, data-index) image version")
+	set.StringVar(&env.servicesImageNamespace, prefix+"services-image-namespace", "", "Set the services (jobs-service, data-index) image namespace")
+	set.StringVar(&env.servicesImageRegistry, prefix+"services-image-registry", "", "Set the services (jobs-service, data-index) image registry")
 	set.StringVar(&env.dataIndexImageTag, prefix+"data-index-image-tag", "", "Set the Kogito Data Index image tag ('services-image-version' is ignored)")
 	set.StringVar(&env.jobsServiceImageTag, prefix+"jobs-service-image-tag", "", "Set the Kogito Jobs Service image tag ('services-image-version' is ignored)")
 	set.StringVar(&env.mgmtConsoleImageTag, prefix+"management-console-image-tag", "", "Set the Kogito Management Console image tag ('services-image-version' is ignored)")
 
 	// build
 	set.StringVar(&env.mavenMirrorURL, prefix+"maven-mirror-url", "", "Maven mirror url to be used when building app in the tests")
-	set.StringVar(&env.buildImageVersion, prefix+"build-image-version", infrastructure.GetRuntimeImageVersion(), "Set the build image version")
+	set.StringVar(&env.buildImageVersion, prefix+"build-image-version", "", "Set the build image version")
+	set.StringVar(&env.buildImageNamespace, prefix+"build-image-namespace", "", "Set the build image namespace")
+	set.StringVar(&env.buildImageRegistry, prefix+"build-image-registry", "", "Set the build image registry")
 	set.StringVar(&env.buildS2iImageTag, prefix+"build-s2i-image-tag", "", "Set the S2I build image full tag")
 	set.StringVar(&env.buildRuntimeImageTag, prefix+"build-runtime-image-tag", "", "Set the Runtime build image full tag")
 
@@ -190,6 +196,16 @@ func GetServicesImageVersion() string {
 	return env.servicesImageVersion
 }
 
+// GetServicesImageRegistry return the registry for the services images
+func GetServicesImageRegistry() string {
+	return env.servicesImageRegistry
+}
+
+// GetServicesImageNamespace return the namespace for the services images
+func GetServicesImageNamespace() string {
+	return env.servicesImageNamespace
+}
+
 // GetDataIndexImageTag return the Kogito Data Index image tag
 func GetDataIndexImageTag() string {
 	return env.dataIndexImageTag
@@ -215,6 +231,16 @@ func GetMavenMirrorURL() string {
 // GetBuildImageVersion return the version for the build images
 func GetBuildImageVersion() string {
 	return env.buildImageVersion
+}
+
+// GetBuildImageNamespace return the namespace for the build images
+func GetBuildImageNamespace() string {
+	return env.buildImageNamespace
+}
+
+// GetBuildImageRegistry return the registry for the build images
+func GetBuildImageRegistry() string {
+	return env.buildImageRegistry
 }
 
 // GetBuildS2IImageStreamTag return the tag for the s2i build image
--- a/test/framework/kogitoapp.go
+++ b/test/framework/kogitoapp.go
@@ -26,6 +26,9 @@ import (
 	"github.com/gobuffalo/packr/v2"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoapp/resource"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/kiegroup/kogito-cloud-operator/test/config"
 )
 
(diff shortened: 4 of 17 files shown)


Sgitario:
I've tested a few combinations with these changes:

- For build_image_registry and build_image_namespace:

- [x] without these new parameters. Neither the image runtime nor image S2I were set.
- [x] with build_image_registry only
- [x] with build_image_namespace only
- [x] with both build_image_registry and build_image_namespace

- For services_image_registry and services_image_namespace:

- [x] without these new parameters. Defaults were set.
- [x] with services_image_registry only
- [x] with services_image_namespace only
- [x] with both services_image_registry and services_image_namespace

pkg/infrastructure/kogito_images.go
```diff
@@ -32,6 +37,11 @@ func GetRuntimeImageVersion() string {
 	return getRuntimeImageVersion(version.Version)
 }
 
+// GetDefaultImage returns the default full image of the image name.
+func GetDefaultImage(imageName string) string {
```
sutaakar, line 41: A nitpick, can you please move `GetDefaultImage` up so `GetRuntimeImageVersion()` and `getRuntimeImageVersion(v string)` are next to each other, as they are related?

pkg/infrastructure/services/deployment_test.go
```diff
@@ -15,14 +15,15 @@
...
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/stretchr/testify/assert"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
-	"testing"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const defaultDataIndexImageFullTag = infrastructure.DefaultDataIndexImageNoVersion + "latest"
+const defaultDataIndexImageFullTag = "quay.io/kiegroup/" + infrastructure.DefaultDataIndexImageName + ":latest"
```
sutaakar, line 26:
```suggestion
var defaultDataIndexImageFullTag = infrastructure.GetDataIndexServiceDefaultImage()
```

test/steps/kogitoapp.go
```diff
@@ -30,7 +30,7 @@ func registerKogitoAppSteps(s *godog.Suite, data *Data) {
 	s.Step(`^Deploy (quarkus|springboot) example service "([^"]*)" with configuration:$`, data.deployExampleServiceWithConfiguration)
 	s.Step(`^Create (quarkus|springboot) service "([^"]*)"$`, data.createService)
 	s.Step(`^Create (quarkus|springboot) service "([^"]*)" with configuration:$`, data.createServiceWithConfiguration)
-	s.Step(`^Deploy service from example file "([^"]*)"$`, data.deployServiceFromExampleFile)
+	s.Step(`^Deploy (quarkus|springboot) service from example file "([^"]*)"$`, data.deployServiceFromExampleFile)
```
sutaakar, line 33: Could you please move this change into separate PR?
It doesn't seem to be related to the JIRA.

pkg/infrastructure/services/deployment_test.go
sutaakar, line 26:
```suggestion
var defaultDataIndexImageFullTag = infrastructure.GetDataIndexServiceDefaultImage()
```

test/steps/kogitoapp.go
Sgitario: I had to make this change in here since we need to set the runtimeType now in the constructor of the KogitoAppStub, so we can resolve the images to use accordingly based on the build type and the runtime type.

pkg/infrastructure/services/deployment_test.go
Sgitario: This is not the same. Runtime version will return 0.10.0 instead of latest and I didn't want to modify this test... What do you think?

pkg/infrastructure/kogito_images.go
Sgitario: Well spotted!

pkg/infrastructure/services/deployment_test.go
sutaakar: The tests here don't check the Data index tag itself, it is used just as a parameter for triggering the tested method.
The change above doesn't affect the tests.

test/steps/kogitoapp.go
sutaakar: Oh, ok, I have overlooked that.

pkg/infrastructure/services/deployment_test.go
Sgitario: ok ok. Will change this then.

test/steps/kogitoapp.go
Sgitario: Also, this step is not used anywhere, but I thought it was better to maintain it just in case you're creating your own scenarios for testing purposes.

test/framework/kogitoapp.go
```diff
@@ -211,6 +214,42 @@ func setupBuildImageStreams(kogitoApp *v1alpha1.KogitoApp) {
...
+	if len(config.GetBuildImageRegistry()) > 0 || len(config.GetBuildImageNamespace()) > 0 {
+		overwriteBuildImageRuntimeIfNotSet(kogitoApp)
+		overwriteBuildImageS2IIfNotSet(kogitoApp)
+	}
+
 	// If "KOGITO_BUILD_IMAGE_VERSION" is defined, it's taken into account, otherwise set the current version
 	kogitoApp.Spec.Build.ImageVersion = config.GetBuildImageVersion()
 }
+
+func overwriteBuildImageRuntimeIfNotSet(kogitoApp *v1alpha1.KogitoApp) {
```
sutaakar, line 226: What about adjusting this function to something like
```suggestion
func getBuildRuntimeImageStreamTag(kogitoApp *v1alpha1.KogitoApp) string {
```
And here you would completely resolve the tag?
The logic will stay the same, just in case` len(config.GetBuildRuntimeImageStreamTag()) > 0` you can return `config.GetBuildRuntimeImageStreamTag()`.
This would IMHO be more readable.

test/steps/kogitoapp.go
```diff
@@ -65,7 +63,7 @@ func (data *Data) createService(runtimeType, serviceName string) error {
 }
 
 func (data *Data) createServiceWithConfiguration(runtimeType, serviceName string, table *messages.PickleStepArgument_PickleTable) error {
-	kogitoApp := framework.GetKogitoAppStub(data.Namespace, serviceName)
+	kogitoApp := framework.GetKogitoAppStub(data.Namespace, runtimeType, serviceName)
 	if err := configureKogitoAppFromTable(table, kogitoApp); err != nil {
 		return err
 	}
```
sutaakar, line 69: Please remove `kogitoApp.Spec.Runtime = v1alpha1.RuntimeType(runtimeType)` on line 71 as it is not needed any more.

Sgitario: ops, sorry. Removed.

test/framework/kogitoapp.go
Sgitario: I refactor this change to try making more readable. Let me know your thoughts now.
Thanks!

test/framework/kogitoapp.go
```diff
@@ -209,8 +212,56 @@ func setupBuildImageStreams(kogitoApp *v1alpha1.KogitoApp) {
...
+
+	return ""
+}
+
+func isBuildImageRegistryOrNamespaceSet() bool {
+	return len(config.GetBuildImageRegistry()) > 0 || len(config.GetBuildImageNamespace()) > 0
+}
+
+func getBuildImage(image string) string {
+	registry := config.GetBuildImageRegistry()
```
radtriste, line 256: why not using the `Image` struc from operator framework package and use the convert method at the end ?

test/config/config.go
```diff
@@ -104,13 +109,17 @@ func BindFlags(set *flag.FlagSet) {
...
+	set.StringVar(&env.servicesImageRegistry, prefix+"services-image-registry", infrastructure.DefaultImageRegistry, "Set the services (jobs-service, data-index) image registry")
 	set.StringVar(&env.dataIndexImageTag, prefix+"data-index-image-tag", "", "Set the Kogito Data Index image tag ('services-image-version' is ignored)")
 	set.StringVar(&env.jobsServiceImageTag, prefix+"jobs-service-image-tag", "", "Set the Kogito Jobs Service image tag ('services-image-version' is ignored)")
 	set.StringVar(&env.mgmtConsoleImageTag, prefix+"management-console-image-tag", "", "Set the Kogito Management Console image tag ('services-image-version' is ignored)")
 
 	// build
 	set.StringVar(&env.mavenMirrorURL, prefix+"maven-mirror-url", "", "Maven mirror url to be used when building app in the tests")
 	set.StringVar(&env.buildImageVersion, prefix+"build-image-version", infrastructure.GetRuntimeImageVersion(), "Set the build image version")
+	set.StringVar(&env.buildImageNamespace, prefix+"build-image-namespace", "", "Set the build image namespace")
```
radtriste, line 121: I find it confusing to have default values for services but not for build ...

test/framework/kogitoapp.go
```diff
@@ -209,8 +212,56 @@ func setupBuildImageStreams(kogitoApp *v1alpha1.KogitoApp) {
...
+	if len(registry) == 0 {
```
radtriste, line 257: having default values for build, would allow us to avoid this check

test/framework/kogitoapp.go
```diff
@@ -209,8 +212,56 @@ func setupBuildImageStreams(kogitoApp *v1alpha1.KogitoApp) {
...
+		registry = infrastructure.DefaultImageRegistry
+	}
+
+	namespace := config.GetBuildImageNamespace()
+	if len(namespace) == 0 {
```
radtriste, line 262: same here for the check

test/config/config.go
Sgitario: I did this way at first, but then I realise that with default values, we'll always be sending the image runtime and image S2I attributes, and then we will stop covering the ability of the operator to set these values.

Also, I didn't want to change how was done at first. With services, we always send the image values. That's why services have defaults and images not. 

What do you think about it? I still can change it..

test/framework/kogitoapp.go
Sgitario: Read my reply above.

test/framework/kogitoapp.go
Sgitario: Read my reply above.

test/framework/kogitoapp.go
Sgitario: Do you mean, to use this method framework.ConvertImageToImageTag(image), right? Will change it.

test/config/config.go
radtriste: > With services, we always send the image values

what do you mean ?
In the code of tests, we always force the services image ?

> and then we will stop covering the ability of the operator to set these values.

This is the same for services in fact.

Indeed, we are setting most of the time the build and services.
But we should have same behavior for build and for services I think.

IMO, we should have default values for image version/namespace/registry and set only if given.
In that case, we can still execute the tests letting the operator set the default values.
wdyt ?

test/framework/kogitoapp.go
radtriste: yes that is right.
That would avoid any mistake when building the image name :)

Sgitario: Changed

test/config/config.go
Sgitario: As agreed, I removed the defaults in the services and images parameters (registry, namespace and version). If not set, the operator will deal with it using the defaults, so we can cover this functionality in the operator.

test/framework/kogitoserviceutils.go
```diff
@@ -122,7 +145,11 @@ func cliInstall(service v1alpha1.KogitoService, cliName string, cliFlags []strin
 		cmd = append(cmd, cliFlag)
 	}
 
-	cmd = append(cmd, "--image", framework.ConvertImageToImageTag(*service.GetSpec().GetImage()))
+	image := framework.ConvertImageToImageTag(*service.GetSpec().GetImage())
```
Sgitario, line 148: This change was not necessary, but it was weird to send only "--image" with a value. In fact, The operator is behaving differently when sending "--image" (without value) and when not sending the "--image". I raised a defect for this: https://issues.redhat.com/browse/KOGITO-1961

test/framework/kogitoapp.go
```diff
@@ -209,8 +213,66 @@ func setupBuildImageStreams(kogitoApp *v1alpha1.KogitoApp) {
...
+		namespace = infrastructure.DefaultImageNamespace
+	}
+
+	version := config.GetBuildImageVersion()
+	if len(version) == 0 {
+		version = infrastructure.GetRuntimeImageVersion()
+	}
+
+	return framework.ConvertImageToImageTag(v1alpha1.Image{
```
radtriste, line 272: why don't you create the image at the beginning of the method and update the fields directly instead of creating variables ?

test/framework/kogitoserviceutils.go
```diff
@@ -98,14 +99,36 @@ func GetCliFlags(persistence, events bool) []string {
...
+func newImageOrDefault(fullImage string, defaultImageName string) v1alpha1.Image {
 	if len(fullImage) > 0 {
 		return framework.ConvertImageTagToImage(fullImage)
 	}
 
-	image := framework.ConvertImageTagToImage(defaultImage)
-	image.Tag = config.GetServicesImageVersion()
-	return image
+	if len(config.GetServicesImageRegistry()) > 0 || len(config.GetServicesImageNamespace()) > 0 || len(config.GetServicesImageVersion()) > 0 {
+		registry := config.GetServicesImageRegistry()
```
radtriste, line 108: same here you, you could create the image with data from config and update the fields if empty ?

test/framework/kogitoserviceutils.go
radtriste: good for me

test/framework/kogitoapp.go
Sgitario: Okis, changed.

test/framework/kogitoserviceutils.go
Sgitario: Okis, changed.

pkg/infrastructure/kogito_images.go
```diff
@@ -16,16 +16,26 @@ package infrastructure
...
 	LatestTag = "latest"
+	// DefaultImageRegistry the default services image repository
+	DefaultImageRegistry = "quay.io"
+	// DefaultImageNamespace the default services image namespace
+	DefaultImageNamespace = "kiegroup"
 )
 
+// GetDefaultImage returns the default full image of the image name.
+func GetDefaultImage(imageName string) string {
+	return DefaultImageRegistry + "/" + DefaultImageNamespace + "/" + imageName + ":" + GetRuntimeImageVersion()
```
ricardozanini, line 36: Please avoid this kind of concatenation on functions body, try this:

```suggestion
	return fmt.sprintf( "%s/%s/%s:%s", DefaultImageRegistry, DefaultImageNamespace, imageName, GetRuntimeImageVersion())
```
Test it before commiting since I wrote from the top of my head. xD

Sgitario: Thanks for spotting. I missed to update this piece by using:

```
return framework.ConvertImageToImageTag(v1alpha1.Image{
		Domain:    DefaultImageRegistry,
		Namespace: DefaultImageNamespace,
		Name:      imageName,
		Tag:       GetRuntimeImageVersion(),
	})
```

Sgitario:
Rebased and tested again.

PR #315, opened 2020-04-22
[KOGITO-1347] Test Data index using external Infinispan

sutaakar:
https://issues.redhat.com/browse/KOGITO-1347

### DIFF
--- a/pkg/controller/kogitoinfra/infinispan/resource.go
+++ b/pkg/controller/kogitoinfra/infinispan/resource.go
@@ -15,6 +15,8 @@
 package infinispan
 
 import (
+	"reflect"
+
 	"github.com/RHsyseng/operator-utils/pkg/resource"
 	"github.com/RHsyseng/operator-utils/pkg/resource/read"
 	infinispan "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
@@ -23,7 +25,6 @@ import (
 	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/logger"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"reflect"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,9 +32,10 @@ import (
 
 const (
 	// InstanceName is the default name for the Infinispan provisioned instance
-	InstanceName     = "kogito-infinispan"
-	secretName       = "kogito-infinispan-credential"
-	identityFileName = "identities.yaml"
+	InstanceName = "kogito-infinispan"
+	secretName   = "kogito-infinispan-credential"
+	// IdentityFileName is the name of YAML file containing list of Infinispan credentials
+	IdentityFileName = "identities.yaml"
 	annotationKeyMD5 = "org.kie.kogito/infraInfinispanCredentialsFileHash"
 	replicasSize     = 1
 )
@@ -137,7 +139,7 @@ func newInfinispanCredentials(kogitoInfra *v1alpha1.KogitoInfra) (*v1.Secret, er
 		},
 		Type: v1.SecretTypeOpaque,
 		StringData: map[string]string{
-			identityFileName: credentialsFileData,
+			IdentityFileName: credentialsFileData,
 		},
 	}
 	return secret, nil
--- a/test/features/install_dataindex.feature
+++ b/test/features/install_dataindex.feature
@@ -17,3 +17,27 @@ Feature: Kogito Data Index
       }
     }
     """
+
+#####
+
+  @externalcomponent
+  @infinispan
+  Scenario: Install Kogito Data Index with persistence using external Infinispan
+    Given Infinispan instance "external-infinispan" is deployed with configuration:
+      | username | developer |
+      | password | mypass |
+
+    When Install Kogito Data Index with 1 replicas with configuration:
+      | infinispan | username | developer                 |
+      | infinispan | password | mypass                    |
+      | infinispan | uri      | external-infinispan:11222 |
+
+    Then Kogito Data Index has 1 pods running within 10 minutes
+    And GraphQL request on service "data-index" is successful within 2 minutes with path "graphql" and query:
+    """
+    {
+      ProcessInstances{
+        id
+      }
+    }
+    """
--- a/test/framework/infinispan.go
+++ b/test/framework/infinispan.go
@@ -1,0 +1,64 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+
+	infinispan "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+
+	infinispaninfra "github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoinfra/infinispan"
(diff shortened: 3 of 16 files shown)


test/features/install_dataindex.feature
```diff
@@ -17,3 +17,29 @@ Feature: Kogito Data Index
       }
     }
     """
+
+#####
+
+  @persistence
+  @external-infinispan
```
radtriste, line 24: shouldn't we use something more general ?
it seems we have 2 information in one tag here
we could put something like:
- @externalcomponent
- @infinispan

test/features/install_dataindex.feature
```diff
@@ -17,3 +17,29 @@ Feature: Kogito Data Index
...
+  Scenario: Install Kogito Data Index with persistence using external Infinispan
+    And Infinispan Secret "external-infinispan-secret" is created with credentials:
```
radtriste, line 26:
```suggestion
    Given Infinispan Secret "external-infinispan-secret" is created with credentials:
```

test/features/install_dataindex.feature
```diff
@@ -17,3 +17,29 @@ Feature: Kogito Data Index
...
+      | developer | mypass                      |
```
radtriste, line 27: could we use a Scenario Outline with Examples for this username/password ?

test/framework/infinispan.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"gopkg.in/yaml.v2"
+
+	infinispan "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+
+	infinispaninfra "github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoinfra/infinispan"
+)
+
+// DeployInfinispanInstance deploys an instance of Prometheus
```
radtriste, line 28:
```suggestion
// DeployInfinispanInstance deploys an instance of Infinispan
```

test/framework/kogitoserviceutils.go
```diff
@@ -108,23 +117,48 @@ func newImageOrDefault(fullImage string, defaultImage string) v1alpha1.Image {
...
+	}
 
-	_, err := ExecuteCliCommandInNamespace(service.GetNamespace(), cmd...)
+	_, err := ExecuteCliCommandInNamespace(serviceHolder.GetNamespace(), cmd...)
 	return err
 }
+
+//IsInfinispanUsernameSpecified Returns true if Infinispan username is specified
+func (serviceHolder *KogitoServiceHolder) IsInfinispanUsernameSpecified() bool {
+	return serviceHolder != nil && len(serviceHolder.Infinispan.Username) > 0
```
radtriste, line 163: so we do the nil check ? :D

test/steps/infinispan.go
```diff
@@ -0,0 +1,122 @@
...
+
+	if err := configureInfinispanFromTable(table, infinispan); err != nil {
+		return err
+	}
+
+	if err := framework.DeployInfinispanInstance(data.Namespace, infinispan); err != nil {
+		return err
+	}
+
+	return framework.WaitForPods(data.Namespace, "app", "infinispan-pod", 1, 3)
```
radtriste, line 62: can we rename that method to `WaitForPodsWithLabel` ? It took me some time to understand what was done here

test/features/install_dataindex.feature
sutaakar: umm, scenario outline would have sense in case we would execute scenario several times with different parameters, right?
In this case both credentials are applied together, first is for Infinispan "user" (used in Kogito), second is used by Infinispan operator

test/framework/kogitoserviceutils.go
sutaakar: will remove it :)

test/steps/infinispan.go
sutaakar: sure, will do

sutaakar:
@radtriste Updated according to comments.

test/features/install_dataindex.feature
sutaakar: Or do you mean to use Scenario outline to increase readability?

sutaakar:
The Infinispan CR is deployed as part of https://github.com/kiegroup/kogito-cloud-operator/pull/315/files#diff-99148689e28fa4a3003f7ae3db113e27R44 .
It is attached by specifying uri parameter in Kogito CR/`--infinispan-url` in CLI.

test/features/install_dataindex.feature
radtriste: no it was just to avoid mistakes but it may indeed increase complexity of reading it.

test/features/install_dataindex.feature
```diff
@@ -17,3 +17,29 @@ Feature: Kogito Data Index
       }
     }
     """
+
+#####
+
+  @externalcomponent
+  @infinispan
```
radtriste, line 24: did not pay attention here but maybe keep also the `@persistence` tag that we use in other scenarios ?

sutaakar: well, it depends on the meaning of `@persistence` tag:
Does it refer to testing of data being persisted?
Or does it refer to usage of Infinispan or other persistence provider?

Maybe we can add a list of available tags into readme together with description.

radtriste: good question :D

I was seeing it more like: `we use persistence in this scenario`
like we have the `@events` for the `we use events in this scenario`

For now it is only infinispan but maybe later it will be also other providers

radtriste: > Maybe we can add a list of available tags into readme together with description.

yeah we should add the definition of each tags as it seems we don't have the same definitions :)

sutaakar: This scenario is a bit blurry for `@persistence` tag as we use Infinispan here, but we don't test persistence explicitly.

radtriste: ok, let's let it like that.
created https://issues.redhat.com/browse/KOGITO-2010 for a tags review :)

test/features/install_dataindex.feature
```diff
@@ -17,3 +17,29 @@ Feature: Kogito Data Index
...
+  Scenario: Install Kogito Data Index with persistence using external Infinispan
+    Given Infinispan Secret "external-infinispan-secret" is created with credentials:
```
sutaakar, line 26: @radtriste When I think about it the Infinispan secret and Infinispan instance steps could be merged together. The Infinispan secret will hardly have any relevance outside of Infinispan instance.

The credentials could be moved to Infinispan instance configuration. The Infinispan configuration then won't reflect the explicit content of CR, but it will simplify usage of this step and make it more readable and understandable.

radtriste: sure they can be merged :)

sutaakar: On the other side we would lose flexibility in case we would like to specify more users for Infinispan.

sutaakar: But I guess it is not a common usecase

pkg/controller/kogitoinfra/infinispan/resource.go
```diff
@@ -23,17 +25,17 @@ import (
...
 
 const (
 	// InstanceName is the default name for the Infinispan provisioned instance
-	InstanceName     = "kogito-infinispan"
-	secretName       = "kogito-infinispan-credential"
-	identityFileName = "identities.yaml"
+	InstanceName = "kogito-infinispan"
+	secretName   = "kogito-infinispan-credential"
+	// IdentityFileName is the name of YAML file containing list of Infinispan credentials
+	IdentityFileName = "identities.yaml"
```
MarianMacik, line 38: Formatting of constants is a little bit off.

sutaakar:
Minor adjustments (merging Infinispan secret and Infinispan instance steps)

test/features/install_dataindex.feature
radtriste: IMO if we really need it, then we will refactor that one.

pkg/controller/kogitoinfra/infinispan/resource.go
sutaakar: The formatting changed because I have added there a comment.
Comment was added to satisfy go-lint as I changed the IdentityFileName to be exported.

MarianMacik: I see.

sutaakar:
Integrated secret creation into Infinispan instance step.
Rebased to latest master.

sutaakar:
@radtriste @MarianMacik Please take another look.

radtriste:
@sutaakar seems there are conflicts. can you please rebase ? :)
@MarianMacik are you good with that one ?

sutaakar:
Rebased and still works.
@radtriste Ready to be merged.

PR #317, opened 2020-04-22
[KOGITO-865] Small refactor to reuse same functions to deploy data index

ricardozanini:
See: https://issues.redhat.com/browse/KOGITO-865

Just a small adjustment to remove some duplications in code.

### DIFF
--- a/cmd/kogito/command/install/data_index.go
+++ b/cmd/kogito/command/install/data_index.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/deploy"
-	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/message"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/shared"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
@@ -152,7 +151,6 @@ func (i *installDataIndexCommand) InitHook() {
 }
 
 func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error {
-	log := context.GetDefaultLogger()
 	var err error
 	if i.flags.Project, err = shared.EnsureProject(i.Client, i.flags.Project); err != nil {
 		return err
@@ -231,12 +229,9 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error 
 		},
 	}
 
-	if err := kubernetes.ResourceC(i.Client).Create(&kogitoDataIndex); err != nil {
-		return fmt.Errorf(message.DataIndexErrCreating, err)
-	}
-
-	log.Infof(message.DataIndexSuccessfulInstalled, i.flags.Project)
-	log.Infof(message.DataIndexCheckStatus, kogitoDataIndex.Name, i.flags.Project)
-
-	return nil
+	return shared.
+		ServicesInstallationBuilder(i.Client, i.flags.Project).
+		OperatorInstalled().
+		InstallDataIndex(&kogitoDataIndex).
+		GetError()
 }
--- a/cmd/kogito/command/install/data_index.go
+++ b/cmd/kogito/command/install/data_index.go
@@ -16,6 +16,7 @@ package install
 
 import (
 	"fmt"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/message"
 
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/deploy"
@@ -162,17 +163,24 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error 
 		return nil
 	}
 
-	if i.flags.infinispan.UseKogitoInfra {
-		if available, err := infrastructure.IsInfinispanOperatorAvailable(i.Client, i.flags.Project); err != nil {
-			return err
-		} else if !available {
-			return fmt.Errorf("Infinispan Operator is not available in the Project: %s. Please make sure to install it before deploying Data Index without infinispan-url provided ", i.flags.Project)
+	// TODO this will be moved to the ServicesInstallerBuild API on KOGITO-911 PR
+	{
+		logger := context.GetDefaultLogger()
+		if i.flags.infinispan.UseKogitoInfra {
+			if infrastructure.IsInfinispanAvailable(i.Client) {
+				if available, err := infrastructure.IsInfinispanOperatorAvailable(i.Client, i.flags.Project); err != nil {
+					return err
+				} else if !available {
+					logger.Info(message.DataIndexInfinispanOperatorNotAvailable)
+				}
+			} else {
+				logger.Infof(message.DataIndexInfinispanNotAvailable, i.flags.Project)
+			}
 		}
-	}
-
-	if i.flags.kafka.UseKogitoInfra {
-		if available := infrastructure.IsStrimziAvailable(i.Client); !available {
-			return fmt.Errorf("Strimzi Operator is not available in the Project: %s. Please make sure to install it before deploying Data Index without Kafka provided ", i.flags.Project)
+		if i.flags.kafka.UseKogitoInfra {
+			if available := infrastructure.IsStrimziAvailable(i.Client); !available {
+				logger.Infof(message.DataIndexKafkaNotAvailable, i.flags.Project)
+			}
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


cmd/kogito/command/install/data_index.go
```diff
@@ -152,7 +151,6 @@ func (i *installDataIndexCommand) InitHook() {
 }
 
 func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error {
-	log := context.GetDefaultLogger()
 	var err error
 	if i.flags.Project, err = shared.EnsureProject(i.Client, i.flags.Project); err != nil {
 		return err
```
radtriste, line 156: Just found a bug on line#165
`if available, err := infrastructure.IsInfinispanOperatorAvailable(i.Client, i.flags.Project); err != nil `

when operator was not installed, then it does install it. but the dependent infinispan operator is not yet installed and this line will return `not available` and we get an error when trying directly to install the data-index...

```
$ oc new-project tradisso-test
$ kogito install data-index
infinispan-url not informed, Infinispan will be automatically deployed via Infinispan Operator
No Kafka information has been given. A Kafka instance will be automatically deployed via Strimzi Operator in the namespace. Kafka Topics will be created accordingly if they don't exist already
Error: Infinispan Operator is not available in the Project: tradisso-test. Please make sure to install it before deploying Data Index without infinispan-url provided 
```

cmd/kogito/command/shared/install_services.go
```diff
@@ -40,17 +42,28 @@ type ServicesInstallation interface {
...
 	SilentlyInstallOperator() ServicesInstallation
+	// OperatorInstalled assumes operator is already installed
+	OperatorInstalled() ServicesInstallation
 	// GetError return any given error during the installation process
 	GetError() error
 }
 
-func (s servicesInstallation) InstallDataIndex() ServicesInstallation {
+func (s servicesInstallation) OperatorInstalled() ServicesInstallation {
+	s.operatorInstalled = true
```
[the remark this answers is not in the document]
ricardozanini: I put some thoughts on that and end up with this, but I'll replace it to install the operator within this chain.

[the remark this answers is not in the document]
ricardozanini: Thinking again, I'd rather to go this route marking the operator as installed to not make another roundtrip to the server to check if the Operator is installed. 

In this use case (and also on jobs and mgmt console), it's better to have the operator to be installed first to have the other operators to at least have their CRD there. Read my other comment regarding the issue you've found.

cmd/kogito/command/install/data_index.go
ricardozanini: I'm afraid we can't do much about it at this time. On OpenShift I can look for the subscription on Infinispan, but on Kubernetes I won't be able to look at that.

I think we can let the CLI to install the Data Index (and the Operator now that we have the `ServiceInstallationBuilder`), and only blocks it if the CRD is not present.

Then we can emit a WARN to users that the Infinispan Types are there, but the operator might not be ready yet (since we haven't found any deployments of the given operator).

I'll update the PR to reflect this.

wdyt?

ricardozanini: PR updated. Take a look and tell me wdyt:

1. I'll keep the `OperatorInstalled` flag to not have to roundtrip again and leave the operator to be installed first 
2. The Infinispan / Kafka check now just logs in the console that the operators might not be ready, but continue the installation any way. Since we do not have a "hard" dependency, since the operator will handle that for us by enforcing the operators to be there, we do not annoy the user
3. This code will be moved to the framework as soon as we agreed with that in #318 PR

cmd/kogito/command/install/data_index.go
```diff
@@ -164,17 +163,24 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error
...
+				logger.Infof(message.DataIndexInfinispanNotAvailable, i.flags.Project)
+			}
 		}
-	}
-
-	if i.flags.kafka.UseKogitoInfra {
-		if available := infrastructure.IsStrimziAvailable(i.Client); !available {
-			return fmt.Errorf("Strimzi Operator is not available in the Project: %s. Please make sure to install it before deploying Data Index without Kafka provided ", i.flags.Project)
+		if i.flags.kafka.UseKogitoInfra {
+			if available := infrastructure.IsStrimziAvailable(i.Client); !available {
```
radtriste, line 181: Shouldn't we check for Strimzi operator as well ?

cmd/kogito/command/install/data_index.go
```diff
@@ -231,12 +237,9 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error
...
-		return fmt.Errorf(message.DataIndexErrCreating, err)
-	}
-
-	log.Infof(message.DataIndexSuccessfulInstalled, i.flags.Project)
-	log.Infof(message.DataIndexCheckStatus, kogitoDataIndex.Name, i.flags.Project)
-
-	return nil
+	return shared.
+		ServicesInstallationBuilder(i.Client, i.flags.Project).
+		OperatorInstalled().
```
radtriste, line 242: I would change this name to something like `SetOperatorInstalledOption(installed bool)` to be more readable

cmd/kogito/command/project/new_project.go
```diff
@@ -95,9 +96,9 @@ func (i *newProjectCommand) Exec(cmd *cobra.Command, args []string) error {
 
 		log.Infof("Project '%s' created successfully", ns.Name)
 
-		install := shared.ServicesInstallationBuilder(i.Client, ns.Name).SilentlyInstallOperator()
+		install := shared.ServicesInstallationBuilder(i.Client, ns.Name).SilentlyInstallOperatorIfNotExists()
```
radtriste, line 99: my point is that we have a different behavior here but the same problem would appear then if Infinispan/Strimzi CRDs are not there or ?

cmd/kogito/command/project/use_project.go
```diff
@@ -96,9 +97,9 @@ func (i *useProjectCommand) Exec(cmd *cobra.Command, args []string) error {
 
 		log.Infof(message.ProjectSet, i.flags.project)
 
-		install := shared.ServicesInstallationBuilder(i.Client, ns.Name).SilentlyInstallOperator()
+		install := shared.ServicesInstallationBuilder(i.Client, ns.Name).SilentlyInstallOperatorIfNotExists()
```
radtriste, line 100: Same here if Infinispan/Strimzi CRDs are not there

cmd/kogito/command/project/new_project.go
radtriste: We have different behaviors in code for

- `new_project --install-data-index`
- `use_project --install-data-index`
- `install data-index`

shouldn't we unify that ?

cmd/kogito/command/shared/install_services.go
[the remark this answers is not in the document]
radtriste: ok, for me but check my comment on unificiation of code ;)

cmd/kogito/command/install/data_index.go
radtriste: good for me. Check my other comment on `use_project` and `new_project`

cmd/kogito/command/project/new_project.go
ricardozanini: No, because `new-project` and `use-project` installs a default Data Index. And in #318 I have unified all the work. I need this one to get merged to continue the refactor there.

> my point is that we have a different behavior here but the same problem would appear then if Infinispan/Strimzi CRDs are not there or ?

No, won't appear because the Data Index will get deployed and the operator controller will handle the Infinispan/Kafka installation.

cmd/kogito/command/project/use_project.go
ricardozanini: See my other comment.

cmd/kogito/command/project/new_project.go
ricardozanini: The idea here is just to inform the user that the dependencies might not be there and what to expect.

That's one of the reasons why I advocate to remove this installation/dependecy from our operator.

cmd/kogito/command/install/data_index.go
ricardozanini: I agree that we can change the name, maybe `OperatorAlreadyInstalled()`. But I won't add a  parameter, this makes the code more complicated than needed.

radtriste: ok for `OperatorAlreadyInstalled()`

cmd/kogito/command/project/new_project.go
radtriste: oki

cmd/kogito/command/install/data_index.go
ricardozanini: We can't at this time (it's out of the scope of this JIRA). Also, their deployment is weird and the deployment hasn't a uniform name, we should put more thoughts/work on other way of doing that. Opened a JIRA to be tracked: https://issues.redhat.com/browse/KOGITO-2011

ricardozanini:
I'll merge this one and rebase #318. There we can unify the refactor and come up with a common ground.

PR #318, opened 2020-04-22
[KOGITO-911][KOGITO-910] CLI: Refactoring and having all services to be installed at once

ricardozanini:

Depends on #317 

See: 
https://issues.redhat.com/browse/KOGITO-910
https://issues.redhat.com/browse/KOGITO-911

Medium refactoring to accommodate the new features to install services via CLI when creating or setting a project to the context.

All in all there's no change in the installation feature itself, just providing a new way to have Jobs Service and Management Console via `new-project` and `use-projects` commands.

I'll rebase once #317 is merged. (in theory I should've included in here as well...)


ricardozanini:
@radtriste I believe I handled everything we talked, since I've removed the `OperatorInstalled` thing and handled everything in the `ServiceInstallationBuilder` for `new/use-project` and custom installations for each service such as `install data-index`.

Also when `new-project --install-data-index` is called chances are that the user WILL see the message complaining about Infinispan/Kafka CRD/deployment. We can't do much right now, but I agree that we can do better in the future like installing `KogitoInfra` and wait for it to finish the installation and then continue. Do you mind opening a JIRA to track this?

cmd/kogito/command/shared/install_services.go
```diff
@@ -15,11 +15,30 @@
...
+var defaultReplicas = int32(1)
+var defaultServiceStatus = v1alpha1.KogitoServiceStatus{ConditionsMeta: v1alpha1.ConditionsMeta{Conditions: []v1alpha1.Condition{}}}
+var defaultServiceSpec = v1alpha1.KogitoServiceSpec{Replicas: &defaultReplicas}
+
+type serviceInfoMessages struct {
+	errCreating                  string
+	installed                    string
+	checkStatus                  string
+	notInstalledNoKogitoOperator string
+	serviceExists                string
```
radtriste, line 39: not used ?

ricardozanini: It's indeed: https://github.com/kiegroup/kogito-cloud-operator/pull/318/files#diff-e787bd9ddec738a5228b3b1931d0edb1R109-R113

Basically all supporting services installations use it.

cmd/kogito/command/shared/install_services.go
```diff
@@ -79,36 +153,81 @@ func (s servicesInstallation) SilentlyInstallOperatorIfNotExists() ServicesInsta
...
+}
+
+// getDefaultJobsService gets the default Jobs Service instance
+func (s *servicesInstallation) getDefaultJobsService() v1alpha1.KogitoService {
+	return &v1alpha1.KogitoJobsService{
+		ObjectMeta: metav1.ObjectMeta{Name: infrastructure.DefaultJobsServiceName, Namespace: s.namespace},
+		Spec: v1alpha1.KogitoJobsServiceSpec{
+			KogitoServiceSpec: defaultServiceSpec,
+			InfinispanMeta:    v1alpha1.InfinispanMeta{InfinispanProperties: v1alpha1.InfinispanConnectionProperties{UseKogitoInfra: false}},
+			KafkaMeta:         v1alpha1.KafkaMeta{KafkaProperties: v1alpha1.KafkaConnectionProperties{UseKogitoInfra: true}},
```
radtriste, line 220: we need Kafka default for jobs-service ?

ricardozanini: To be able to communicate with Data Index, yes.

cmd/kogito/command/install/data_index.go
```diff
@@ -239,7 +210,8 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error
 
 	return shared.
 		ServicesInstallationBuilder(i.Client, i.flags.Project).
-		OperatorInstalled().
+		SilentlyInstallOperatorIfNotExists().
+		WarnIfDependenciesNotReady(i.flags.infinispan.UseKogitoInfra, i.flags.kafka.UseKogitoInfra).
 		InstallDataIndex(&kogitoDataIndex).
```
radtriste, line 215: There is a different behavior here between the `kogito new|use-project --install-XXX` command and the `kogito install XXX` command.
Indeed, into the `kogito install XXX` command, the default image is set to `infrastructure.DefaultDataIndexImageNoVersion+infrastructure.GetRuntimeImageVersion()` where as in the `kogito new|use-project --install-XXX`, there is none by default.

So if I use another CLI version than the operator then the installation, with `kogito install XXX` command, it tries to install the version from the CLI.
For example in my tests CLI is version 0.10 and OperatorHub is version 0.9.1. So executing `kogito install data-index` with CLI on version 0.10, it tries to get the data-index version 0.10, which does not exist yet.
But this could be in another way, if user did not update to latest CLI, for example still using version 0.8.0, then it would install version 0.8 instead of 0.9 expected.

IMO, we should let the [image flag](https://github.com/kiegroup/kogito-cloud-operator/blob/f469c6f5f738fdadde943624a7d8a8a2464a4817/cmd/kogito/command/install/data_index.go#L143) as empty string by default, instead of `infrastructure.DefaultDataIndexImageNoVersion+infrastructure.GetRuntimeImageVersion()`. Then the image will be set by the operator itself and not the CLI.
And this of course for data-index, jobs-service and management-console.

wdyt ?

radtriste:
Also, we provide the `--install-data-index` on `kogito install operator` command. 
Shouldn't we provide the others `--install-XXX` as well ?

```
$ kogito install operator -h
Installs the Kogito Operator via custom Kubernetes resources. This feature won't create custom subscriptions with the OLM.

Usage:
  kogito install operator [flags]

Examples:
install operator -p my-project

Flags:
  -f, --force                When set, the operator will be installed in the current namespace using a custom image, e.g. quay.io/kiegroup/kogito-cloud-operator:my-custom-tag
  -h, --help                 help for operator
  -i, --image string         The operator image (default "quay.io/kiegroup/kogito-cloud-operator:0.10.0")
      --install-data-index   Installs the default instance of Data Index being provisioned by the Kogito Operator in the project
  -p, --project string       The project name where the operator will be deployed

Global Flags:
  -o, --output string   output format (when defined, 'json' is supported)
  -v, --verbose         verbose output
      --version         display version
```

cmd/kogito/command/shared/install_services.go
ricardozanini: I'll change to `false` then

ricardozanini:
@radtriste could you take another look? I believe I addressed them all.

cmd/kogito/command/install/kogito_operator.go
```diff
@@ -22,10 +22,12 @@ import (
...
 type installKogitoOperatorFlags struct {
-	namespace        string
-	image            string
-	installDataIndex bool
-	force            bool
+	namespace          string
+	image              string
+	installDataIndex   bool
+	installJobsService bool
+	installMgmtConsole bool
```
radtriste, line 29: missing installAll ?

ricardozanini: I thought about that, but I don't think we should add it in here. Won't be a little weird from UX perspective? 

```
kogito install operator --install-all-services
```
IDK. Again, I think we should bring UX team to these discussions and review the entire CLI commands.

radtriste: agreed. but for know, I think we should just have consistency :)
`--install-all-services` could be good

ricardozanini: OK, I'll add it.

PR #323, opened 2020-04-24
[KOGITO-1771] Checking build status before triggering a new build

ricardozanini:

See:
https://issues.redhat.com/browse/KOGITO-1771

Small refactoring introducing the `build` package. It's basically the same thing, but we will follow this up on [KOGITO-1999](https://issues.redhat.com/browse/KOGITO-1999).

I think this solved the issue, at least on my side, never happened again.


pkg/controller/kogitoapp/build/trigger.go
```diff
@@ -0,0 +1,109 @@
...
+	if !t.isBuildRunning(builds) {
+		if _, err = openshift.BuildConfigC(t.client).TriggerBuild(t.buildConfig, t.triggeredBy); err != nil {
+			return
+		}
+		result.Started = true
+	}
+	return
+}
+
+func (t *trigger) isBuildRunning(builds *v1alpha1.Builds) bool {
```
sutaakar, line 97: The `trigger` pointer is not used in this method, it can be removed.

pkg/controller/kogitoapp/build/trigger.go
```diff
@@ -0,0 +1,109 @@
...
+
+// TriggerResult structure to hold build trigger information
+type TriggerResult struct {
+	Started   bool
+	BuildName string
+}
+
+type trigger struct {
+	triggeredBy   string
+	labelSelector map[string]string
```
sutaakar, line 51: This parameter isn't used.

pkg/util/maps.go
```diff
@@ -0,0 +1,96 @@
...
+func MapContainsMap(source, expected map[string]string) bool {
+	if len(source) == 0 && len(expected) == 0 {
+		return false
+	}
+	if len(source) > 0 && len(expected) == 0 {
+		return false
+	}
+	if len(source) == 0 && len(expected) > 0 {
+		return false
+	}
```
MarianMacik, line 38: Just for better readability/less exit points of a function, can we do something like:
```go
if len(source) == 0 && len(expected) == 0 ||
   len(source) > 0 && len(expected) == 0 ||
   len(source) == 0 && len(expected) > 0 {
   return false
}
```

pkg/util/maps.go
```diff
@@ -0,0 +1,96 @@
...
+func FromStringsKeyPairToMap(array []string) map[string]string {
+	if array == nil || len(array) == 0 {
+		return nil
+	}
+	keyPairMap := map[string]string{}
+	for _, item := range array {
+		keyPair := strings.SplitN(item, keyPairSeparator, 2)
+		if len(keyPair) == 0 {
+			break
+		}
```
MarianMacik, line 66: This if is redundant as from the documentation this could only happen if both the `item` and `keyPairSeparator` were empty.

```
Split slices s into all substrings separated by sep and returns a slice of the substrings between those separators.
If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.
If sep is empty, Split splits after each UTF-8 sequence.
If both s and sep are empty, Split returns an empty slice.
```

pkg/util/maps.go
```diff
@@ -0,0 +1,96 @@
...
+// FromMapToString converts a map into a string format such as key1=value1,key2=value2
+func FromMapToString(labels map[string]string) string {
+	b := new(bytes.Buffer)
+	for k, v := range labels {
+		fmt.Fprintf(b, "%s=%s%s", k, v, pairSeparator)
+	}
+	return strings.TrimSuffix(b.String(), pairSeparator)
+}
+
+// FromStringsKeyPairToMap converts a string array in the key/pair format (key=value) to a map. Unconvertable strings will be skipped.
```
MarianMacik, line 56: If inconvertible strings are skipped, shouldn't we use continue in the loop?

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -287,29 +288,33 @@ func (r *ReconcileKogitoApp) updateKogitoAppStatus(request *reconcile.Request, i
...
 	}
 
-	// Trigger only the S2I builds, they will trigger runtime builds
-	if bcMap[string(kogitores.BuildTypeS2I)] != nil {
-		for _, bc := range bcMap[string(kogitores.BuildTypeS2I)] {
-			log.Infof("Buildconfigs are created, triggering build %s", bc.GetName())
-			if _, err := openshift.BuildConfigC(r.client).TriggerBuild(&bc, instance.Name); err != nil {
-				return err
-			}
+	// we don't have a s2i build, let's trigger the build source from image since it could be changed by an image stream
```
MarianMacik, line 304: The comment is quite confusing to me. Is this for binary builds? If so, then what "could be changed by an image stream"?

sutaakar:
List of tests being done:
- S2I build, update image - passed
- runtime build, update image - blocked by https://issues.redhat.com/browse/KOGITO-2017
- binary build, update image - blocked by https://issues.redhat.com/browse/KOGITO-2017

- S2I build, update source - TBD
- binary build, reupload binary - TBD
- S2I build, update source while other build running - TBD
- binary build, update source while other build running - TBD

ricardozanini:
Guys, this is a workaround to fix this bug, I'll address some of your comments, but like described everything will be rewritten soon: https://issues.redhat.com/browse/KOGITO-1999

pkg/controller/kogitoapp/build/trigger.go
ricardozanini: yeah, I was using it on an other alternative, forget to remove it.

sutaakar:
Ok, will count with that.
Just need to make sure that there are no regressions during transition period, before KOGITO-1999 is fully implemented.

pkg/util/maps.go
ricardozanini: I have mixed feelings about this. Not sure if it has a better readability.

pkg/util/maps.go
ricardozanini: This code has been moved to this new file. I won't touch on this functions in this JIRA to not insert regressions since it's used by CLI, out of the scope to test it now. If you see that we can improve it, please open a JIRA and we can rethink those functions.

pkg/controller/kogitoapp/kogitoapp_controller.go
ricardozanini: If the image stream for the final runtime image is changed and we don't have `s2i` builds, we trigger this build.

I just ported the code from the old function. This will be refactored on https://issues.redhat.com/browse/KOGITO-1999

ricardozanini:
> Ok, will count with that.
> Just need to make sure that there are no regressions during transition period, before KOGITO-1999 is fully implemented.

Agreed. I was not talking about your tests, that make a lot of sense to me. :)

pkg/util/maps.go
ricardozanini: Thanks, just to make sure I retested the function with coverage and this branch has not been reached.

pkg/controller/kogitoapp/kogitoapp_controller.go
MarianMacik: I see, then please bear in mind https://issues.redhat.com/browse/KOGITO-2017 that runtime builds is missing image stream change trigger as @sutaakar pointed out.

pkg/util/maps.go
MarianMacik: Actually when I think about it right now this is the same as writing:

```go
if len(source) == 0 || len(expected) == 0 {
   return false
}
```
:smiley:

MarianMacik:
@ricardozanini I looked at the updates, looks good, just found that the if can be indeed even nicer :) Other than that looks good to me so far :wink:

ricardozanini:
@sutaakar by seeing in your tests, it looks like you haven't faced the same issues as @MarianMacik . Are you guys using the same env? Looks like the problem does not happen on CRC.

pkg/util/maps.go
ricardozanini: LOL
yeah, I over thought this.

sutaakar:
Yes, we should be using same environment (OCP 4.4, 3 master nodes)
I haven't seen that issue yet.

ricardozanini:
> Yes, we should be using same environment (OCP 4.4, 3 master nodes)
> I haven't seen that issue yet.

@MarianMacik by looking at your logs, maybe your operator version wasn't updated.

MarianMacik:
The operator was built straight from this PR. I will try to update images version as well as I use 0.9.0.

ricardozanini:
I was just wondering why @sutaakar hasn’t seen it yet. I’ll take a look after my coffee.

MarianMacik:
@ricardozanini Yes, good question. Maybe I run more tests with examples? I run Quarkus native/non-native + Spring Boot, all of that with and without persistence times 4 as I test 40k, 80k, 160k and 320k. And with 40k and 80k the issue doesn't manifest as the test finishes before the second build is finished. But with 160k and 320k the build finishes in the middle of the test and a pod is restarted and thus failing my test run. So I think that's why.

ricardozanini:
It's a race condition problem.
OCP is starting the first build by the ImageChange event:

```
Build trigger cause:	Image change
Image ID:		image-registry.openshift-image-registry.svc:5000/mmacik-perf-jenkins-cucumber-565n/kogito-quarkus-ubi8-s2i@sha256:1eb22ea8b16af416179789775e3b6c7ba31c6ceb3193acd113d2f3f4333183bf
Image Name/Kind:	kogito-quarkus-ubi8-s2i:0.9.1 / ImageStreamTag
```

Our code is not catching this build because it's created after ours probably. In the logs, I can see that we are only triggering once. Let's try to reconcile after the BC is created, then we will have time to catch all objects.

pkg/controller/kogitoapp/build/trigger.go
```diff
@@ -0,0 +1,109 @@
...
+	client        *client.Client
+	buildConfig   *buildv1.BuildConfig
+}
+
+func (t *trigger) HasBuildConfiguration() bool {
+	return t.buildConfig != nil
+}
+
+func (t *trigger) OnBuildConfig(buildConfig *buildv1.BuildConfig) Trigger {
```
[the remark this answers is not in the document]
ricardozanini: Not yet, but I've decided to let it there since we will use in the KogitoBuild CR. Just to not forget this interface.

PR #327, opened 2020-04-28
[KOGITO-1852]: Update check for internal registry

radtriste:
https://issues.redhat.com/browse/KOGITO-1852

https://github.com/kiegroup/kogito-cloud-operator/pull/304 was not enough
we need also to confirm that the image pull secret is correctly attached to the pod...

### DIFF
--- a/test/framework/kubernetes.go
+++ b/test/framework/kubernetes.go
@@ -218,3 +218,13 @@ func CreateSecret(namespace, name string, secretContent map[string]string) error
 
 	return kubernetes.ResourceC(kubeClient).Create(secret)
 }
+
+// CheckPodHasImagePullSecretWithPrefix checks that a pod has an image pull secret starting with the given prefix
+func CheckPodHasImagePullSecretWithPrefix(pod *corev1.Pod, imagePullSecretPrefix string) bool {
+	for _, secretRef := range pod.Spec.ImagePullSecrets {
+		if strings.HasPrefix(secretRef.Name, imagePullSecretPrefix) {
+			return true
+		}
+	}
+	return false
+}
--- a/test/framework/operator.go
+++ b/test/framework/operator.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
 	infra "github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/operator"
 	"github.com/kiegroup/kogito-cloud-operator/test/config"
 
 	olmapiv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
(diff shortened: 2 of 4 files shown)


ricardozanini:
Wait. Is this a WIP?

test/framework/operator.go
```diff
@@ -120,10 +122,47 @@ func IsKogitoOperatorRunning(namespace string) (bool, error) {
...
+			running, err := IsKogitoOperatorRunning(namespace)
+			if err != nil {
+				return false, err
+			}
+
+			// If not running, make sure the image pull secret is present in pod
+			// If not present, delete the pod to allow its reconstruction with correct pull secret
+			// Note that this is specific to Openshift
+			if !running {
+				podList, err := GetPodsWithLabels(namespace, map[string]string{"name": "kogito-operator"})
```
Sgitario, line 134: "kogito-operator" string is in Operator.Name.

test/framework/operator.go
```diff
@@ -120,10 +122,47 @@ func IsKogitoOperatorRunning(namespace string) (bool, error) {
...
+							return false, nil
+						}
+					}
+				}
+			}
+			return running, nil
 		})
 }
 
+func podHasImagePullSecretPrefix(pod *corev1.Pod, imagePullSecretPrefix string) bool {
```
Sgitario, line 155: This method should be part of the kubernetes script instead?

radtriste:
@ricardozanini no worry, almost done.

@Sgitario it is totally because sometimes Openshift is too slow ...
The first fix was checking that the image pull secret was created into the namespace before creating the operator. But going further in the tests, into the Kogito Openshift, we noticed that sometimes it happens that, even if the secret is created, it is not attached to the pod when operator pod is created.
The idea of the fix is really to make sure that the secret is attached. If it is not, the pod is switched into `ErrImagePull` and we get a `authentication required`, and so the scenario will fail ...

So we have 2 things now that are done:

1. We wait for the image pull secret to be created.
2. Waiting for Kogito operator, we check the secret is attached to the pod.

workaround #2 will just almost never happen. It is mostly for really low clusters...

test/framework/operator.go
radtriste: will change that

radtriste:
tested on Kogito Openshift. Took 4 hours but no ErrImagePull :)

test/framework/kubernetes.go
```diff
@@ -218,3 +218,15 @@ func CreateSecret(namespace, name string, secretContent map[string]string) error
 
 	return kubernetes.ResourceC(kubeClient).Create(secret)
 }
+
+// CheckPodHasImagePullSecretWithPrefix checks that a pod has an image pull secret starting with the given prefix
+func CheckPodHasImagePullSecretWithPrefix(pod *corev1.Pod, imagePullSecretPrefix string) bool {
+	if len(pod.Spec.ImagePullSecrets) > 0 {
```
MarianMacik, line 224: Just a note that this is unnecessary even if ImagePullSecrets is nil.

radtriste: oh right ! will remove that

spolti:
Is this PR ready?

PR #329, opened 2020-04-29
KOGITO-1959: BDD tests: add possibility to check for errors in the wa…

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1959
Description: Add a new errorCondition closure to the WaitFor utility in order to define when to stop waiting for a condition.

As golang does not support having the same method with different arguments, we provide default methods to add the error condition (at the moment, only two):
- NoErrorCondition: when we don't want any error condition
- CheckPodsByDeploymentConfigInError: check whether pods under one deployment config is in error state (at the moment, only checks whether is ErrImagePull - the image can't be found
- CheckPodsWithLabelConfigInError: same as above but using one label

### DIFF
--- a/test/framework/openshift.go
+++ b/test/framework/openshift.go
@@ -88,7 +88,7 @@ func WaitForDeploymentConfigRunning(namespace, dcName string, podNb int, timeout
 				GetLogger(namespace).Debugf("Deployment config has %d available replicas\n", dc.Status.AvailableReplicas)
 				return dc.Status.AvailableReplicas == int32(podNb), nil
 			}
-		})
+		}, CheckPodsByDeploymentConfigInError(namespace, dcName))
 }
 
 // GetDeploymentConfig retrieves a deployment config
@@ -183,9 +183,9 @@ func WaitAndRetrieveRouteURI(namespace, serviceName string) (string, error) {
 	return routeURI, nil
 }
 
-// WaitForOnOpenshift is a specific method
-func WaitForOnOpenshift(namespace, display string, timeoutInMin int, condition func() (bool, error)) error {
-	return WaitFor(namespace, display, GetOpenshiftDurationFromTimeInMin(timeoutInMin), condition)
+// WaitForOnOpenshift waits for a specification condition
+func WaitForOnOpenshift(namespace, display string, timeoutInMin int, condition func() (bool, error), errorConditions ...func() (bool, error)) error {
+	return WaitFor(namespace, display, GetOpenshiftDurationFromTimeInMin(timeoutInMin), condition, errorConditions...)
 }
 
 // GetOpenshiftDurationFromTimeInMin will calculate the time depending on the configured cluster load factor
--- a/test/framework/util.go
+++ b/test/framework/util.go
@@ -77,8 +77,8 @@ func ReadFromURI(uri string) (string, error) {
 	return string(data), nil
 }
 
-// WaitFor waits for a specific condition to be met
-func WaitFor(namespace, display string, timeout time.Duration, condition func() (bool, error)) error {
+// WaitFor waits for a specification condition to be met or until one error condition is met
+func WaitFor(namespace, display string, timeout time.Duration, condition func() (bool, error), errorConditions ...func() (bool, error)) error {
 	GetLogger(namespace).Infof("Wait %s for %s", timeout.String(), display)
 
 	timeoutChan := time.After(timeout)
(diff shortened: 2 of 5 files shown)


Sgitario:
I've tested a few features and they ran ok, but still need to verify more features.

test/framework/util.go
```diff
@@ -78,7 +78,7 @@ func ReadFromURI(uri string) (string, error) {
 }
 
 // WaitFor waits for a specific condition to be met
-func WaitFor(namespace, display string, timeout time.Duration, condition func() (bool, error)) error {
+func WaitFor(namespace, display string, timeout time.Duration, condition func() (bool, error), errorCondition func() (bool, error)) error {
```
sutaakar, line 81: Wouldn't it have more sense to create a different function which will accept the errorCondition and keep this WaitFor as it is?
That way we can avoid having to specify NoErrorCondition() everywhere.

test/framework/openshift.go
```diff
@@ -184,8 +184,8 @@ func WaitAndRetrieveRouteURI(namespace, serviceName string) (string, error) {
 }
 
 // WaitForOnOpenshift is a specific method
-func WaitForOnOpenshift(namespace, display string, timeoutInMin int, condition func() (bool, error)) error {
-	return WaitFor(namespace, display, GetOpenshiftDurationFromTimeInMin(timeoutInMin), condition)
+func WaitForOnOpenshift(namespace, display string, timeoutInMin int, condition func() (bool, error), errorCondition func() (bool, error)) error {
```
radtriste, line 187: we could create also a WaitForOnOpenshift(namespace, display, timeoutInMin, condition) method that would directly call this one with NoErrorCondition ?

radtriste: woops, @sutaakar was faster XD

test/framework/util.go
Sgitario: I wanted to avoid to have two methods: WaitForWithErrorCondition and another for WaitFor. However, will do as you requested.

test/framework/openshift.go
Sgitario: We cannot have the same methods with different arguments. I can create another method WaitForOnOpenshiftWithErrorCondition which is a bit verbose bearing in mind that providing an error condition should be a good practice.... but as you prefer ...

Sgitario:
@sutaakar and @radtriste I've updated the PR according to your comments by adding a new method with the error condition closure. However, I would rather prefer having only one method as it was previously. Having only one method empathises that when using the waitFor is a good practice to provide also an error condition. With two methods, maybe we even won't notice about this possibility.

sutaakar:
@Sgitario @radtriste I have an idea, what about using varargs for error conditions?
That way we can keep just one method and allow defining multiple error conditions.

radtriste:
sounds good to me

Sgitario:
@sutaakar good idea! I've already changed the method to use optional arguments, this way:
- we only have one method
- we can provide more than one error conditions if needed
Thanks!

test/framework/util.go
```diff
@@ -77,8 +77,8 @@ func ReadFromURI(uri string) (string, error) {
 	return string(data), nil
 }
 
-// WaitFor waits for a specific condition to be met
-func WaitFor(namespace, display string, timeout time.Duration, condition func() (bool, error)) error {
+// WaitFor waits for a specification condition to be met or until one error condition is met
+func WaitFor(namespace, display string, timeout time.Duration, condition func() (bool, error), errorConditions []func() (bool, error)) error {
```
radtriste, line 81: why don't we have a varargs as well here for `errorConditions` ?

Sgitario: This was the first thing I did, but golang does not support varargs when this is sent to different methods:

1.- errorConditions as varargs in WaitForOnOpenshift
2.- In WaitForOnOpenshift, send the errorConditions to WaitFor which is also set as varargs -> Compiler error because golang things errorConditions is an array, so it does not match with the varargs contract (I know, this with Java does not happen)

Therefore, we would need two WaitFor methods, one using an array of errorCondition and another using errorCondition as varargs. But taking into account we don't use the WaitFor directly anywhere (at least now), having the second method is useless.

radtriste: https://github.com/kiegroup/kogito-cloud-operator/blob/8f96a5c892f619967472e9021d62bfaaa2459c90/test/framework/cli.go#L52

radtriste: taking varargs and giving varargs :)

Sgitario: haha I see dots, lot of dots.
Thanks Tristan, will update this and I just learned something new! :)

Sgitario: Changed.

Sgitario:
I ran a few scenarios and everything looks working fine.

radtriste:
@MarianMacik can you please have a look ?

@Sgitario Can you rebase ?

Sgitario:
Rebased and conflicts resolved.

PR #332, opened 2020-05-04
KOGITO-953: Cucumber tests: Test OLM subscription

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-953
Description: Simple test for checking that OLM subscription is working on alpha and dev-preview channels. 
This test will be executed as a post-release step. So, it's disabled now and marked with the tag @release.

### DIFF
--- a/test/features/olm/subscription_operator.feature
+++ b/test/features/olm/subscription_operator.feature
@@ -1,0 +1,16 @@
+@olm
+@release
+Feature: Deploy Kogito Resources from subscription
+
+  Background:
+    Given Namespace is created
+
+  Scenario Outline: Deploy Kogito Operator from subscription
+    When Kogito Operator is deployed from subscription using channel "<channel>"
+    Then Kogito Operator should be installed from subscription with dependencies
+
+  Examples:
+      | channel      |
+      | alpha        |
+      | dev-preview  |
+
--- a/test/main_test.go
+++ b/test/main_test.go
@@ -33,11 +33,21 @@ import (
 )
 
 const (
-	disabledTag    = "@disabled"
-	cliTag         = "@cli"
-	smokeTag       = "@smoke"
-	performanceTag = "@performance"
+	disabledTag = "@disabled"
+	cliTag      = "@cli"
 )
+
+var customTags = []customTag{
+	{Name: "@performance", IsEnabled: config.IsPerformanceTests, IgnoreIfDisabled: true},
+	{Name: "@release", IsEnabled: config.IsReleaseTests, IgnoreIfDisabled: true},
+	{Name: "@smoke", IsEnabled: config.IsSmokeTests, IgnoreIfDisabled: false},
+}
+
+type customTag struct {
+	Name             string
+	IsEnabled        func() bool
+	IgnoreIfDisabled bool
+}
 
 var opt = godog.Options{
 	Output:    colors.Colored(os.Stdout),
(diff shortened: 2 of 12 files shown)


test/features/olm/subscription_operator.feature
```diff
@@ -0,0 +1,17 @@
...
+@olm
+@release
+Feature: Deploy Kogito Resources from subscription
+
+  Background:
+    Given Namespace is created
+
+  Scenario Outline: Deploy Kogito Operator from subscription
+    When Kogito Operator is deployed from subscription using channel "<channel>"
+    Then Kogito Operator should be installed from subscription with dependencies
```
Sgitario, line 11: This step will check that the installed version of the operator matches with the CSV version, otherwise it fails.

Sgitario:
I've ran the tests and they work fine. Should we mark them with @smoke as well? Or with @release is enough?

sutaakar:
Isn't such test executed on OpenShift side when they accept the operator PR?

test/features/olm/subscription_operator.feature
radtriste: which CSV version is that ?

Sgitario: The CSV version is checked from the subscription depending on the channel. It's using this method: https://github.com/kiegroup/kogito-cloud-operator/blob/c873d19970784825e8547f1367ee2649dfb2ab6f/pkg/infrastructure/operator.go#L61

Sgitario:
> Isn't such test executed on OpenShift side when they accept the operator PR?

If this is something that OpenShift does when they accept the operator, then we should not do this. What do you think @radtriste ?

test/features/olm/subscription_operator.feature
```diff
@@ -0,0 +1,17 @@
+@disabled
```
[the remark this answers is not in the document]
Sgitario: I thought the tags were automatically injected. I've made some changes in regards of this, so we can define the custom tags (smoke, performance and now release).
With this change, we can now compose tags by for example, setting @smoke and another tag @performance altogether.

test/main_test.go
```diff
@@ -33,12 +33,22 @@ import (
...
+var customTags = []customTag{
+	{Name: "@performance", IsEnabled: config.IsPerformanceTests, IgnoreIfDisabled: true},
+	{Name: "@release", IsEnabled: config.IsReleaseTests, IgnoreIfDisabled: true},
+	{Name: "@smoke", IsEnabled: config.IsSmokeTests, IgnoreIfDisabled: false},
+}
+
+type customTag struct {
+	Name             string
+	IsEnabled        func() bool
+	IgnoreIfDisabled bool
```
Sgitario, line 49: This might be tricky as it's only used by the smoke tag because it's not excluded if it's not set. If smoke tag is set, we only run the smoke tests, if not set, we run all the tests (including smokes). 
If this is too much, I can restore it as it was before, but I like the idea of having all the custom tag configurations in the same place to see/change behaviour when/if needed.

sutaakar:
@spolti Do you know what checks are done when Kogito operator is added to the catalog (scorecard checks?)? Does it check whether the operator is successfully installed using OLM subscription?

radtriste:
AFAIK the scorecard is running checks before the csv is added to the OLM.
@spolti will add more details

test/main_test.go
Sgitario: I think both are tricky and a bit verbose. More ideas? IgnoreByDefault?

spolti:
@sutaakar @radtriste yeah, we run scorecard before push the new CSV.
When a PR is created on operator-framework/community-operators a few other tests are made, you can check all here: https://github.com/operator-framework/community-operators/pull/1635

Sgitario:
> @sutaakar @radtriste yeah, we run scorecard before push the new CSV.
> When a PR is created on operator-framework/community-operators a few other tests are made, you can check all here: [operator-framework/community-operators#1635](https://github.com/operator-framework/community-operators/pull/1635)

If this is done, should we reject the JIRA and close this PR without merging?

radtriste:
I tried to remember why I opened this issue but cannot ... I think I just wanted to make sure we have the correct version set into the OperatorHub.

With a release automation pipeline, we should not need for such a check anyway I think.
Let's close it for now and see later if needed ...

Sgitario:
No problem. Will close the PR without merging and reject the ticket.

PR #333, opened 2020-05-04
KOGITO-1700 - Add optional configuration of Infinispan in BDD

MarianMacik:
https://issues.redhat.com/browse/KOGITO-1700

Things to discuss:

1. I found that for Infinispan env vars we use 2 maps (Quarkus/Spring Boot) in the operator framework:
https://github.com/kiegroup/kogito-cloud-operator/blob/d967c4f6331ab79d4ee71906058dd1bad2035bce/pkg/controller/kogitoapp/resource/deployment_config.go#L69-L82
But they are not public, so I cannot use them and have to use my own, see the `addInfinispanEnvVars` method. Basically I have to use if now. Can we make them public so they can be used in the test framework, wdyt?

2. I created the `KogitoAppHolder` to follow the pattern of `KogitoServiceHolder` in the framework. For KogitoApp it is not needed on the framework side right now (not used in the framework at all, just in the steps) but I wanted to be consistent with KogitoService.

3. First time, I tried to use the secrets but then found out that you cannot set an environment variable value from secret using Kogito CLI - maybe a proposal for a new feature? You can set just an environment variable name and its value. So right now I have fallen back to using plaintext passwords and it works :+1: 

4. Commented out steps like `Then Service "process-quarkus-example" contains <requests> instances of process` because of currently not having a viable solution to properly fetch a large number of processes, i.e. using pagination. The list of features which belong to these commented steps is provided at the beginning of the performance feature file.

5. Of course, KogitoApp can now be configured for external Infinispan using:
```
| infinispan      | username | developer                 |
| infinispan      | password | mypass                    |
| infinispan      | uri      | external-infinispan:11222 |
```

Many thanks for submiting your Pull Request :heart: :racing_car:! 

### DIFF
--- a/pkg/apis/app/v1alpha1/kogitoservices_types.go
+++ b/pkg/apis/app/v1alpha1/kogitoservices_types.go
@@ -169,6 +169,23 @@ func (k *KogitoServiceSpec) AddEnvironmentVariable(name, value string) {
 	return
 }
 
+// AddEnvironmentVariableFromSecret adds a new environment variable from the secret under the key
+func (k *KogitoServiceSpec) AddEnvironmentVariableFromSecret(variableName, secretName, secretKey string) {
+	env := corev1.EnvVar{
+		Name: variableName,
+		ValueFrom: &corev1.EnvVarSource{
+			SecretKeyRef: &corev1.SecretKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: secretName,
+				},
+				Key: secretKey,
+			},
+		},
+	}
+	k.Envs = append(k.Envs, env)
+	return
+}
+
 // AddResourceRequest adds new resource request. Works also on uninitialized Requests field.
 func (k *KogitoServiceSpec) AddResourceRequest(name, value string) {
 	if k.Resources.Requests == nil {
--- a/test/features/kogito_service_performance.feature
+++ b/test/features/kogito_service_performance.feature
@@ -1,7 +1,6 @@
 # Commented code will be addressed by further enhancements:
-# https://issues.redhat.com/browse/KOGITO-1699
-# https://issues.redhat.com/browse/KOGITO-1700
 # https://issues.redhat.com/browse/KOGITO-1701
+# https://issues.redhat.com/browse/KOGITO-1888
 
 @performance
 Feature: Kogito Service Performance
@@ -10,11 +9,11 @@ Feature: Kogito Service Performance
     Given Namespace is created
 
   @quarkus
-  Scenario Outline: Quarkus Kogito Service Performance with requests <requests>, native <native> but without persistence
+  Scenario Outline: Quarkus Kogito Service Performance with native <native>, without persistence and with requests <requests>
     Given Kogito Operator is deployed
     And Deploy quarkus example service "process-quarkus-example" with configuration:
       | config      | native       | <native> |
-      | runtime-env | JAVA_OPTIONS | -Xmx8G   |
+      | runtime-env | JAVA_OPTIONS | -Xmx10G  |
     And Kogito application "process-quarkus-example" has 1 pods running within <minutes> minutes
     And Service "process-quarkus-example" with process name "orders" is available within 3 minutes
 
@@ -53,12 +52,17 @@ Feature: Kogito Service Performance
 
   @quarkus
   @persistence
-  Scenario Outline: Quarkus Kogito Service Performance with requests <requests>, native <native> and persistence
+  Scenario Outline: Quarkus Kogito Service Performance with native <native>, with persistence and with requests <requests>
     Given Kogito Operator is deployed with Infinispan operator
+    And Infinispan instance "external-infinispan" is deployed for performance within 5 minute(s) with configuration:
+      | username | developer |
+      | password | mypass    |
     And Deploy quarkus example service "process-quarkus-example" with configuration:
-      | config      | native       | <native> |
-      | config      | persistence  | enabled  |
-      | runtime-env | JAVA_OPTIONS | -Xmx8G   |
+      | config      | native       | <native>                  |
+      | runtime-env | JAVA_OPTIONS | -Xmx10G                   |
+      | infinispan  | username     | developer                 |
+      | infinispan  | password     | mypass                    |
+      | infinispan  | uri          | external-infinispan:11222 |
     And Kogito application "process-quarkus-example" has 1 pods running within <minutes> minutes
     And Service "process-quarkus-example" with process name "orders" is available within 3 minutes
 
@@ -73,8 +77,8 @@ Feature: Kogito Service Performance
       }
       """
 
-    Then Service "process-quarkus-example" contains <requests> instances of process with name "orders" within 1 minutes
-    And Service "process-quarkus-example" contains <requests> instances of process with name "orderItems" within 1 minutes
+    #Then Service "process-quarkus-example" contains <requests> instances of process with name "orders" within 1 minutes
+    #And Service "process-quarkus-example" contains <requests> instances of process with name "orderItems" within 1 minutes
     #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
 
 
@@ -96,10 +100,10 @@ Feature: Kogito Service Performance
 #####
 
   @springboot
-  Scenario Outline: Spring Boot Kogito Service Performance with requests <requests> but without persistence
+  Scenario Outline: Spring Boot Kogito Service Performance without persistence and with requests <requests>
     Given Kogito Operator is deployed
     And Deploy springboot example service "process-springboot-example" with configuration:
-      | runtime-env | JAVA_OPTIONS | -Xmx8G |
+      | runtime-env | JAVA_OPTIONS | -Xmx10G |
     And Kogito application "process-springboot-example" has 1 pods running within <minutes> minutes
     And Service "process-springboot-example" with process name "orders" is available within 3 minutes
 
@@ -129,11 +133,16 @@ Feature: Kogito Service Performance
 
   @springboot
   @persistence
-  Scenario Outline: Spring Boot Kogito Service Performance with requests <requests> and persistence
+  Scenario Outline: Spring Boot Kogito Service Performance with persistence and with requests <requests>
     Given Kogito Operator is deployed with Infinispan operator
+    And Infinispan instance "external-infinispan" is deployed for performance within 5 minute(s) with configuration:
+      | username | developer |
+      | password | mypass    |
     And Deploy springboot example service "process-springboot-example" with configuration:
-      | config      | persistence  | enabled |
-      | runtime-env | JAVA_OPTIONS | -Xmx8G  |
+      | runtime-env | JAVA_OPTIONS | -Xmx10G                   |
+      | infinispan  | username     | developer                 |
+      | infinispan  | password     | mypass                    |
+      | infinispan  | uri          | external-infinispan:11222 |
     And Kogito application "process-springboot-example" has 1 pods running within <minutes> minutes
     And Service "process-springboot-example" with process name "orders" is available within 3 minutes
 
@@ -148,8 +157,8 @@ Feature: Kogito Service Performance
       }
       """
 
-    Then Service "process-springboot-example" contains <requests> instances of process with name "orders" within 1 minutes
-    And Service "process-springboot-example" contains <requests> instances of process with name "orderItems" within 1 minutes
+    #Then Service "process-springboot-example" contains <requests> instances of process with name "orders" within 1 minutes
+    #And Service "process-springboot-example" contains <requests> instances of process with name "orderItems" within 1 minutes
     #And All human tasks on path "orderItems" with path task name "Verify_order" are successfully "completed" with timing "true"
 
     Examples:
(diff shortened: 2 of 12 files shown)


sutaakar:
> First time, I tried to use the secrets but then found out that you cannot set an environment variable using Kogito CLI - maybe a proposal for a new feature? So right now I have fallen back to using plaintext passwords and it works +1

What environment variable do you mean? Kogito CLI has a possibility to define environment variables

`-e, --env stringArray           Key/Pair value environment variables that will be set to the service runtime. For example 'MY_VAR=my_value'. Can be set more than once.
`

Or do you mean defining content of a secret?

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -169,6 +169,23 @@ func (k *KogitoServiceSpec) AddEnvironmentVariable(name, value string) {
 	return
 }
 
+// AddEnvironmentVariableFromSecret adds a new environment variable from the secret under the key
+func (k *KogitoServiceSpec) AddEnvironmentVariableFromSecret(name, secret, key string) {
```
sutaakar, line 173: Meaning of those parameters is not clear on the first look. I would suggest to adjust them like this:
```suggestion
func (k *KogitoServiceSpec) AddEnvironmentVariableFromSecret(name, secretName, secretKey string) {
```

MarianMacik: I see a difference between secret and secretName but key and secretKey makes no difference for me in this context. But sure, will change it :)

MarianMacik:
> > First time, I tried to use the secrets but then found out that you cannot set an environment variable using Kogito CLI - maybe a proposal for a new feature? So right now I have fallen back to using plaintext passwords and it works +1
> 
> What environment variable do you mean? Kogito CLI has a possibility to define environment variables
> 
> `-e, --env stringArray Key/Pair value environment variables that will be set to the service runtime. For example 'MY_VAR=my_value'. Can be set more than once. `
> 
> Or do you mean defining content of a secret?
> 
> Edit - I guess you mean using `ValueFrom` pointing env variable value to some other object, right?

Ah, sorry, forgot to write it exactly (and I was reading this after me). Sure, env variables work - as I currently made it possible that way.

pkg/apis/app/v1alpha1/kogitoservices_types.go
sutaakar: Well, when I read the `key` for the first time I didn't associated it with key in secret, but was thinking about some key in environment variable.

radtriste:
for point #1, I think we could move that to the `infrastructure` package

pkg/apis/app/v1alpha1/kogitoservices_types.go
MarianMacik: I see now. I have also changed `name` to `variableName` because when I looked at it after some time, it was not enough. Thanks for the input!

MarianMacik:
@radtriste You mean to the `infinispan.go` file?
https://github.com/kiegroup/kogito-cloud-operator/blob/c873d19970784825e8547f1367ee2649dfb2ab6f/pkg/infrastructure/infinispan.go#L1

And after that I would probably make these variables and constants public, wdyt?

radtriste:
Those are kogitoapp related.
So I would move them to `kogito_app.go`

test/steps/infinispan.go
```diff
@@ -22,43 +22,60 @@ import (
...
 
+/*
+	DataTable for Infinispan example:
+	| username | developer |
+	| password | mypass    |
+ */
+
 const (
-	//DataTable first column
-	infinispanUsernameKey     = "username"
```
sutaakar, line 30: @MarianMacik @radtriste I think that the situation with DataTable column names deserves deeper discussion.
The problem with sharing of those constants is that we create an artificial dependency between go files. It also reduces visibility of the columns - how to find out what are the possible values for column?

One option would be to add an example for each DataTable, showing every possible configuration which can be applied - as done few lines above. The possible downside could be that example and actual functionality can get out of sync.
Another option is to list all the column names as constants with specific unique identifier (as constants in  same package are shared) as shown in these removed lines.

I would personally prefer combination of both methods, but I am open to another options.

What do you think?

test/framework/kubernetes.go
```diff
@@ -139,6 +139,16 @@ func IsPodRunning(pod *corev1.Pod) bool {
 	return pod.Status.Phase == corev1.PodRunning
 }
 
+// IsPodStatusConditionReady returns true if all pod's containers are ready (really running)
+func IsPodStatusConditionReady(pod *corev1.Pod) bool{
```
sutaakar, line 143: Makes me think - can we replace method `IsPodRunning` with this one? They seem to have same purpose.

test/steps/infinispan.go
MarianMacik: So you suggest to export these constants to a separate file? I thought that `dataindex.go` will serve this purpose. In fact it now does what you meant by

> Another option is to list all the column names as constants with specific unique identifier (as constants in same package are shared) as shown in these removed lines.

I will just add KogitoApp DataTable example to the `utils.go` method as this mostly contains parsing functions for KogitoApp.

Or we can merge everything about parsing DataTable to one file and move also examples of DataTables there.

test/framework/kubernetes.go
MarianMacik: Can be, this one is more "thorough" I would say as it really returns true only once the pod as a whole (all internal containers) are running. If you check only a Pod Running status, it may be running, but doesn't have to be ready.

test/steps/infinispan.go
radtriste: We could have one datatable handler file like @MarianMacik suggested.
And I like having specific unique identifier which should be self documentation.
Adding a datatable example is also a good thing but we need to keep it up to date :)

MarianMacik: OK, so we agree on leaving the constants in one file?

Then I will rename utils to utils_datatable or just datatable for example and will merge everything about Data Table in this one file.

About example tables, I can probably leave there a comment with each constant which will describe to which resources the constant applies (KogitoApp, DataIndex, Infinispan etc.) as the name of columns overlap. And having 3 different constants just for the same name ("infinispan") is a little bit overkill, don't you think? I am not sure about this last point.

radtriste: > And having 3 different constants just for the same name ("infinispan") is a little bit overkill

What do you mean ?

sutaakar: My idea is to handle the DataTable in the same steps file where it belongs, with all the constants to be there too.

For example the test/steps/infinispan.go 
The file contains definitions of steps. So I would keep DataTable handling there too, as from my perspective it is part of steps definition. Also it would contain all the constants defining DataTable columns, with proper identifier (as the columns are specific for Infinispan).

So, the test/steps/infinispan.go would contain:
- constants (infinispanUsernameKey, infinispanPasswordKey.... ,prefixed with infinispan to mark that these constants belong to Infinispan)
- DataTable example showing all possible columns and their usage
- DataTable parsing function for Infinispan step (as it is integral part of step definition)

Then, for example test/steps/kogitodataindex.go would contain:
- constants (dataIndexInfinispanUsernameKey, dataIndexInfinispanPasswordKey.... ,prefixed with dataIndex to mark that these constants belong to DataIndex)
- DataTable example showing all possible columns and their usage
- DataTable parsing function for DataIndex step

Moving all DataTable parsing functions to one file is an option too. I am just worry that it will become fragile, less readable and can have tendencies to become out of sync as DataTable will be on different place than step definition.

MarianMacik: > > And having 3 different constants just for the same name ("infinispan") is a little bit overkill
> 
> What do you mean ?

I meant what Karel is suggesting :) So for "user" we will have in the end 3 different constants as "user" is used in the first column for data index, kogito app and infinispan data tables. Just wanted to be sure I understood correctly as there will be duplication.

test/framework/kubernetes.go
MarianMacik: Switched `CheckPodsAreRunning()` to `CheckPodsAreReady()` and now it calls `IsPodStatusConditionReady()` function. I left `IsPodRunning()` still in the code as it is used for checking if the pod is being monitored via logs. See here: https://github.com/kiegroup/kogito-cloud-operator/blob/2fc81b0c1c1a7888b7642c88f98174c04f781f34/test/framework/logging.go#L142

In this situation, it is enough for a pod to be `Running` as by then all containers should be present I guess and we can get their logs with follow. Wdyt? Waiting for all containers to be `Ready` might not be the best option as it can start producing output even before the start command finished (which only then marks the container as  Ready I guess)?

So let me know what you think about it.

sutaakar: Looks good to me, thanks.

MarianMacik:
Reviews addressed. `utils.go` now contains only functions to retrieve columns of a table and one parsing function which is used in the `kogitoapp.go` and the `openshift.go`.

Also infinispan secret, Kogito external secret and its keys are in `infinispan.go` right now and shared as that is something I believe we can share, this won't change as there is only one Infinispan and credentials should be shared among apps so only one kogito external secret is needed.

test/steps/infinispan.go
```diff
@@ -22,43 +22,64 @@ import (
...
+
 	if err := framework.DeployInfinispanInstance(data.Namespace, infinispan); err != nil {
 		return err
 	}
 
-	return framework.WaitForPodsWithLabel(data.Namespace, "app", "infinispan-pod", 1, 3)
+	return framework.WaitForInfinispanPodsToBeRunningWithConfig(data.Namespace, performanceInfinispanContainerSpec, 1, timeOutInMin)
 }
 
 func (data *Data) createInfinispanSecret(name string, table *messages.PickleStepArgument_PickleTable) error {
```
radtriste, line 85: Do we really need to associate this method to `Data` ?
We are just using the namespace here

sutaakar: Good point, that is a leftover from my PR.
Initially I have been considering to have this function as a separate step but then I changed my mind and forgot to remove data. Can be removed (possibly as separate PR).

sutaakar:
@spolti Can you please take a look as Zanini is away?

test/features/kogito_service_performance.feature
```diff
@@ -73,8 +77,8 @@ Feature: Kogito Service Performance
       }
       """
 
-    Then Service "process-quarkus-example" contains <requests> instances of process with name "orders" within 1 minutes
-    And Service "process-quarkus-example" contains <requests> instances of process with name "orderItems" within 1 minutes
+    #Then Service "process-quarkus-example" contains <requests> instances of process with name "orders" within 1 minutes
```
spolti, line 80: why is it commented out?

test/features/kogito_service_performance.feature
```diff
@@ -148,8 +157,8 @@ Feature: Kogito Service Performance
       }
       """
 
-    Then Service "process-springboot-example" contains <requests> instances of process with name "orders" within 1 minutes
-    And Service "process-springboot-example" contains <requests> instances of process with name "orderItems" within 1 minutes
+    #Then Service "process-springboot-example" contains <requests> instances of process with name "orders" within 1 minutes
```
spolti, line 160: same here

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -169,6 +169,23 @@ func (k *KogitoServiceSpec) AddEnvironmentVariable(name, value string) {
 	return
 }
 
+// AddEnvironmentVariableFromSecret adds a new environment variable from the secret under the key
+func (k *KogitoServiceSpec) AddEnvironmentVariableFromSecret(variableName, secretName, secretKey string) {
```
spolti, line 173: maybe add a unit test for it?
plus, isn't a similar func available somewhere else?

spolti:
plus, ci is failing, can you please verify?

test/steps/infinispan.go
MarianMacik: I originally thought the same as Karel, that we want to leave it as a separate step should anybody need it in future :) Resolved now.

pkg/apis/app/v1alpha1/kogitoservices_types.go
MarianMacik: There is a similar function but for a different purpose (setting variable directly on the container):
https://github.com/kiegroup/kogito-cloud-operator/blob/d6ed53bbeccf739a14efa51dfb658d4b69eb72db/pkg/framework/env_var.go#L110-L129

Also a test would be maybe helpful, although right now there isn't any test class for it so maybe we can address it as a separate JIRA if it is really needed? But to be honest, this function is really short and trivial so probably I wouldn't waste time with writing a test for it right now.

test/features/kogito_service_performance.feature
MarianMacik: Please see the point no. 4 in the PR description.

test/features/kogito_service_performance.feature
MarianMacik: See the point no. 4 in the PR description.

MarianMacik:
> plus, ci is failing, can you please verify?

@spolti Unfortunately I am not able to log in to the Jenkins running on Openshift. I tried it using github but I am still returned back to the "Log in with OpenShift" page.

MarianMacik:
@spolti Are we ok with merging?

spolti:
> @spolti Unfortunately I am not able to log in to the Jenkins running on Openshift. I tried it using github but I am still returned back to the "Log in with OpenShift" page.

ask @msmagnanijr for ths instructions.

PR #335, opened 2020-05-05
[KOGITO-2010] Review scenario tags and add a tags dictionary

sutaakar:

https://issues.redhat.com/browse/KOGITO-2010

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Table of Contents
             * [Running BDD tests with custom Kogito Build images' version](#running-bdd-tests-with-custom-kogito-build-images-version)
             * [Running smoke tests](#running-smoke-tests)
             * [Running performance tests](#running-performance-tests)
+            * [List of test tags](#list-of-test-tags)
          * [Running the Kogito Operator locally](#running-the-kogito-operator-locally)
       * [Contributing to the Kogito Operator](#contributing-to-the-kogito-operator)
 
@@ -1083,6 +1084,8 @@ You can set those optional keys:
     - "~@wip": exclude all scenarios with wip tag
     - "@wip && ~@new": run wip scenarios, but exclude new
     - "@wip,@undone": run wip or undone scenarios
+
+  Complete list of supported tags and descriptions can be found in [List of test tags](#list-of-test-tags)
 - `concurrent` is the number of concurrent tests to be ran.  
   *Default is 1.*
 - `timeout` sets the timeout in minutes for the overall run.  
@@ -1187,6 +1190,39 @@ It will run only tests tagged with `@performance`.
 All options from BDD tests do also apply here.
 
 **NOTE:** Performance tests should be run without concurrency.
+
+#### List of test tags
+
+| Tag name           | Tag meaning                                                                        |
+| ------------------ | ---------------------------------------------------------------------------------- |
+| @smoke             | Smoke tests verifying basic functionality                                          |
+| @performance       | Performance tests                                                                  |
+| @olm               | OLM integration tests                                                              |
+| @travelagency      | Travel agency tests                                                                |
+|                    |                                                                                    |
+| @disabled          | Disabled tests, usually with comment describing reasons                            |
+| @cli               | Tests to be executed only using Kogito CLI                                         |
+|                    |                                                                                    |
+| @springboot        | SpringBoot tests                                                                   |
+| @quarkus           | Quarkus tests                                                                      |
+|                    |                                                                                    |
+| @dataindex         | Tests including DataIndex                                                          |
+| @jobsservice       | Tests including Jobs service                                                       |
+| @managementconsole | Tests including Management console                                                 |
+| @infra             | Tests checking KogitoInfra functionality                                           |
+|                    |                                                                                    |
+| @binary            | Tests using Kogito applications built locally and uploaded to OCP as a binary file |
+| @native            | Tests using native build                                                           |
+| @persistence       | Tests verifying persistence capabilities                                           |
+| @events            | Tests verifying eventing capabilities                                              |
+| @discovery         | Tests checking service discovery functionality                                     |
+| @usertasks         | Tests interacting with user tasks to check authentication/authorization            |
+|                    |                                                                                    |
+| @resources         | Tests checking resource requests and limits                                        |
+|                    |                                                                                    |
+| @externalcomponent | Tests checking externally deployed component                                       |
+| @infinispan        | Tests checking external infinispan component                                       |
+| @kafka             | Tests checking external kafka component                                            |
 
 ### Running the Kogito Operator locally
 
--- a/test/features/deploy_quarkus_service.feature
+++ b/test/features/deploy_quarkus_service.feature
@@ -121,7 +121,6 @@ Feature: Deploy quarkus service
 #####
 
   @events
-  @persistence
   Scenario Outline: Data Index retrieves Quarkus process' events with native <native>
     Given Kogito Operator is deployed with Infinispan and Kafka operators
     And Install Kogito Data Index with 1 replicas
--- a/test/features/deploy_service_resources.feature
+++ b/test/features/deploy_service_resources.feature
@@ -6,8 +6,6 @@ Feature: Deploy the service by configuring the resource requests and limits
     Given Namespace is created
     And Kogito Operator is deployed
 
-  @runtimerequests 
-  @runtimelimits
   Scenario Outline: Setting runtime resource resource requests cpu <runtime-cpu-request>, mem <runtime-memory-request> and limits cpu <runtime-cpu-limit>, mem <runtime-memory-limit >
     Given Clone Kogito examples into local directory
     And Local example service "ruleunit-quarkus-example" is built by Maven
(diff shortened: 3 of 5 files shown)


sutaakar:
Please check what tags we need and what not :) I have added there some comments (in brackets) for tags which need to be clarified.

sutaakar:
I tried to insert empty rows separating groups of tags which could logically be together, though it is not much visible.

msmagnanijr:
rebuild

README.md
```diff
@@ -1188,6 +1191,43 @@ All options from BDD tests do also apply here.
...
+|                    |                                                                                    |
+| @binary            | Tests using Kogito applications built locally and uploaded to OCP as a binary file |
+| @native            | Tests using native build                                                           |
+| @persistence       | Tests verifying persistence capabilities                                           | (or tests running with enabled persistence?)
+| @events            | Tests verifying eventing capabilities                                              | (or tests running with enabled events?)
+| @discovery         | Tests checking service discovery functionality                                     |
+|                    |                                                                                    |
+| @resources         | Tests checking resource requests and limits                                        |
+| @runtimerequests   | Tests checking resource runtime requests                                           | (are these tags needed? what about using just @resources?)
```
Sgitario, line 1222: Agree about this. Just @resources tag should be enought.

README.md
```diff
@@ -1188,6 +1191,43 @@ All options from BDD tests do also apply here.
...
+| @disabled          | Disabled tests, usually with comment describing reasons                            |
+| @cli               | Tests to be executed only using cli                                                |
+|                    |                                                                                    |
+| @springboot        | SpringBoot tests                                                                   |
+| @quarkus           | Quarkus tests                                                                      |
+|                    |                                                                                    |
+| @dataindex         | Tests including DataIndex                                                          |
+| @jobsservice       | Tests including Jobs service                                                       |
+| @managementconsole | Tests including Management console                                                 |
+| @infra             | Tests checking KogitoInfra functionality                                           | (should mark also every test which use Kogito infra?)
```
Sgitario, line 1212: Maybe @infra tag is too much vague? If a test is using Infinispan, we can use @infinispan. If it's also using @externalcomponent, I know that it's not being deployed using the Kogito infra.

README.md
```diff
@@ -1188,6 +1191,43 @@ All options from BDD tests do also apply here.
...
+
+| Tag name           | Tag meaning                                                                        |
+| ------------------ | ---------------------------------------------------------------------------------- |
+| @smoke             | Smoke tests verifying basic functionality                                          |
+| @performance       | Performance tests                                                                  |
+| @olm               | OLM integration tests                                                              |
+| @travelagency      | Travel agency tests                                                                |
+|                    |                                                                                    |
+| @disabled          | Disabled tests, usually with comment describing reasons                            |
+| @cli               | Tests to be executed only using cli                                                |
```
Sgitario, line 1204: @cli is the Kogito CLI? The another way is to use directly the OC command, right?

README.md
```diff
@@ -1188,6 +1191,43 @@ All options from BDD tests do also apply here.
...
 **NOTE:** Performance tests should be run without concurrency.
 
+#### List of test tags
+
+| Tag name           | Tag meaning                                                                        |
+| ------------------ | ---------------------------------------------------------------------------------- |
+| @smoke             | Smoke tests verifying basic functionality                                          |
+| @performance       | Performance tests                                                                  |
+| @olm               | OLM integration tests                                                              |
+| @travelagency      | Travel agency tests                                                                |
```
Sgitario, line 1201: Should not this tag be a feature rather than a tag?

README.md
```diff
@@ -1188,6 +1191,43 @@ All options from BDD tests do also apply here.
...
+| @managementconsole | Tests including Management console                                                 |
+| @infra             | Tests checking KogitoInfra functionality                                           | (should mark also every test which use Kogito infra?)
+| @discovery         | Tests checking service discovery functionality                                     |
+|                    |                                                                                    |
+| @binary            | Tests using Kogito applications built locally and uploaded to OCP as a binary file |
+| @native            | Tests using native build                                                           |
+| @persistence       | Tests verifying persistence capabilities                                           | (or tests running with enabled persistence?)
+| @events            | Tests verifying eventing capabilities                                              | (or tests running with enabled events?)
+| @discovery         | Tests checking service discovery functionality                                     |
```
jakubschwan, line 1219: discovery tag is mentioned there twice - on line [1213](https://github.com/kiegroup/kogito-cloud-operator/pull/335/files#diff-04c6e90faac2675aa89e2176d2eec7d8R1213) and here on line 1219

README.md
sutaakar: It marks the scenarios which require Kogito CLI (for example they test that Kogito CLI installs the Kogito operator when new project is created).

README.md
sutaakar: Well, it doesn't have to be so in future.
Right now when Kogito infra is used then we know that Infinispan/Kafka will be installed. But in the future some other providers can be used.

README.md
sutaakar: The travel agency tests are located in separate feature file.
Question is whether we need a specific tag for it?

README.md
sutaakar: Done, tags removed from feature file.

README.md
radtriste: Travel agency like onboarding (annotation @discovery) are the complex examples that we have.
I would like to be able to launch them separately.
If you think of another name for the tag, then we can change it :)

README.md
radtriste: we should have one tag for each component and one that could group all the dependent components (kafka, infinispan)
`@infra` was put at the beginning but we could change it to something like `dependency` ?

On those `@infra` we are testing that the installation of the dependencies is happening well.
So I would not annotate others that are using those components as it is not exactly what we are testing IMO ...

README.md
```diff
@@ -1188,6 +1191,39 @@ All options from BDD tests do also apply here.
...
+| @quarkus           | Quarkus tests                                                                      |
+|                    |                                                                                    |
+| @dataindex         | Tests including DataIndex                                                          |
+| @jobsservice       | Tests including Jobs service                                                       |
+| @managementconsole | Tests including Management console                                                 |
+| @infra             | Tests checking KogitoInfra functionality                                           | (should mark also every test which use Kogito infra?)
+|                    |                                                                                    |
+| @binary            | Tests using Kogito applications built locally and uploaded to OCP as a binary file |
+| @native            | Tests using native build                                                           |
+| @persistence       | Tests verifying persistence capabilities                                           | (or tests running with enabled persistence?)
```
radtriste, line 1216: IMHO we should annotate the tests with what we want to test inside, not if we are using the functionality.
so for `@persistence`, those are indeed `Tests verifying persistence capabilities` and not test with persistence enabled.
or ?

README.md
```diff
@@ -1188,6 +1191,39 @@ All options from BDD tests do also apply here.
...
+| @events            | Tests verifying eventing capabilities                                              | (or tests running with enabled events?)
```
radtriste, line 1217: same things for `@events`

README.md
sutaakar: Yes, sounds good to me.

README.md
sutaakar: Well, there is a slight difference as `discovery` test a Kogito capability of discovering other services using tags. Travel agency doesn't test any specific capability, it is "just" a complex scenario.

Though it may have sense to keep the tag there, for the case of checking Travel agency functionality.

README.md
sutaakar: > @infra was put at the beginning but we could change it to something like dependency ?

I would keep the `@infra` there as it is aligned to KogitoInfra object name.

> On those @infra we are testing that the installation of the dependencies is happening well.

ok

README.md
sutaakar: Done.

sutaakar:
@radtriste Removed the `@persistence` tags in scenarios where the persistence is not tested specifically.

README.md
```diff
@@ -1188,6 +1191,39 @@ All options from BDD tests do also apply here.
...
+|                    |                                                                                    |
+| @dataindex         | Tests including DataIndex                                                          |
+| @jobsservice       | Tests including Jobs service                                                       |
+| @managementconsole | Tests including Management console                                                 |
+| @infra             | Tests checking KogitoInfra functionality                                           |
+|                    |                                                                                    |
+| @binary            | Tests using Kogito applications built locally and uploaded to OCP as a binary file |
+| @native            | Tests using native build                                                           |
+| @persistence       | Tests verifying persistence capabilities                                           |
+| @events            | Tests verifying eventing capabilities                                              |
```
radtriste, line 1217: Just thinking about `@events`, there are of 2 types or ?

- Send process instances to Kafka (and then retrieved by Data Index)
- Communication between services (like travelagency) via Kafka

We have not annotated the travelagency with any for now but it is using events.

Should we differentiate the two different ? Would we want that ?

MarianMacik: For me events are "our" events, meaning these implemented directly by Kogito. If events are implemented as a 3rd party feature for communication between KogitoApps by a developer, I wouldn't use it for these scenario. (Which is what Travel Agency implements if I understand)

radtriste: well, events are set in the BPMN process.
we still need to be sure that the communication is working good or ?

sutaakar: hmm, good question.
From my point of view it depends on the purpose of Kafka configuration made by Kogito operator. Is this configuration meant to be used by eventing capability of processes? Or is it meant just for communication with Data index and possibility of reusing it for process messaging is just a side effect?

Considering that user can use any messaging provider for Kogito I would guess the latter is more likely. So in such case I would mark tests with `@events` just for Data index integration.
Anyway the eventing in Travel agency tests is an implementation detail for us (as the purpose of our scenario is to check that Travel agency example works without any specific knowledge of implementation) so I wouldn't annotate it with `@events`.

radtriste: oki. good for me then if only dataindex is instantiated

PR #326, opened 2020-04-27
[KOGITO-128] Introducing KogitoRuntime CR

ricardozanini:
See: 
https://issues.redhat.com/browse/KOGITO-128

In this PR we introduce a new way to deploy Kogito Services, compatible with Kubernetes! :clinking_glasses: 

TODO:
- [x] Add support to add more objects to the `ServiceDeployer` (protobuf configMap must be created)
- [x] Write more unit tests to cover all features
- [x] Test on OCP
- [x] Test on Kubernetes
- [x] Write doc section
~Implement CLI~ (see: [KOGITO-2064](https://issues.redhat.com/browse/KOGITO-2064))

Many thanks for submiting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.MD#sending-a-pull-request)
- [x] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Your feature/bug fix has a unit test that verifies it
- [x] You've tested the new feature/bug fix in an actual OpenShift cluster

PR #347, opened 2020-05-14
[KOGITO-1503] Replace Kafka and Infinispan environment variable configuration with ConfigMap

xiezhang7:
Description: Replace Infinispan and Kafka configuration directly in the ConfigMap instead of relying on Environment Variables
Jira: https://issues.redhat.com/browse/KOGITO-1503


sutaakar:
For Data index the config map seems to be created correctly, but when I deployed KogitoApp the config map is still empty.

Also, it seems that these changes address just quarkus properties, but not SpringBoot ones.

xiezhang7:
> For Data index the config map seems to be created correctly, but when I deployed KogitoApp the config map is still empty.
> 
> Also, it seems that these changes address just quarkus properties, but not SpringBoot ones.

Thanks. Forgot to handle the kafka and infinispan deployment for kogito app. Will take care of that.

pkg/controller/kogitoapp/kogitoapp_controller.go
```diff
@@ -340,29 +342,38 @@ func (r *ReconcileKogitoApp) ensureKogitoImageStream(instance *v1alpha1.KogitoAp
...
 		if err != nil {
 			return true, err
 		}
 		if ready {
-			if err := kogitores.SetInfinispanEnvVars(r.client, infra, instance, requestedDeployment); err != nil {
+			envs, aps, err := kogitores.CreateInfinispanProperties(r.client, infra, instance)
+			if err != nil {
+				return true, err
+			}
+			envMap := framework.EnvVarArrayToEnvVarMap(instance.Spec.Envs)
```
sutaakar, line 371: Can this be replaced by `EnvOverride`?  https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/framework/env_var.go#L67

pkg/controller/kogitoapp/kogitoapp_controller_test.go
```diff
@@ -469,8 +469,32 @@ func TestReconcileKogitoApp_updateKogitoAppStatus(t *testing.T) {
 
 func TestReconcileKogitoApp_PersistenceEnabledWithInfra(t *testing.T) {
 	kogitoApp := createFakeKogitoApp()
+	kogitoApp.Spec.EnablePersistence = true
+	kogitoApp.Spec.EnableEvents = true
 	imgs := createFakeImages(kogitoApp.Name, map[string]string{framework.LabelKeyOrgKiePersistenceRequired: "true"})
-	fakeClient := test.CreateFakeClient([]runtime.Object{kogitoApp}, imgs, nil)
+	kogitoInfra := &v1alpha1.KogitoInfra{
```
sutaakar, line 475: Wouldn't it have sense to create another test making sure that KogitoInfra is created when reconciliation is called and persistence and/or events are enabled?

pkg/controller/kogitoapp/resource/deployment_config.go
```diff
@@ -39,15 +39,16 @@ const (
 	// ServiceAccountName is the name of service account used by Kogito Services Runtimes
 	ServiceAccountName = "kogito-service-viewer"
 
-	envVarInfinispanServerList     = "SERVER_LIST"
-	envVarInfinispanUseAuth        = "USE_AUTH"
+	appPropInfinispanServerList    = "INFINISPAN_SERVER_LIST"
```
sutaakar, line 42: These properties serve like keys only, their value doesn't have any relevance, right?
Have you considered using https://yourbasic.org/golang/iota/ for their value? To hint code readers that the value is not relevant?

pkg/controller/kogitoapp/resource/deployment_config.go
```diff
@@ -66,19 +67,25 @@ var (
...
-		envVarInfinispanUseAuth:       "QUARKUS_INFINISPAN_CLIENT_USE_AUTH",
-		envVarInfinispanUser:          "QUARKUS_INFINISPAN_CLIENT_AUTH_USERNAME",
-		envVarInfinispanPassword:      "QUARKUS_INFINISPAN_CLIENT_AUTH_PASSWORD",
-		envVarInfinispanSaslMechanism: "QUARKUS_INFINISPAN_CLIENT_SASL_MECHANISM",
+	propertiesInfinispanQuarkus = map[string]string{
+		appPropInfinispanServerList:    "quarkus.infinispan-client.server-list",
+		appPropInfinispanUseAuth:       "quarkus.infinispan-client.use-auth",
+		appPropInfinispanSaslMechanism: "quarkus.infinispan-client.sasl-mechanism",
+
+		appPropKafkaBootstrapServerList: "quarkus.kafka.bootstrap-servers",
```
sutaakar, line 75: I think that `bootstrap.servers` should be used for both Quarkus and SpringBoot.

pkg/controller/kogitoapp/resource/deployment_config.go
```diff
@@ -66,19 +67,25 @@ var (
...
-		envVarInfinispanUseAuth:       "INFINISPAN_REMOTE_USE_AUTH",
-		envVarInfinispanUser:          "INFINISPAN_REMOTE_AUTH_USERNAME",
-		envVarInfinispanPassword:      "INFINISPAN_REMOTE_AUTH_PASSWORD",
-		envVarInfinispanSaslMechanism: "INFINISPAN_REMOTE_SASL_MECHANISM",
+	propertiesInfinispanSpring = map[string]string{
+		appPropInfinispanServerList:    "infinispan.remote.server-list",
+		appPropInfinispanUseAuth:       "infinispan.remote.use-auth",
+		appPropInfinispanSaslMechanism: "infinispan.remote.sasl-mechanism",
+
+		appPropKafkaBootstrapServerList: "spring.kafka.bootstrap-servers",
```
sutaakar, line 85: Same as above

pkg/controller/kogitoapp/resource/deployment_config.go
sutaakar: https://kafka.apache.org/documentation/#bootstrap.servers

sutaakar: https://quarkus.io/guides/kafka-streams#topic-configuration

pkg/controller/kogitoapp/resource/deployment_config.go
```diff
@@ -194,54 +201,75 @@ func setReplicas(kogitoApp *v1alpha1.KogitoApp, dc *appsv1.DeploymentConfig) {
...
-		if len(dc.Spec.Template.Spec.Containers) > 0 {
-			framework.SetEnvVar(envVarKafkaBootstrapURI, uri, &dc.Spec.Template.Spec.Containers[0])
-			// let's also add a secret feature that injects all _BOOTSTRAP_SERVERS env vars with the correct uri :p
-			for _, env := range dc.Spec.Template.Spec.Containers[0].Env {
-				if strings.HasSuffix(env.Name, envVarKafkaBootstrapSuffix) {
-					framework.SetEnvVar(env.Name, uri, &dc.Spec.Template.Spec.Containers[0])
-				}
+
+		appProps[vars[appPropKafkaBootstrapServerList]] = uri
+		envs[envVarKafkaBootstrapURI] = corev1.EnvVar{Name: envVarKafkaBootstrapURI, Value: uri}
```
sutaakar, line 264: Does it have sense to create an env variable with URI when it is supplied as application property?

pkg/framework/image_metadata.go
```diff
@@ -190,11 +193,19 @@ func ExtractPrometheusConfigurationFromImage(dockerImage *dockerv10.DockerImage)
 }
 
 // IsPersistenceEnabled verifies if the image has labels indicating that persistence is enabled
-func IsPersistenceEnabled(dockerImage *dockerv10.DockerImage) (enabled bool) {
+func IsPersistenceEnabled(name, namespace string, cli *client.Client) (enabled bool) {
+	imageStreamName := types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	}
+	dockerImage, err := openshift.ImageStreamC(cli).FetchDockerImage(imageStreamName)
```
sutaakar, line 201: This implementation becomes OpenShift specific. Wouldn't it have more sense to keep previous function signature (with DockerImage) and retrieve the Docker image in kogitoapp_controller.go?

pkg/infrastructure/infinispan_integration.go
```diff
@@ -31,14 +31,16 @@ const (
...
-	infinispanEnvKeyUseAuth           string = "INFINISPAN_USEAUTH"
-	infinispanEnvKeyAuthRealm         string = "INFINISPAN_AUTHREALM"
-	infinispanEnvKeySasl              string = "INFINISPAN_SASLMECHANISM"
-	infinispanEnvKeyCredSecret        string = "INFINISPAN_CREDENTIAL_SECRET"
-	infinispanEnvKeyServerList        string = "INFINISPAN_CLIENT_SERVER_LIST"
-	infinispanEnvKeyQuarkusServerList string = "QUARKUS_INFINISPAN_CLIENT_SERVER_LIST"
+	infinispanEnvKeyUsername string = "INFINISPAN_USERNAME"
+	infinispanEnvKeyPassword string = "INFINISPAN_PASSWORD"
+
+	infinispanAppPropUseAuth           string = "quarkus.infinispan-client.use-auth"
```
sutaakar, line 37: There properties are duplicated in deployment_config.go

pkg/infrastructure/services/configmap.go
```diff
@@ -51,24 +54,40 @@ func GetAppPropConfigMapContentHash(name, namespace string, cli *client.Client)
 		return "", nil, err
 	}
 
+	appPropsToApply := map[string]string{}
 	if exist {
-		if _, ok := configMap.Data[appPropContentKey]; !ok {
-			exist = false
+		if _, ok := configMap.Data[appPropContentKey]; ok {
```
sutaakar, line 59: Can you place this code to separate function to make the function flow easier to read?
Something like getApplicationProperties(configMap &corev1.ConfigMap)

pkg/infrastructure/services/kafka.go
```diff
@@ -20,19 +20,20 @@ import (
...
-	quarkusTopicBootstrapEnvVar = "MP_MESSAGING_%s_%s_BOOTSTRAP_SERVERS"
+	quarkusTopicBootstrapAppProp = "mp.messaging.%s.%s.bootstrap.servers"
 )
 
-var quarkusBootstrapEnvVars = []string{"KAFKA_BOOTSTRAP_SERVERS", "QUARKUS_KAFKA_BOOTSTRAP_SERVERS"}
+var quarkusBootstrapEnvVars = []string{"KAFKA_BOOTSTRAP_SERVERS"}
 
-// fromKafkaTopicToQuarkusEnvVar transforms a given Kafka Topic name into a environment variable to be read by Quarkus Kafka client used by Kogito Services
-func fromKafkaTopicToQuarkusEnvVar(topic KafkaTopicDefinition) string {
+var quarkusBootstrapAppProps = []string{"quarkus.kafka.bootstrap-servers"}
```
sutaakar, line 31: Same as above.

pkg/controller/kogitoapp/resource/deployment_config.go
xiezhang7: I am not sure if the job service works without this env according to the [scripts](https://github.com/kiegroup/kogito-images/blob/7a13116a98b7f6de9fba2ff61975106187ddc2c1/modules/kogito-jobs-service/added/launch/kogito-jobs-service.sh#L38), so I kept it.

xiezhang7:
> can you please update the README file with an example about how user can set/configure the infinispan and kafka properties?

@spolti I don't think user should configure the infinispan or kafka properties with the configmap manually. Because user is supposed to use the CR to control the infinispan and kafka settings. 

The infinispan and kafka properties in the configmap are actually from the CR settings. Even if user set properties in the configmap, the properties will be replaced by the settings from the CR.

spolti:
> > can you please update the README file with an example about how user can set/configure the infinispan and kafka properties?
> 
> @spolti I don't think user should configure the infinispan or kafka properties with the configmap manually. Because user is supposed to use the CR to control the infinispan and kafka settings.
> 
> The infinispan and kafka properties in the configmap are actually from the CR settings. Even if user set properties in the configmap, the properties will be replaced by the settings from the CR.

yeah, this needs to be clarified in the docs :)

pkg/controller/kogitoapp/resource/deployment_config.go
```diff
@@ -194,54 +203,72 @@ func setReplicas(kogitoApp *v1alpha1.KogitoApp, dc *appsv1.DeploymentConfig) {
 	dc.Spec.Replicas = replicas
 }
 
-// SetInfinispanEnvVars sets Infinispan variables to the given KogitoApp instance DeploymentConfig by reading information from the KogitoInfra
-func SetInfinispanEnvVars(cli *client.Client, kogitoInfra *v1alpha1.KogitoInfra, kogitoApp *v1alpha1.KogitoApp, dc *appsv1.DeploymentConfig) error {
-	if dc != nil && kogitoApp != nil &&
+// CreateInfinispanProperties creates Infinispan properties by reading information from the KogitoInfra
+func CreateInfinispanProperties(cli *client.Client, kogitoInfra *v1alpha1.KogitoInfra, kogitoApp *v1alpha1.KogitoApp) (envs []corev1.EnvVar, appProps map[string]string, err error) {
```
sutaakar, line 207: Please coordinate these changes with https://github.com/kiegroup/kogito-cloud-operator/pull/326
The `CreateInfinispanProperties` and `CreateKafkaProperties` functions don't have to be in `deployment_config.go ` as they don't have anything to do with DeploymentConfig any more, can be transferred to infrastructure package.

pkg/infrastructure/services/configmap_test.go
```diff
@@ -143,10 +55,31 @@ func TestGetAppPropConfigMapContentHash(t *testing.T) {
 		wantErr bool
 	}{
 		{
-			"New ConfigMap",
+			"New Empty ConfigMap 1",
```
sutaakar, line 58:
```suggestion
			"New Empty ConfigMap",
```

pkg/infrastructure/services/configmap_test.go
```diff
@@ -143,10 +55,31 @@ func TestGetAppPropConfigMapContentHash(t *testing.T) {
...
+					Namespace: serviceNamespace,
+				},
+				Data: map[string]string{
+					appPropFileName: defaultAppPropContent,
+				},
+			},
+			false,
+		},
+		{
+			"New Empty ConfigMap 2",
```
sutaakar, line 78:
```suggestion
			"Nil ConfigMap",
```

pkg/controller/kogitoapp/resource/deployment_config.go
xiezhang7: These two functions are specialized for kogito app only. They are supposed to be deprecated as part of https://issues.redhat.com/browse/KOGITO-1998 . And there is already generic code for creating infinispan and kafka properties in the `infrastructure` package. So I moved the two functions to two new files and kept them in the kogito app package. Please take a look. Thanks.

ricardozanini:
Please hold this until #326 is merged.

pkg/controller/kogitoapp/resource/deployment_config.go
sutaakar: Will recheck once the https://github.com/kiegroup/kogito-cloud-operator/pull/326 is merged and this PR is rebased.

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -225,7 +225,9 @@ func (k *KogitoServiceSpec) AddResourceLimit(name, value string) {
 func (k *KogitoServiceSpec) GetDeploymentLabels() map[string]string { return k.DeploymentLabels }
 
 // SetDeploymentLabels ...
-func (k *KogitoServiceSpec) SetDeploymentLabels(labels map[string]string) { k.DeploymentLabels = labels }
+func (k *KogitoServiceSpec) SetDeploymentLabels(labels map[string]string) {
+	k.DeploymentLabels = labels
+}
```
ricardozanini, line 230: Please let this on one single line for better readability

xiezhang7:
> If I change the Kafka/Infinispan configuration manually in the `configMap`, having then set into the CR like `useKogitoInfra: true`, the operator will change them back? But still, I can add my own properties to the configMap, right? That's why you added the `mergo` dependency?

Yes. The Kafak and Infinispan properties that are derived from the CR will override the same properties user manually changes in the ConfigMap. All other properties that user adds to the ConfigMap will not be modified by the operator.

pkg/apis/app/v1alpha1/kogitoservices_types.go
xiezhang7: Changed it back. Thanks.

pkg/controller/kogitoapp/resource/kafka.go
```diff
@@ -0,0 +1,68 @@
...
+
+	envVarKafkaBootstrapSuffix = "_BOOTSTRAP_SERVERS"
+)
+
+var (
+	propertiesKafkaQuarkus = map[kafkaProperty]string{
+		appPropKafkaBootstrapServerList: services.QuarkusBootstrapAppProp,
+	}
+	propertiesKafkaSpring = map[kafkaProperty]string{
+		appPropKafkaBootstrapServerList: "spring.kafka.bootstrap-servers",
```
sutaakar, line 39: Shouldn't this be externalized to a constant like `services.QuarkusBootstrapAppProp` above?

pkg/infrastructure/services/resources.go
```diff
@@ -197,11 +200,10 @@ func (s *serviceDeployer) applyKafkaConfigurations(deployment *appsv1.Deployment
...
 		framework.SetEnvVar(enableEventsEnvKey, "true", &deployment.Spec.Template.Spec.Containers[0])
 		for _, kafkaTopic := range s.definition.KafkaTopics {
-			framework.SetEnvVar(fromKafkaTopicToQuarkusEnvVar(kafkaTopic), URI, &deployment.Spec.Template.Spec.Containers[0])
-		}
-		for _, kafkaEnv := range quarkusBootstrapEnvVars {
-			framework.SetEnvVar(kafkaEnv, URI, &deployment.Spec.Template.Spec.Containers[0])
+			appProps[fromKafkaTopicToQuarkusAppProp(kafkaTopic)] = URI
 		}
+		framework.SetEnvVar(quarkusBootstrapEnvVar, URI, &deployment.Spec.Template.Spec.Containers[0])
+		appProps[QuarkusBootstrapAppProp] = URI
```
sutaakar, line 206: Shouldn't be applied SpringBoot property here in case of SpringBoot runtime?

sutaakar:
The code looks good (minor comments above).
Tomorrow I will build it and test it on OpenShift.

pkg/controller/kogitoapp/resource/kafka.go
xiezhang7: Corrected. Thanks.

pkg/infrastructure/services/resources.go
xiezhang7: Corrected. Thanks.

pkg/infrastructure/services/resources.go
```diff
@@ -197,10 +200,13 @@ func (s *serviceDeployer) applyKafkaConfigurations(deployment *appsv1.Deployment
 	if len(URI) > 0 {
 		framework.SetEnvVar(enableEventsEnvKey, "true", &deployment.Spec.Template.Spec.Containers[0])
 		for _, kafkaTopic := range s.definition.KafkaTopics {
-			framework.SetEnvVar(fromKafkaTopicToQuarkusEnvVar(kafkaTopic), URI, &deployment.Spec.Template.Spec.Containers[0])
+			appProps[fromKafkaTopicToQuarkusAppProp(kafkaTopic)] = URI
```
radtriste, line 203: We set Quarkus even for SpringBoot apps ?

pkg/infrastructure/services/resources.go
```diff
@@ -197,10 +200,13 @@ func (s *serviceDeployer) applyKafkaConfigurations(deployment *appsv1.Deployment
...
 		}
-		for _, kafkaEnv := range quarkusBootstrapEnvVars {
-			framework.SetEnvVar(kafkaEnv, URI, &deployment.Spec.Template.Spec.Containers[0])
+		framework.SetEnvVar(quarkusBootstrapEnvVar, URI, &deployment.Spec.Template.Spec.Containers[0])
```
radtriste, line 205: Why setting the env var if we set the app prop just after ?

pkg/infrastructure/services/resources.go
ricardozanini: We should not.

xiezhang7: Corrected. Thanks.

pkg/infrastructure/services/resources.go
xiezhang7: It seems the env is still needed by [job service](https://github.com/kiegroup/kogito-images/blob/7a13116a98b7f6de9fba2ff61975106187ddc2c1/modules/kogito-jobs-service/added/launch/kogito-jobs-service.sh#L38).

radtriste: ok

This one is for Quarkus right ?
We should also not set it for Spring Boot or ?

xiezhang7: Corrected. Thanks.

pkg/controller/kogitoapp/resource/test_util.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
```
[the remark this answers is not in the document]
xiezhang7: Renamed, thanks.

PR #349, opened 2020-05-15
KOGITO-2192: BDD tests: Should provide a file with created namespaces

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-2192
Description: Create a namespace history log to delete the namespaces in case of test framework didn't end well and therefore it could not delete it.

Requirements:
- Create history file
- Append the namespaces to the history file (for parallel test executions)
- Delete namespace from history in case of the test deleted the namespace
- Provide an utility go function to prune the rest of namespaces from history

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -83,6 +83,7 @@ pipeline {
                     dir("${WORKING_DIR}") {
                         archiveArtifacts artifacts: 'test/logs/**/*.log', allowEmptyArchive: true
                         junit testResults: 'test/logs/**/junit.xml', allowEmptyResults: true
+                        sh "cd test && go run scripts/prune_namespaces.go"
                     }
                 }
             }
--- a/hack/run-tests.sh
+++ b/hack/run-tests.sh
@@ -123,6 +123,7 @@ FEATURE=""
 TIMEOUT=240
 DEBUG=false
 CRDS_UPDATE=true
+KEEP_NAMESPACE=false
 LOAD_DEFAULT_CONFIG=false
 
 while (( $# ))
@@ -292,6 +293,7 @@ case $1 in
     shift
   ;;
   --keep_namespace)
+    KEEP_NAMESPACE=true
     addParam "--tests.keep-namespace"
     shift
   ;;
@@ -358,6 +360,14 @@ if ${CRDS_UPDATE}; then
   deploy_folder=`mktemp -d`
   download_remote_crds ${deploy_folder} ${MASTER_RAW_URL}
   apply_crds ${deploy_folder}
+fi
+
+if [ "${KEEP_NAMESPACE}" = "false" ]; then
+  echo "-------- Pruning namespaces"
+  cd ${SCRIPT_DIR}/../test
+  go run scripts/prune_namespaces.go
+  echo "Pruning namespaces done."
+  cd -
 fi
 
 exit ${exit_code}
--- a/test/README.md
+++ b/test/README.md
@@ -90,3 +90,12 @@ The program will wait until you launch the Debug mode from VS Code (next step).
 Launch debug session by selecting Run/Start Debugging from top menu.
 
 Further information [here](https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code).
+
+### Prune namespaces
+
+In case you stopped a test execution, the test framework won't delete the namespaces in the cluster. So to avoid waste of resources in the cluster, you need to manually delete these namespaces. In order to ease this, now we have a namespace history log and an utility to do this by running:
+
+```sh
+cd test
+go run scripts/prune_namespaces.go
+``` 
--- a/test/framework/kubernetes.go
+++ b/test/framework/kubernetes.go
@@ -51,46 +51,6 @@ func InitKubeClient() error {
 	return nil
 }
 
-// CreateNamespace creates a new namespace
-func CreateNamespace(namespace string) error {
-	GetLogger(namespace).Infof("Create namespace %s", namespace)
-	_, err := kubernetes.NamespaceC(kubeClient).Create(namespace)
-	if err != nil {
-		return fmt.Errorf("Cannot create namespace %s: %v", namespace, err)
-	}
-	return nil
-}
-
-// DeleteNamespace deletes a namespace
-func DeleteNamespace(namespace string) error {
-	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
-	GetLogger(namespace).Infof("Delete namespace %s", namespace)
-	err := kubernetes.ResourceC(kubeClient).Delete(ns)
-	if err != nil {
-		return fmt.Errorf("Cannot delete namespace %s: %v", namespace, err)
-	}
-	return nil
-}
-
-// IsNamespace checks wherher a namespace exists
-func IsNamespace(namespace string) (bool, error) {
-	ns, err := kubernetes.NamespaceC(kubeClient).Fetch(namespace)
-	if err != nil {
-		return false, fmt.Errorf("Cannot create namespace %s: %v", namespace, err)
-	}
-	return ns != nil, nil
-}
-
-// OperateOnNamespaceIfExists do some operations on the namespace if that one exists
-func OperateOnNamespaceIfExists(namespace string, operate func(namespace string) error) error {
-	if ok, er := IsNamespace(namespace); er != nil {
-		return fmt.Errorf("Error while checking namespace: %v", er)
-	} else if ok {
-		return operate(namespace)
-	}
-	return nil
-}
-
 // WaitForPodsWithLabel waits for pods with specific label to be available and running
 func WaitForPodsWithLabel(namespace, labelName, labelValue string, numberOfPods, timeoutInMin int) error {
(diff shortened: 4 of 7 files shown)


test/README.md
```diff
@@ -90,3 +90,13 @@ The program will wait until you launch the Debug mode from VS Code (next step).
...
 Further information [here](https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code).
+
+## Miscellaneous
+
+### Prune namespaces
+
+In case you stopped a test execution, the test framework won't delete the namespaces in the cluster. So to avoid waste of resources in the cluster, you need to manually delete these namespaces. In order to ease this, now we have a namespace history log and an utility to do this by running:
+
+```sh
+go run utils/prune_namespaces.go
```
Sgitario, line 101: @radtriste I provided this function to be used by the pipeline and I think it could be useful for us too (I tend to stop test executions when running the tests locally if I spot something or don't want to wait for the test to finish).

test/framework/kubernetes.go
```diff
@@ -143,7 +145,7 @@ func IsPodRunning(pod *corev1.Pod) bool {
 }
 
 // IsPodStatusConditionReady returns true if all pod's containers are ready (really running)
-func IsPodStatusConditionReady(pod *corev1.Pod) bool{
+func IsPodStatusConditionReady(pod *corev1.Pod) bool {
```
Sgitario, line 148: VS automatically added this whitespace here, I think this is ok. Otherwise, let me know and I will revert this change.

test/README.md
```diff
@@ -90,3 +90,13 @@ The program will wait until you launch the Debug mode from VS Code (next step).
 Launch debug session by selecting Run/Start Debugging from top menu.
 
 Further information [here](https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code).
+
+## Miscellaneous
```
radtriste, line 94: I would not create a new big section. I would let it in `Development` section or move to the `Run tests` section

test/README.md
radtriste: cool thing ! will have a look.
maybe we could also add a section in jenkins pipeline(s) to always call this script after the tests ran, so it can clean if any interruption :)

radtriste: and could the `run-tests.sh` script call this one after running the tests ?
(if no `keep-namespace` is set of course)
that would give a second security I think

test/utils/prune_namespaces.go
```diff
@@ -0,0 +1,14 @@
+package main
```
radtriste, line 1: could we call the folder like `scripts` or similar ?
I find `utils` as a test package name ...

test/framework/kubernetes.go
```diff
@@ -69,6 +70,7 @@ func DeleteNamespace(namespace string) error {
 	if err != nil {
 		return fmt.Errorf("Cannot delete namespace %s: %v", namespace, err)
 	}
+	OnNamespacePostDeleted(namespace)
```
radtriste, line 73: we could also migrate the `CreateNamespace` and `DeleteNamespace` methods to `namespace.go` ?
wdyt ?

test/framework/namespace.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+const (
+	namespaceLogFile = "logs/namespace_history.log"
```
radtriste, line 25: I like you put that file in logs, so we can ignore it :)

test/framework/kubernetes.go
radtriste: I am ok with that

test/framework/namespace.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	defer file.Close()
+	if _, err = file.WriteString(fmt.Sprintf("%s\n", namespace)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func removeNamespacesFromHistory(namespace string) error {
```
sutaakar, line 73: From the function signature I would say it should be called `removeNamespaceFromHistory`

test/framework/namespace.go
```diff
@@ -0,0 +1,84 @@
...
+	namespaces := GetNamespacesInHistory()
+	var newNamespaces []string
+	for _, oldNamespace := range namespaces {
+		if !strings.Contains(oldNamespace, namespace) {
```
sutaakar, line 77: Wouldn't it be safer to use equality of strings instead of contains? Equality would make sure that only exact match is found.

sutaakar, line 77: Wouldn't it be safer to use `==` ?

test/README.md
Sgitario: It makes sense to prune namespaces in run-tests.sh (in case when we stop the test execution) and in the pipeline (in case when the pipeline stops).
Added.

test/framework/namespace.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
```
sutaakar, line 1: IMHO the file name should be adjusted so it won't be confused with general namespace operations. What about for example `namespace_history.go`?

test/utils/prune_namespaces.go
Sgitario: Sure. Renamed to scritps.

test/framework/kubernetes.go
radtriste: if following comment from Karel (https://github.com/kiegroup/kogito-cloud-operator/pull/349#discussion_r425678478), then this is useless ;)

test/framework/namespace.go
sutaakar: Or as a second option to move namespace functionality here as mentioned by @radtriste

Sgitario: Actually, in my first implementation, I named this script as "namespace_history.go", but then I changed it to just "namespace.go" because I wrote the methods OnNamespacePostCreated and OnNamespacePostDeleted in a way that can be refactored to do more things if needed in the future apart from just keeping this history log file. 

I would prefer to move the CreateNamespace and DeleteNamespace from kubernetes to here instead as @radtriste suggested. Are you ok with?

test/framework/namespace.go
Sgitario: As the namespace will contain the whole namespace, I thought it would make more sense to use the contains. But I agree with you, this is not necessary.
Changed to !=.

test/framework/namespace.go
Sgitario: Thanks for spotting!

test/framework/namespace.go
sutaakar: sure, fine for me
Please move also `IsNamespace` and `OperateOnNamespaceIfExists` to have it all in one place.

test/README.md
Sgitario: Changed.

test/framework/namespace.go
Sgitario: Done

hack/run-tests.sh
```diff
@@ -360,4 +362,12 @@ if ${CRDS_UPDATE}; then
   apply_crds ${deploy_folder}
 fi
 
+if [ "${KEEP_NAMESPACE}" = "false" ]; then
+  echo "-------- Pruning namespaces"
+  cd ${SCRIPT_DIR}/../test
```
Sgitario, line 367: I need to make the test directory the work directory. We can avoid this by sending the history log file in the argument:
```sh
go run test/scripts/prune_namespaces.go test/logs/namespace_history.log
```
But I would need to tweak the GetNamespacesInHistory method a bit for doing so and I wanted to avoid to over-complicate things.
Also, I don't think this would be an issue when running the pipeline.

test/README.md
```diff
@@ -90,3 +90,11 @@ The program will wait until you launch the Debug mode from VS Code (next step).
 Launch debug session by selecting Run/Start Debugging from top menu.
 
 Further information [here](https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code).
+
+### Prune namespaces
+
+In case you stopped a test execution, the test framework won't delete the namespaces in the cluster. So to avoid waste of resources in the cluster, you need to manually delete these namespaces. In order to ease this, now we have a namespace history log and an utility to do this by running:
+
```
radtriste, line 97: I would precise to be in test folder to execute this

Jenkinsfile.nightly
```diff
@@ -83,6 +83,7 @@ pipeline {
                     dir("${WORKING_DIR}") {
                         archiveArtifacts artifacts: 'test/logs/**/*.log', allowEmptyArchive: true
```
Sgitario, line 84: @radtriste I didn't archive the history log file in the end as the sh command will remove it and I don't think it will be very necessary as you already said. What do you think?

hack/run-tests.sh
radtriste: it is ok like that

Jenkinsfile.nightly
radtriste: fair enough

test/README.md
Sgitario: done

radtriste:
Seems the prune has been correctly called into the PR check pipeline.
Merging this one

PR #352, opened 2020-05-20
KOGITO-1613: Cucumber tests: Test upload source files (dmn, drl, bpmn…

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1613
Description: With https://issues.redhat.com/browse/KOGITO-256, new functionality has been introduced to upload application from source files (dmn, drl, bpmn, bpmn2) with CLI

Covered:
- Test with DMN
- Test with BPMN
- Test with folders that contain properties and bpmn files altogether

In order to also cover DRL and/or BPMN2, we would need to create a new kogito example that does not require any Java sources.

### DIFF
--- a/pkg/client/openshift/buildconfig.go
+++ b/pkg/client/openshift/buildconfig.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	checkBcRetries         = 4
+	checkBcRetries         = 30
 	checkBcRetriesInterval = 2 * time.Second
 )
 
--- a/test/features/deploy_source_files.feature
+++ b/test/features/deploy_source_files.feature
@@ -1,0 +1,51 @@
+@cli
+Feature: Deploy source files (dmn, drl, bpmn, bpmn2, ...) with CLI
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+
+  @smoke
+  Scenario: Deploy .dmn source files with CLI
+    Given Clone Kogito examples into local directory
+    
+    When Deploy file "Traffic Violation.dmn" from example service "dmn-quarkus-example"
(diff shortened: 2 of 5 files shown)


pkg/client/openshift/buildconfig.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	checkBcRetries         = 4
+	checkBcRetries         = 30
```
Sgitario, line 36: This is the issue that @radtriste found where the CLI was not waiting enough time to check whether the BuildConfig was present. Using 30 retries (which means 2 minutes in total) works fine.

Sgitario:
I ran the new feature successfully in OCP 4.5 (production).

test/features/deploy_source_files.feature
```diff
@@ -0,0 +1,50 @@
+@smoke
```
sutaakar, line 1: Please add a `@cli` tag, these tests can be executed just with CLI.
Also, should they be marked as smoke?

sutaakar:
>  In order to also cover DRL and/or BPMN2, we would need to create a new kogito example that does not require any Java sources.

I don't think we should cover them here, as this is out of scope for operator tests. Here we should just check that file and folder upload works in general.

test/features/deploy_source_files.feature
Sgitario: Will add the @cli, about @smoke, maybe not as is not considered basic functionality, right? @radtriste

radtriste: I would not put the whole feature as smoke. Just one like the `Scenario: Deploy .dmn source files with CLI`

Sgitario: Done

evacchi:
status? :)

PR #361, opened 2020-05-27
[KOGITO-1729] Consolidate HTTP Request parameters into one struct type

MarianMacik:
https://issues.redhat.com/browse/KOGITO-1729

JIRA was created as a result of the feedback in https://github.com/kiegroup/kogito-cloud-operator/pull/257#discussion_r403309366.

Things to discuss:

1. Name of the struct is `HTTPRequestInfo`.
I didn't want to name it `HTTPRequest` as then the parameter name, e.g. `request`, would clash with the actual request from Go's http package and it would be confusing in methods like for example here:
https://github.com/kiegroup/kogito-cloud-operator/blob/e9069757dcc2fa0068019cc3a76d49a90719f139/test/framework/http.go#L53-L56 I was also considering `HTTPRequestData` but this isn't just data and `HTTPRequestMetaData` but that's too long. The parameter would then need to be something like `requestMetaData` etc.

1. I think the goal of the JIRA was achieved, HTTP methods have much shorter signatures now, but now we need to call the constructor to create an `HTTPRequestInfo` struct each time we want to issue a request. Also some logging methods are now longer as struct fields needs to be accessed using the prefix.

1. I have also found a small bug, so I decided to fix that:
https://github.com/kiegroup/kogito-cloud-operator/blob/e9069757dcc2fa0068019cc3a76d49a90719f139/test/steps/http.go#L91-L92

Many thanks for submiting your Pull Request :heart: :racing_car:! 

### DIFF
--- a/test/framework/http.go
+++ b/test/framework/http.go
@@ -26,8 +26,15 @@ import (
 	"time"
 )
 
-// HTTPRequestResult represents the success or error of an HTTP request
-type HTTPRequestResult string
+type (
+	// HTTPRequestResult represents the success or error of an HTTP request
+	HTTPRequestResult string
+
+	// HTTPRequestInfo structure encapsulates all information needed to execute an HTTP request
+	HTTPRequestInfo struct {
+		HTTPMethod, URI, Path, BodyFormat, BodyContent string
+	}
+)
 
 const (
 	// HTTPRequestResultSuccess in case of success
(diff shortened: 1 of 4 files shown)


test/framework/http.go
```diff
@@ -199,11 +206,22 @@ func IsHTTPResponseArraySize(namespace, httpMethod, uri, path string, bodyFormat
...
+	resultBody, err := ExecuteHTTPRequestWithStringResponse(namespace, requestInfo)
 	if err != nil {
 		return false, err
 	}
 
 	return strings.Contains(resultBody, responseContent), nil
 }
+
+// NewHTTPRequestInfo constructor creates a new HTTPRequestInfo struct
+func NewHTTPRequestInfo(httpMethod, uri, path, bodyFormat, bodyContent string) HTTPRequestInfo {
```
sutaakar, line 219: What about providing separate function for GET and POST types?
For GET type you don't need to supply bodyFormat and bodyContent.

MarianMacik: Hmm, I was actually thinking about it, but just to get away with GET and POST strings. But your suggestion is even better, thanks!

MarianMacik:
Added 2 methods:
* NewGETHTTPRequestInfo
* NewPOSTHTTPRequestInfo

I tried to change it to NewHTTPRequestInfoGET, but that would change the meaning slightly. Also I think GET should be all caps, so inevitably we will have a lot of capital letters together. If you know a better approach, feel free to suggest it.

sutaakar:
Looks good to me.

spolti:
Folks, this PR was not merged yet because the CI is reporting error?

sutaakar:
I guess so, seems to be approved and passing checks.

MarianMacik:
Hi, on Friday I was unable to restart the build, but @radtriste did and not it is finally green! Merging. Thanks all for review.

PR #362, opened 2020-05-28
KOGITO-2327: KogitoRuntime Support in Kubernetes

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/KOGITO-2327
Description: KogitoRuntime Support in Kubernetes

Pending tasks:
- [x] Run basic test
- [x] Install Openshift community catalog in Kubernetes (I propose to do this as part of another ticket using a real Kubernetes instance)
- [x] Run tests with Infinispan and Kafka (Tests with Kafka are not working, but not because these changes)
- [x] Investigate metrics log errors when deploying the Kogito Operator (raised KOGITO-2404)

How to test locally:

1.- Run minikube (v1.11 or later):

```
minikube start --addons ingress --addons olm --driver kvm2
```

The OC client will be configured transparently for us.

2.- Configure your /etc/hosts file with:

```
[YOUR K8S IP] data-index
[YOUR K8S IP] process-springboot-example
[YOUR K8S IP] process-quarkus-example
```

To find the IP, run "minikube ip".

3.- Login to the quay.io repository using your account:

```
docker login quay.io
```

And make your repositories public:

```
process-springboot-example-default
process-quarkus-example-default
process-springboot-example-persistence
process-quarkus-example-persistence
process-springboot-example-events-persistence
process-quarkus-example-events-persistence
```

Note that these repositories are created in the next step. 

4.- Run the tests:

```
../hack/run-tests.sh --feature "features/deploy_kogito_runtime.feature" --runtime_application_image_namespace <quay account> --runtime_application_image_registry quay.io --cluster kubernetes --local-cluster
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,7 @@ ci=
 cr_deployment_only=false
 load_default_config=false
 container_engine=
+domain_suffix=
 # operator information
 operator_image=
 operator_tag=
@@ -123,12 +124,14 @@ dry_run=false
 keep_namespace=false
 disabled_crds_update=false
 namespace_name=
+local_cluster=false
 run-tests:
 	declare -a opts \
 	&& if [ "${debug}" = "true" ]; then opts+=("--debug"); fi \
 	&& if [ "${smoke}" = "true" ]; then opts+=("--smoke"); fi \
 	&& if [ "${performance}" = "true" ]; then opts+=("--performance"); fi \
 	&& if [ "${local}" = "true" ]; then opts+=("--local"); fi \
+	&& if [ "${local_cluster}" = "true" ]; then opts+=("--local_cluster"); fi \
 	&& if [ "${cr_deployment_only}" = "true" ]; then opts+=("--cr_deployment_only"); fi \
 	&& if [ "${show_scenarios}" = "true" ]; then opts+=("--show_scenarios"); fi \
 	&& if [ "${show_steps}" = "true" ]; then opts+=("--show_steps"); fi \
@@ -168,6 +171,7 @@ run-tests:
 		--namespace_name ${namespace_name} \
 		--load_factor ${load_factor} \
 		--container_engine ${container_engine} \
+		--domain_suffix ${domain_suffix} \
 		$${opts_str}
 
 .PHONY: run-smoke-tests
--- a/hack/run-tests.sh
+++ b/hack/run-tests.sh
@@ -43,11 +43,12 @@ function usage(){
   printf "\n--smoke\n\tFilter to run only the tests tagged with '@smoke'."
   printf "\n--performance\n\tFilter to run only the tests tagged with '@performance'. If not provided and the tag itself is not specified, these tests will be ignored."
   printf "\n--load_factor {INT_VALUE}\n\tSet the tests load factor. Useful for the tests to take into account that the cluster can be overloaded, for example for the calculation of timeouts. Default value is 1."
-  printf "\n--local\n\tSpecify whether you run test in local."
+  printf "\n--local\n\tSpecify whether you run test in local using either a local or remote cluster."
   printf "\n--ci {CI_NAME}\n\tSpecify whether you run test with ci, give also the name of the CI."
   printf "\n--cr_deployment_only\n\tUse this option if you have no CLI to test against. It will use only direct CR deployments."
   printf "\n--load_default_config\n\tTo be used if you want to directly use the default test config contained into ${SCRIPT_DIR}/../test/.default_config"
   printf "\n--container_engine\n\tTo be used if you want to specify engine to interact with images and containers. Default is docker."
+  printf "\n--domain_suffix\n\tTo be used if you want to set a domain suffix for exposed services. Ignored when running tests on Openshift."
 
   # operator information
   printf "\n--operator_image {NAME}\n\tOperator image name. Default is 'quay.io/kiegroup/kogito-cloud-operator' one."
@@ -90,6 +91,7 @@ function usage(){
   printf "\n--keep_namespace\n\tDo not delete namespace(s) after scenario run (WARNING: can be resources consuming ...)."
   printf "\n--disabled_crds_update\n\tDisable the update of CRDs."
   printf "\n--namespace_name\n\tSpecify name of the namespace which will be used for scenario execution (intended for development purposes)."
+  printf "\n--local_cluster\n\tSpecify whether you run test using a local cluster."
   printf "\n"
 }
 
@@ -203,6 +205,10 @@ case $1 in
   --container_engine)
     shift
     if addParamKeyValueIfAccepted "--tests.container-engine" ${1}; then shift; fi
+  ;;
+  --domain_suffix)
+    shift
+    if addParamKeyValueIfAccepted "--tests.domain-suffix" ${1}; then shift; fi
   ;;
 
   # operator information
@@ -330,6 +336,10 @@ case $1 in
   --namespace_name)
     shift
     if addParamKeyValueIfAccepted "--tests.dev.namespace-name" ${1}; then shift; fi
+  ;;
+  --local_cluster)
+    addParam "--tests.dev.local-cluster"
+    shift
   ;;
 
   # others
--- a/test/config/config.go
+++ b/test/config/config.go
@@ -31,6 +31,7 @@ type TestConfig struct {
 	ciName           string
 	crDeploymentOnly bool
 	containerEngine  string
+	domainSuffix     string
 
 	// operator information
 	operatorImageName string
@@ -70,6 +71,7 @@ type TestConfig struct {
 	dryRun        bool
 	keepNamespace bool
 	namespaceName string
+	localCluster  bool
 }
 
 const (
@@ -100,10 +102,11 @@ func BindFlags(set *flag.FlagSet) {
 	set.BoolVar(&env.smoke, prefix+"smoke", false, "Launch only smoke tests")
 	set.BoolVar(&env.performance, prefix+"performance", false, "Launch performance tests")
 	set.IntVar(&env.loadFactor, prefix+"load-factor", defaultLoadFactor, "Set the tests load factor. Useful for the tests to take into account that the cluster can be overloaded, for example for the calculation of timeouts. Default value is 1.")
-	set.BoolVar(&env.localTests, prefix+"local", false, "If tests are launch on local machine")
+	set.BoolVar(&env.localTests, prefix+"local", false, "If tests are launch on local machine using either a local or remote cluster")
 	set.StringVar(&env.ciName, prefix+"ci", "", "If tests are launch on ci machine, give the CI name")
 	set.BoolVar(&env.crDeploymentOnly, prefix+"cr-deployment-only", false, "Use this option if you have no CLI to test against. It will use only direct CR deployments.")
 	set.StringVar(&env.containerEngine, prefix+"container-engine", defaultContainerEngine, "Engine used to interact with images and local containers.")
+	set.StringVar(&env.domainSuffix, prefix+"domain-suffix", "", "Set the domain suffix for exposed services. Ignored when running tests on Openshift.")
 
 	// operator information
 	set.StringVar(&env.operatorImageName, prefix+"operator-image-name", defaultOperatorImageName, "Operator image name")
@@ -143,6 +146,7 @@ func BindFlags(set *flag.FlagSet) {
 	set.BoolVar(&env.dryRun, prefix+"dry-run", false, "Dry Run the tests.")
 	set.BoolVar(&env.keepNamespace, prefix+"keep-namespace", false, "Do not delete namespace(s) after scenario run (WARNING: can be resources consuming ...)")
 	set.StringVar(&env.namespaceName, developmentOptionsPrefix+"namespace-name", "", "Use the specified namespace for scenarios, don't generate random namespace.")
+	set.BoolVar(&env.localCluster, developmentOptionsPrefix+"local-cluster", false, "If tests are launch using a local cluster")
 }
 
 // tests configuration
@@ -180,6 +184,11 @@ func IsCrDeploymentOnly() bool {
 // GetContainerEngine returns engine used to interact with images and local containers
 func GetContainerEngine() string {
 	return env.containerEngine
+}
+
+// GetDomainSuffix returns the domain suffix for exposed services
+func GetDomainSuffix() string {
+	return env.domainSuffix
 }
 
 // operator information
@@ -327,4 +336,9 @@ func IsKeepNamespace() bool {
 // GetNamespaceName return namespace name if it was defined
 func GetNamespaceName() string {
 	return env.namespaceName
+}
+
+// IsLocalCluster return whether tests are executed using a local cluster
+func IsLocalCluster() bool {
+	return env.localCluster
 }
--- a/test/framework/check_setup.go
+++ b/test/framework/check_setup.go
@@ -1,0 +1,43 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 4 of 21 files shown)


test/framework/operator.go
```diff
@@ -69,6 +72,17 @@ var (
...
+	// CommunityCatalog operator catalog for community
+	CommunityCatalog = operatorCatalog{
+		source:    "community-operators",
+		namespace: "openshift-marketplace",
+	}
+	// OperatorHubCatalog operator catalog of Operator Hub
+	OperatorHubCatalog = operatorCatalog{
+		source:    "operatorhubio-catalog",
+		namespace: "olm",
+	}
```
Sgitario, line 85: SMELL: I don't think this is related to kubernetes or openshift.... However, minikube installs the operatorhubio catalog in the olm namespace. I need to see if we can install the openshift-marketplace in Kubernetes.

Sgitario:
# Known Issues with Minikube

## Minikube cannot pull Docker images

Check what driver is using minikube. It should be using "kvm2" (there is a known issue with "docker").

## The operator is not started/Deployed
This might be caused because the OLM deployment stopped or failed. Go to the namespace "olm" and check that all the pods starting with "operatorhubio-catalog" are running. If not, delete it and Kubernetes will recreate it. If this was the issue, the test will be automatically corrected. 

## Random incompatibilities with Docker when running minikube and the tests

Minikube starts up Docker internally, so it might occur an inconsistency between the local Docker instance and Minikube. The problem is when the tests need Docker. If so, use podman instead (using  --container_engine podman)

Makefile
```diff
@@ -89,6 +89,7 @@ ci=
 cr_deployment_only=false
 load_default_config=false
 container_engine=
+cluster=openshift
```
ricardozanini, line 92: Just a minor observation: we have a function in `client` package to discover in which cluster we are. Not sure if your intention is to run this command telling explicit to the test framework where you are, but I think you could leverage from this function instead and run Kubernetes specifics when `client.IsOpenShift == false`.

Sgitario: Will try to run the tests using "client.NewForConsole().IsOpenshift()" and everything goes as expected, I will get rid of this new cluster parameter. Thanks!

Sgitario: I've just tested in K8s and Openshift and the above works. I've already removed this new cluster argument.

test/framework/kubernetes.go
```diff
@@ -285,3 +289,73 @@ func checkPodContainerHasEnvVariableWithValue(pod *corev1.Pod, containerName, en
...
+	ingress, err := WaitForIngress(namespace, serviceName)
+	if err != nil || ingress == nil || len(ingress.Spec.Rules) == 0 {
+		return "", fmt.Errorf("Ingress %s does not exist in namespace %s: %v", serviceName, namespace, err)
+	}
+
+	return fmt.Sprintf("http://%s:80", ingress.Spec.Rules[0].Host), nil
+}
+
+// WaitForIngress waits for ingress by a service
+func WaitForIngress(namespace, serviceName string) (*v1beta1.Ingress, error) {
```
sutaakar, line 304: This function doesn't have much sense from my point of view.
It could be merged together with GetIngressURI() above

test/framework/kubernetes.go
```diff
@@ -285,3 +289,73 @@ func checkPodContainerHasEnvVariableWithValue(pod *corev1.Pod, containerName, en
...
+	}
+}
+
+// ExposeServiceOnKubernetes adds ingress CR to expose a service
+func ExposeServiceOnKubernetes(namespace, serviceName string) error {
+	host := serviceName + namespace
+	if config.IsLocalTests() {
+		host = serviceName
+	}
+
```
sutaakar, line 322: Please add a possibility to specify hostname using configuration property.
Service name will be added as a hostname prefix.

test/config/config.go
```diff
@@ -182,6 +183,16 @@ func GetContainerEngine() string {
 	return env.containerEngine
 }
 
+// IsOpenshift returns whether the cluster is running on Openshift
+func IsOpenshift() bool {
+	return client.NewForConsole().IsOpenshift()
```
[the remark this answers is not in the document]
Sgitario: Will reuse the client in [here](https://github.com/kiegroup/kogito-cloud-operator/blob/aeca95dbaca9b9b14309d7b693a4911d29b4db01/test/framework/kubernetes.go#L33)

test/framework/http.go
```diff
@@ -43,6 +45,25 @@ const (
 	HTTPRequestResultError HTTPRequestResult = "error"
 )
 
+// WaitAndRetrieveRouteURI waits for a route and returns its URI
+func WaitAndRetrieveRouteURI(namespace, serviceName string) (string, error) {
```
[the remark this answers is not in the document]
Sgitario: Renamed and keep it in http.go script.

test/framework/operator.go
```diff
@@ -45,11 +43,16 @@ type dependentOperator struct {
...
+	source    string
+	namespace string
+}
+
 var (
 	kogitoOperatorPullImageSecretPrefix = operator.Name + "-dockercfg"
 
-	// KogitoOperatorCommunityDependencies contains list of community operators to be used together with Kogito operator
-	KogitoOperatorCommunityDependencies = map[string]dependentOperator{
+	// KogitoOperatorDependencies contains list of operators to be used together with Kogito operator
```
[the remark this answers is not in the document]
Sgitario: As the dependencies might change depending on the source, what about adding the field "dependencies" in the operatorCatalog type. Concretely, these two changes:
- change KogitoOperatorDependencies to an array of list of dependencies: Infinispan, Kafka and Keycloak where these names are internal for our tests
- add "dependencies" field to the operatorCatalog type, so we can define the details of the above dependency "Infinispan" per repository.

test/framework/kubernetes.go
Sgitario: Good idea. Merged.

test/framework/operator.go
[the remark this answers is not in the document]
radtriste: good for me :)

test/framework/kubernetes.go
Sgitario: I don't see why adding a new configuration property is needed here as this is for local purposes only as we need to update the "/etc/hosts" file. 
Without the "--local" param, the host will work as expected (appending the namespace into the service name).

Are you suggesting this change because you know that we won't cope with a specific use case running the tests with the "--local" param?

test/framework/kubernetes.go
```diff
@@ -285,3 +289,73 @@ func checkPodContainerHasEnvVariableWithValue(pod *corev1.Pod, containerName, en
...
+	} else if !exists {
+		return nil, nil
+	} else {
+		return ingress, nil
+	}
+}
+
+// ExposeServiceOnKubernetes adds ingress CR to expose a service
+func ExposeServiceOnKubernetes(namespace, serviceName string) error {
+	host := serviceName + namespace
```
Sgitario, line 318: I will be adding a new parameter to add a hostname suffix here. This new parameter is only for Kubernetes, as Openshfit is adding the hostname suffix automatically.

test/framework/kubernetes.go
Sgitario: After speaking with @sutaakar , his concern is that the "--local" parameter means that we'll be running the tests from local using either a local or remote cluster. Therefore, I'll be adding a new "--local-cluster" to state that the cluster is running locally.

hack/run-tests.sh
```diff
@@ -331,6 +337,10 @@ case $1 in
     shift
     if addParamKeyValueIfAccepted "--tests.dev.namespace-name" ${1}; then shift; fi
   ;;
+  --local_cluster)
+    addParam "--tests.local-cluster"
```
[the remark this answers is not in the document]
Sgitario: Sure, it makes sense.

Makefile
```diff
@@ -89,6 +89,7 @@ ci=
 cr_deployment_only=false
 load_default_config=false
 container_engine=
+hostname_suffix=
```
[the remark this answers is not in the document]
Sgitario: +1 with first option. I used to avoid acronyms.

hack/run-tests.sh
```diff
@@ -43,11 +43,12 @@ function usage(){
...
   printf "\n--load_factor {INT_VALUE}\n\tSet the tests load factor. Useful for the tests to take into account that the cluster can be overloaded, for example for the calculation of timeouts. Default value is 1."
-  printf "\n--local\n\tSpecify whether you run test in local."
+  printf "\n--local\n\tSpecify whether you run test in local using either a local or remote cluster."
   printf "\n--ci {CI_NAME}\n\tSpecify whether you run test with ci, give also the name of the CI."
   printf "\n--cr_deployment_only\n\tUse this option if you have no CLI to test against. It will use only direct CR deployments."
   printf "\n--load_default_config\n\tTo be used if you want to directly use the default test config contained into ${SCRIPT_DIR}/../test/.default_config"
   printf "\n--container_engine\n\tTo be used if you want to specify engine to interact with images and containers. Default is docker."
+  printf "\n--domain_suffix\n\tTo be used if you want to set a domain suffix for exposed services. Ignored when running tests on Openshift."
```
radtriste, line 51: ` Ignored when running tests on Openshift.`
Not sure we need to write it here. anyway that domain suffix is automatically into the route

test/framework/kubernetes.go
```diff
@@ -280,3 +284,69 @@ func checkPodContainerHasEnvVariableWithValue(pod *corev1.Pod, containerName, en
...
+	}
+
+	return fmt.Sprintf("http://%s:80", ingress.Spec.Rules[0].Host), nil
+}
+
+// ExposeServiceOnKubernetes adds ingress CR to expose a service
+func ExposeServiceOnKubernetes(namespace, serviceName string) error {
+	host := serviceName
+	if !config.IsLocalCluster() {
+		host += namespace + config.GetDomainSuffix()
```
radtriste, line 306: is there here any `.` to be added to setup the host correctly ?

hack/run-tests.sh
Sgitario: For Openshift, this is properly configured. That's this is not necessary to be done when running the tests in Openshift.

test/framework/kubernetes.go
```diff
@@ -280,3 +284,69 @@ func checkPodContainerHasEnvVariableWithValue(pod *corev1.Pod, containerName, en
...
+				{
+					Host: host,
+					IngressRuleValue: v1beta1.IngressRuleValue{
+						HTTP: &v1beta1.HTTPIngressRuleValue{
+							Paths: []v1beta1.HTTPIngressPath{
+								{
+									Path: "/",
+									Backend: v1beta1.IngressBackend{
+										ServiceName: serviceName,
+										ServicePort: intstr.FromInt(8080),
```
sutaakar, line 326: We should get existing port (retrieved from service?) instead of hardcoding as port will be configurable (https://github.com/kiegroup/kogito-cloud-operator/pull/378).

test/framework/kubernetes.go
```diff
@@ -280,3 +284,69 @@ func checkPodContainerHasEnvVariableWithValue(pod *corev1.Pod, containerName, en
...
+	}
+
+	return fmt.Sprintf("http://%s:80", ingress.Spec.Rules[0].Host), nil
+}
+
+// ExposeServiceOnKubernetes adds ingress CR to expose a service
+func ExposeServiceOnKubernetes(namespace, serviceName string) error {
+	host := serviceName
+	if !config.IsLocalCluster() {
+		host += fmt.Sprintf(".%s.%s", namespace, config.GetDomainSuffix())
```
sutaakar, line 306: We should make sure that `GetDomainSuffix()` is defined in this place, to provide meaningful error in case user forgets to define the domain suffix.

sutaakar:
Apart from comments above looks good to me.

test/framework/kubernetes.go
Sgitario: Good catch. I've added a verification stage before running the scenarios in order to check this kind of setup as I think it might be useful in other use cases. (See the new check_setup.go and my comments.

test/framework/kubernetes.go
Sgitario: Will do once this PR is merged.

Sgitario:
Depends on https://github.com/kiegroup/kogito-cloud-operator/pull/378

test/main_test.go
```diff
@@ -136,23 +136,29 @@ func configureTestOutput() {
...
 	// Create kube client
-	if err := framework.InitKubeClient(); err != nil{
+	if err := framework.InitKubeClient(); err != nil {
 		return err
 	}
 
+	// Verify Setup
+	if err := framework.CheckSetup(); err != nil {
+		return err
+	}
```
Sgitario, line 148: I've added this check setup stage in order to early fail the test.
This stage must be done right after the InitKubeClient is invoked.

test/framework/check_setup.go
```diff
@@ -0,0 +1,47 @@
...
+
+package framework
+
+import (
+	"fmt"
+	"github.com/kiegroup/kogito-cloud-operator/test/config"
+)
+
+var verifications = []checker{
+	checkKubernetesAndDomainSuffix(),
```
Sgitario, line 23: In order to add new checkers, it would be just a matter to add a new function here.

Sgitario:
@radtriste and @sutaakar I requested you to review the latest changes around the check_setup.go script. This script is to detect wrong configuration and early fail the test (I think we might want to add more verifications the most we support Kubernetes).

test/framework/check_setup.go
radtriste: couldn't we just give the reference to the method instead of calling a method returning a `func` ?
like we do in the steps

Sgitario: I thought adding the checker type function was clearer and this is used in other places within the Operator source. However, not a big deal. I've already updated to use the approaches in the steps.

radtriste:
@Sgitario good to be merged ? or still WIP ?

sutaakar:
Looks good though it makes me think...
What approach to take in case we need to check requirement for parameter which is used just in subset of tests? For example `runtime_application_image_registry` and `runtime_application_image_namespace`. These parameters are required just in case we run runtime tests. IMHO it doesn't have much sense to demand it for not related tests.

What do you think about creating a specific step checking parameter availability for this? Something like `Check that runtime registry is configured`?

Sgitario:
> Looks good though it makes me think...
> What approach to take in case we need to check requirement for parameter which is used just in subset of tests? For example `runtime_application_image_registry` and `runtime_application_image_namespace`. These parameters are required just in case we run runtime tests. IMHO it doesn't have much sense to demand it for not related tests.
> 
> What do you think about creating a specific step checking parameter availability for this? Something like `Check that runtime registry is configured`?

Indeed, I did this new script also thinking in the example you shared. But as it was kind of unrelated, I didn't change anything around that. However, I think it would be quite easy to extend this new script to check also requirements of the list of features to be executed. Anyhow, adding a new step "Check that runtime registry is configured" also works. 

@radtriste we're good to be merged.

sutaakar:
Right, it is unrelated to this PR.
I think that having a separate step for it would be easier to maintain as we won't need to keep a relation between requirements and features in a separate place.
Will create a JIRA for it.

ricardozanini:
This should be on 0.11.0 since it's meant to rest `KogitoRuntime`. I'm updating the milestones to have a clear view of the PRs. Please don't forget to update them (not sure if you have the permission) when sending a PR.

spolti:
@Sgitario  @ricardozanini does it really depends on https://github.com/kiegroup/kogito-cloud-operator/pull/378

I don't think we should merge that since that changes could bring new issues to 0.11 release, that's not safe I mean.

sutaakar:
The dependency on https://github.com/kiegroup/kogito-cloud-operator/pull/378 is related just for the port configuration for Ingress. IMHO it can be updated later.

ricardozanini:
> The dependency on #378 is related just for the port configuration for Ingress. IMHO it can be updated later.

Agreed

ricardozanini:
> @Sgitario @ricardozanini does it really depends on #378?
> 
> I don't think we should merge that since that changes could bring new issues to 0.11 release, that's not safe I mean.

Oh, I didn't said that it depends on #378! hahaha

ricardozanini:
@radtriste @sutaakar since this is a BDD PR related, I think you it's up to you to merge. I would.

Sgitario:
> The dependency on #378 is related just for the port configuration for Ingress. IMHO it can be updated later.

Fine with me. I can create a follow-up ticket for this. 
So, I guess @radtriste this is ok for being merged.

test/framework/kubernetes.go
Sgitario: Raised https://issues.redhat.com/browse/KOGITO-2466

PR #363, opened 2020-05-28
[KOGITO-2329] Implement image cache support for KogitoRuntime tests

sutaakar:

https://issues.redhat.com/browse/KOGITO-2329

### DIFF
--- a/test/features/deploy_kogito_runtime.feature
+++ b/test/features/deploy_kogito_runtime.feature
@@ -9,12 +9,11 @@ Feature: Deploy Kogito Runtime
   Scenario Outline: Deploy Simplest Scenario using Kogito Runtime
     Given Kogito Operator is deployed
     And Clone Kogito examples into local directory
-    And Local example service "<example-service>" is built by Maven using profile "<profile>"
-    And Local example service "<example-service>" is deployed to image registry, image tag stored as variable "built-image"
+    And Local example service "<example-service>" is built by Maven using profile "<profile>" and deployed to runtime registry
 
-    When Deploy <runtime> example service using image in variable "built-image" with configuration:
+    When Deploy <runtime> example service "<example-service>" from runtime registry with configuration:
       | config | persistence | disabled |
-    
+
     Then Kogito Runtime "<example-service>" has 1 pods running within 10 minutes
     And Service "<example-service>" with process name "orders" is available within 2 minutes
 
@@ -40,10 +39,9 @@ Feature: Deploy Kogito Runtime
   Scenario Outline: Deploy Scenario with persistence using Kogito Runtime
     Given Kogito Operator is deployed with Infinispan operator
     And Clone Kogito examples into local directory
-    And Local example service "<example-service>" is built by Maven using profile "<profile>"
-    And Local example service "<example-service>" is deployed to image registry, image tag stored as variable "built-image"
+    And Local example service "<example-service>" is built by Maven using profile "<profile>" and deployed to runtime registry
 
-    When Deploy <runtime> example service using image in variable "built-image" with configuration:
+    When Deploy <runtime> example service "<example-service>" from runtime registry with configuration:
       | config | persistence | enabled |
     And Kogito Runtime "<example-service>" has 1 pods running within 10 minutes
     And Start "orders" process on service "<example-service>" within 3 minutes with body:
@@ -87,10 +85,9 @@ Feature: Deploy Kogito Runtime
     Given Kogito Operator is deployed with Infinispan and Kafka operators
     And Install Kogito Data Index with 1 replicas
     And Clone Kogito examples into local directory
-    And Local example service "<example-service>" is built by Maven using profile "<profile>"
-    And Local example service "<example-service>" is deployed to image registry, image tag stored as variable "built-image"
+    And Local example service "<example-service>" is built by Maven using profile "<profile>" and deployed to runtime registry
 
-    When Deploy <runtime> example service using image in variable "built-image" with configuration:
+    When Deploy <runtime> example service "<example-service>" from runtime registry with configuration:
       | config | persistence | enabled |
       | config | events      | enabled  |
     And Kogito Runtime "<example-service>" has 1 pods running within 10 minutes
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,7 @@ local=false
 ci=
 cr_deployment_only=false
 load_default_config=false
+container_engine=
 # operator information
 operator_image=
 operator_tag=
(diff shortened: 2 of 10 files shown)


sutaakar:
If test framework discovers that registry contains previously built image then the image is used, image build is skipped.

test/features/deploy_kogito_runtime.feature
```diff
@@ -9,12 +9,11 @@ Feature: Deploy Kogito Runtime
   Scenario Outline: Deploy Simplest Scenario using Kogito Runtime
     Given Kogito Operator is deployed
     And Clone Kogito examples into local directory
-    And Local example service "<example-service>" is built by Maven using profile "<profile>"
-    And Local example service "<example-service>" is deployed to image registry, image tag stored as variable "built-image"
+    And Local example service "<example-service>" is built by Maven using profile "<profile>" and deployed to registry if not available, image tag stored as variable "built-image"
 
-    When Deploy <runtime> example service using image in variable "built-image" with configuration:
+    When Deploy <runtime> example service "<example-service>" using image in variable "built-image" with configuration:
```
Sgitario, line 14: I would rewrite these steps to: 
- "Local example service "<example-service>" is built by Maven using profile "<profile>" and deployed to registry"
Where internally, the variable is called "built-image-<example-service>"
- "Deploy <runtime> example service "<example-service>" from registry with configuration:"
Where from registry will search the above variable transparently

sutaakar: Problem is that to resolve image name you have to know maven profiles used to build the image.
And I would like to avoid passing maven profiles to deploy step as it is irrelevant there

Sgitario: I meant to keep using the variable with the resolved image name, but without needing to pass the variable name, but having like a contract between these two steps.

sutaakar: So you mean to store built image to property named like "<service name>-runtime-image" where service name will be a key for resolving image name?

From my perspective:
- pros - Simpler step syntax
- cons - Hiding of dependency between steps (casual reader won't realize that these steps are connected and how)
           - "Artificial dependency" - deployed service has to have exactly same name as built service folder name
           - A bit more logic in the step implementation (more code to take care of :) )

I would incline to current solution, though I would be ok with either of those.
@radtriste What do you think?

radtriste: Thinking that we should always be explicit for readability to any user.
The writer of the feature may be not correlated to the writer of the implementation and the implementation could be changed later, forgetting this dependency.
Having it explicitly defined in the feature helps to understand what is happening really in the scenario, without having to dig in the code.
I think we should be explicit whenever we can.

Sgitario: I disagree that we're hiding dependencies between steps:
- As far I can see it, saying "to registry" and "from registry" is enough and understandable for readers. What I mean is that saying "to registry with variable "X"" and "from registry using variable "X"", the variable "X" is a bit verbose for me as this should be something internal for the registry part and the service to be deployed in the current scenario.

- Also, about the logic in the step implementation, take into account that the variables are stored in the scenario context, so we don't need to deal with parallel execution. Therefore, it's a matter to use the variable name "builtImage" or to use the variable name "<service name>". Don't see this overcomplicate the logic here.

sutaakar: That can be fine for one image being used in one scenario.
But what about having for example 3 deployments in one scenario (viz service discovery)?:

Local example service "onboarding-example/hr" ... and deployed to registry"
Local example service "onboarding-example/payroll" ... and deployed to registry"
Local example service "onboarding-example/onboarding-quarkus" ... and deployed to registry"

Deploy example service "hr" from registry...
Deploy example service "payroll" from registry...
Deploy example service "onboarding-quarkus" from registry...

Sgitario: There would be three variables, one for each service. data.ScenarioContext[imagePrefix + serviceName].

sutaakar: yes, that means that reader needs to realize "built folder name" == "deployed service name"

Sgitario: Why? the reader only knows that the service is in the registry, and that it will used afterwards.

sutaakar: And how does reader know that for example "hr" service will be in the registry and for example "management" service doesn't exist there?

Sgitario: Do you concern is that the word "registry" is too open? Maybe it is. What about to say "runtime registry" instead of just "registry"? 

```
Local example service "onboarding-example/hr" ... and deployed to runtime registry"
Deploy example service "hr" from runtime registry...
```

This way it does also reference to the runtime registry arguments as well.

ricardozanini:
The CI is failing. We should hold until it passes.

sutaakar:
CI rebuild passed.
Lets wait for clarification with @Sgitario before merging.

test/features/deploy_kogito_runtime.feature
radtriste: We are already using some internal variables, for example for the local checkout of project and build, and so some steps are already dependent on each other.
I would go for the internal var instead of "in step" description. We habe been using the `service name` as a key for most of our steps until now and we could continue with it.
Moreover, we have a "default" behavior if the step `Local example service ...` is not existing: we would be generating the image's name with our code anyway, so it would be easy to retrieve from `runtime registry` (I would add the `runtime` like suggested by @Sgitario to be more explicit where we take the image from and that this is happening).

Finally, I think we should start to use internal var if we have different images for the same service. If we have only one image per service, it is easy. But I don't think we need this, at least for now ...

sutaakar:
@Sgitario @radtriste Please take a look again, I have adjusted the PR according to your comments.

ricardozanini:
> CI rebuild passed.
> Lets wait for clarification with @Sgitario before merging.

I'm still seeing failures. I think we might need to rethink this CI check, maybe move to kind and review CircleCI steps.

sutaakar:
Hmm, this time the failure seems infrastructure related:
`Transfer failed for https://repository.jboss.org/nexus/content/groups/public/org/kie/kogito/kogito-springboot-starter/8.0.0-SNAPSHOT/kogito-springboot-starter-8.0.0-SNAPSHOT.pom: Unknown host repository.jboss.org `
Looks like some instability, in this case the KIND wouldn't help.

Will retrigger the build.

ricardozanini:
> Looks like some instability, in this case the KIND wouldn't help.
Perhaps, maybe the problem was on Jenkins accessing the url. :)

sutaakar:
CI is finally green :)

PR #368, opened 2020-06-02
[KOGITO-1999] - Introducing KogitoBuild

ricardozanini:
### Overview

See: https://issues.redhat.com/browse/KOGITO-1999

In this PR we introduce a new CRD, which enables users to manage their Kogito service builds independently, choosing the right build for their needs.

For now, there's no integration with CLI, this will be done on https://issues.redhat.com/browse/KOGITO-1998.

Design notes: https://docs.google.com/presentation/d/1scHt4ccU7r67K0TrU6leWIXcv6dOscRiEEvOCmg6jb0/

### How to test

There's an example `process-quarkus-example-runtime-build.yaml` in the `deploy\examples` directory that shows how this integration will happen. An user can create a KogitoBuild CR according to their needs (either Local/RemoteSource or Binary), start the build and then deploy a `KogitoRuntime` with the same name as the `KogitoBuild`. They will integrate with each other through a shared ImageStream, owned by both instances. 

If a developer wishes to also add other kind of build, it is also possible. But in this case the attribute `.spec.targetKogitoRuntime` must be filled with the `KogitoRuntime` name.

Enjoy!

### Dependencies

Depends on bumping to 0.12 to fix the CRD generation.


sutaakar:
Looks good to me.
I will try some experiments with this on OCP/Kubernetes.

sutaakar:
It seems that KogitoBuild status is not being refreshed properly.
It shows:
```
status:
  builds:
    pending:
      - example-quarkus-builder-2
```
even when the build is finished.

ricardozanini:
> It seems that KogitoBuild status is not being refreshed properly.
> It shows:
> 
> ```
> status:
>   builds:
>     pending:
>       - example-quarkus-builder-2
> ```
> 
> even when the build is finished.
> 
> Edit: so it seems that the status is updated, but after very long time (xy minutes?).

Not sure about that. The status should be updated on every reconciliation cycle. Since the `Build` object is not controlled by the operator, the cycle won't be happening any time soon, only after 10~15 minutes, that's the default reconciliation interval.

The difference between the `KogitoApp` is that in this CR we won't have too many reconciliation cycles since there's no resources being updated.

What we can do in the future to improve it, is to have a go routine watching `Builds` or even watch for `Builds` objects. It's too edge case IMO, just for a status update. But I'm opened for suggestions.

ricardozanini:
@radtriste @msmagnanijr @sutaakar you guys know why the CI is failing?

```
error finding local image "quay.io/kiegroup/kogito-cloud-operator:0.11.0": error locating image with name "quay.io/kiegroup/kogito-cloud-operator:0.11.0" ([quay.io/kiegroup/kogito-cloud-operator:0.11.0]): image not known

script returned exit code 1
```
I understand that the 0.11 image version is not there yet, but what did I do to provoke this error?

sutaakar:
@ricardozanini It seems to be an issue with operator image build:
`error creating build container: Error committing the finished image: error adding layer with blob "sha256:9a481924bf76e6135dc9442a303ad4acb01a11fc7d4c209deec6ac51d4d012b3": Error processing tar file(io: read/write on closed pipe): `

Possibly a CI instability.

sutaakar:
@ricardozanini About status update - fine for me to keep it like that. I was just surprised by it.

ricardozanini:
> @ricardozanini About status update - fine for me to keep it like that. I was just surprised by it.

Your concern is super valid, IMO. But I intentionally decreased the number of reconciliation cycles to avoid long lines of logging / resources consumption. If you feel like this is an important feature, please open a JIRA for me and I can work on it later :D

pkg/controller/kogitobuild/build/trigger.go
```diff
@@ -0,0 +1,100 @@
...
+						return false, err
+					}
+					return true, err
+				})
+				if err != nil {
+					log.Errorf("Failed to cancel build %s: %s", build.Name, err)
+					return
+				}
+				// wait for the build to be cancelled
+				err = wait.Poll(500*time.Millisecond, cancelUpdateTimeOut, func() (bool, error) {
```
spolti, line 80: maybe move 500 to a const?

pkg/controller/kogitobuild/kogitobuild_controller_test.go
```diff
@@ -0,0 +1,175 @@
...
+	return
+}
+
+func assertReconcileMustRequeue(t *testing.T, r ReconcileKogitoBuild, instance *v1alpha1.KogitoBuild) (result reconcile.Result) {
+	result = assertReconcile(t, r, instance)
+	assert.True(t, result.Requeue)
+	return
+}
+
+func assertReconcileMustNotRequeue(t *testing.T, r ReconcileKogitoBuild, instance *v1alpha1.KogitoBuild) (result reconcile.Result) {
```
spolti, line 56: maybe move it to a helper class/pkg so it can be reused in more test files?

pkg/controller/kogitoruntime/kogitoruntime_controller.go
```diff
@@ -105,6 +105,7 @@ func (r *ReconcileKogitoRuntime) Reconcile(request reconcile.Request) (reconcile
 		OnDeploymentCreate: onDeploymentCreate,
 		OnObjectsCreate:    onObjectsCreate,
 		OnGetComparators:   onGetComparators,
+		CanBeBuilt:         true,
```
spolti, line 108: is this correct? I mean, shouldn't it come from parameter or something else?

deploy/crds/app.kiegroup.org_v1alpha1_kogitobuild_cr.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: app.kiegroup.org/v1alpha1
```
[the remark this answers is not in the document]
ricardozanini: No, this file is used by the CSV to fill the examples in the OLM console.

deploy/examples/process-quarkus-example-runtime-build.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  buildImage:
+    tag: "0.10"
+  gitSource:
+    contextDir: process-quarkus-example
+    uri: https://github.com/kiegroup/kogito-examples
+  runtime: quarkus
+  runtimeImage:
+    tag: "0.10"
+  type: RemoteSource
```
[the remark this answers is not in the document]
ricardozanini: We can't guess which kind of build the user will create. If it's left in blank and there's a Git URL, we assume `RemoteSource`. But local and binary builds have the same attributes.

pkg/apis/app/v1alpha1/types.go
```diff
@@ -26,6 +28,22 @@ type Image struct {
 	Tag       string `json:"tag,omitempty"`
 }
 
+// IsEmpty verifies if this Image instance is empty
+func (i *Image) IsEmpty() bool {
```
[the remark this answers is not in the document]
ricardozanini: LOL, nice catch! Will fix it.

pkg/infrastructure/services/resources.go
```diff
@@ -156,8 +181,14 @@ func (s *serviceDeployer) getKogitoServiceImage(imageHandler *imageHandler, inst
 	if len(image) > 0 {
 		return image, nil
 	}
-	log.Warnf("Image for the service %s not found yet in the namespace %s. Please make sure that the informed image %s exists in the given registry.",
-		instance.GetName(), instance.GetNamespace(), imageHandler.resolveRegistryImage())
+	if instance.GetSpec().HasImage() {
```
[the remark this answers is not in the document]
ricardozanini: `IsEmpty` is more suitable I think. I'll rephrase it.

pkg/controller/kogitobuild/build/trigger.go
ricardozanini: Maybe :)

pkg/controller/kogitobuild/kogitobuild_controller_test.go
ricardozanini: This is specific to this test suit. If others would also need, we can then think on generalize. I don't like to create generic functions upfront, is not a good practice.

pkg/controller/kogitoruntime/kogitoruntime_controller.go
ricardozanini: It's coming from the parameter. The Service Deployer API is highly customizable and for this use case, we need to know if the service is a "custom" one or not. Do you have any other suggestion for this name? It's not a parameter that a user would inform.

ricardozanini:
> Nice work :)
> just a few things, more questions though.

Thank you! Will work on the suggestions tomorrow morning. Too tired right now :sleeping:

sutaakar:
@ricardozanini There seems to be up to 30 second delay between creation of KogitoBuild and creation of BuildConfig. Would it be possible to reduce the delay?

ricardozanini:
> @ricardozanini There seems to be up to 30 second delay between creation of KogitoBuild and creation of BuildConfig. Would it be possible to reduce the delay?

Sure! What about 10s?

sutaakar:
> Sure! What about 10s?

Much better :)

deploy/examples/process-quarkus-example-runtime-build.yaml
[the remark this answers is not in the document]
ricardozanini: I forgot to add: the type also differentiates the `BuildConfigs`, and it's not an arbitrary field, but an `enum` being `RemoteSource`, `LocalSource` and `Binary`. Kubernetes Open API validates the CR when you create a KogitoBuild instance.

ricardozanini:
@sutaakar @spolti I believe I addressed your comments, mind taking a second look? @sutaakar I hope your tests are going well :pray: 
This is a huge change. I'd like to ask an extra care on the "cancel build" feature, which is the way we handle simultaneous changes: if you change the CR (like adding an env, or the git source), the operator would cancel any running builds before starting a new one.

pkg/controller/kogitoruntime/kogitoruntime_controller.go
ricardozanini: Removed this parameter and renamed the builder.

pkg/controller/kogitobuild/kogitobuild_controller.go
```diff
@@ -0,0 +1,218 @@
...
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a KogitoBuild object and makes changes based on the state read
+// and what is in the KogitoBuild.Spec
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileKogitoBuild) Reconcile(request reconcile.Request) (result reconcile.Result, resultErr error) {
+	log.Infof("Reconciling KogitoRuntime for %s in %s", request.Name, request.Namespace)
```
sutaakar, line 105:
```suggestion
	log.Infof("Reconciling KogitoBuild for %s in %s", request.Name, request.Namespace)
```

pkg/controller/kogitobuild/kogitobuild_controller.go
```diff
@@ -0,0 +1,218 @@
...
+	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	"time"
+)
+
+var log = logger.GetLogger("kogitoruntime_controller")
```
sutaakar, line 39:
```suggestion
var log = logger.GetLogger("kogitobuild_controller")
```

sutaakar:
@ricardozanini If KogitoRuntime is created sooner than KogitoBuild finish the build, then application is not deployed (KogitoRuntime is not reconciled once image is available).
This issue seems to be related to the fact that image stream has two owners and controller is watching just on first owner, see https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/framework/controller_watcher.go#L123 and parameter IsController.

Should it be adjusted to watch for all owners (IsController: false)?

sutaakar:
When I define custom runtime image like this:
```
kind: KogitoBuild
metadata:
  name: process-springboot-example
spec:
  type: RemoteSource
  gitSource:
    uri: https://github.com/kiegroup/kogito-examples
    contextDir: process-springboot-example
  runtimeImage:
    tag: 0.10.0
    name: experiment
  buildImage:
    tag: 0.10.0
  runtime: springboot
```

Then operator creates `custom-experiment` image stream (expected), but this image stream contains bad docker image reference:
```
      from:
        kind: DockerImage
        name: 'quay.io/kiegroup/custom-experiment:0.10.0'
```

I would expect the reference to be 'quay.io/kiegroup/experiment:0.10.0'

ricardozanini:
> @ricardozanini If KogitoRuntime is created sooner than KogitoBuild finish the build, then application is not deployed (KogitoRuntime is not reconciled once image is available).
> This issue seems to be related to the fact that image stream has two owners and controller is watching just on first owner, see https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/framework/controller_watcher.go#L123 and parameter IsController.
> 
> Should it be adjusted to watch for all owners (IsController: false)?

hmm, I think I need to take a look on the Triggers for the Deployment ImageStream annotation. The deployment will happen once the reconciliation loop starts. The controller is not watching because a resource can only have one `Controller`, that's why both are `false`. So either I change the triggers or give the control to the `KogitoRuntime` instead. I'll take a look, thanks for reporting it!

ricardozanini:
> When I define custom runtime image like this:
> 
> ```
> apiVersion: app.kiegroup.org/v1alpha1
> kind: KogitoBuild
> metadata:
>   name: process-springboot-example
> spec:
>   type: RemoteSource
>   gitSource:
>     uri: https://github.com/kiegroup/kogito-examples
>     contextDir: process-springboot-example
>   runtimeImage:
>     tag: 0.10.0
>     name: experiment
>   buildImage:
>     tag: 0.10.0
>   runtime: springboot
> ```
> 
> Then operator creates `custom-experiment` image stream (expected), but this image stream contains bad docker image reference:
> 
> ```
>       from:
>         kind: DockerImage
>         name: 'quay.io/kiegroup/custom-experiment:0.10.0'
> ```
> 
> I would expect the reference to be 'quay.io/kiegroup/experiment:0.10.0'

We need to add a prefix, otherwise if you set the name of the image the same we have on our namespace, the cluster will pull the wrong image.

pkg/apis/app/v1alpha1/kogitobuild_types.go
```diff
@@ -0,0 +1,236 @@
...
+	// Set this field targeting the desired KogitoRuntime when this KogitoBuild instance has a different name than the KogitoRuntime.
+	// By default this KogitoBuild instance will generate a final image named after its own name (.metadata.name).
+	// On OpenShift, an ImageStream will be created causing a redeployment on any KogitoRuntime with the same name.
+	// On Kubernetes, the final image will be pushed to the KogitoRuntime deployment.
+	// If you have multiple KogitoBuild instances (let's say BinaryBuildType and Remote Source), you might need that both target the same KogitoRuntime.
+	// Both KogitoBuilds will update the same ImageStream or generate a final image to the same KogitoRuntime deployment.
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Base Image"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +optional
+	TargetKogitoRuntime string `json:"targetKogitoRuntime,omitempty"`
```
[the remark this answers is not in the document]
ricardozanini: We can't add image streams to the `KogitoRuntime` since this is an OpenShift object, and we are all about supporting core Kubernetes API.

I got your point, we would have an array of possible images in the Runtime interface. From an UX perspective this seems awkward, a service having more than one image. Having the build targeting a service seems more natural IMO. I won't change that now, I'd rather see how this works for the community and hear more feedback from it out there. Think that the operator builds can be done on a separate namespace/environment, having a CI to promote the generated images. Having lot's of builds (building on different time frames) targeting the same image, would make things easier. Or, if that won't work for the user, they could just create a new build with no target runtime and promote the image retagging it. A list of images in the runtime is a no go to me at this moment.

> Because doing that way, I can update any runtime by mistake and with something that could completely unrelated ...

I don't follow. wdym?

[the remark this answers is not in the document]
radtriste: that can also be done the other way. I was myself a bit confused I think ...
Ok to let it like that and see the feedbacks

ricardozanini:
@sutaakar I believe I addressed your concerns, can you take another look?

I'll fix the coverage problem Monday morning.

sutaakar:
@ricardozanini Sure, on monday morning :)

sutaakar:
@ricardozanini Custom image tag is still broken. If I create:
```
kind: KogitoBuild
metadata:
  name: process-springboot-example
spec:
  buildImage:
    name: kogito-springboot-ubi8-nightly
    tag: latest
  gitSource:
    contextDir: process-springboot-example
    uri: 'https://github.com/kiegroup/kogito-examples'
  runtime: springboot
  runtimeImage:
    name: kogito-springboot-ubi8-s2i-nightly
    tag: latest
  type: RemoteSource
```

Then image stream tag is correctly created, however BuildConfig points to "plain" image stream tag name, without custom prefix - so it cannot be resolved.
Current process-springboot-example-builder BC:
```
    sourceStrategy:
      from:
        kind: ImageStreamTag
        namespace: ksuta-local-cucumber-6c63
        name: 'kogito-springboot-ubi8-nightly:latest'
```

Expected:
```
    sourceStrategy:
      from:
        kind: ImageStreamTag
        namespace: ksuta-local-cucumber-6c63
        name: 'custom-kogito-springboot-ubi8-nightly:latest'
```
ImageChange trigger seems to be also wrong.

sutaakar:
The KogitoRuntime reconciliation seems fixed.

pkg/controller/kogitobuild/build/image_stream_kogito.go
```diff
@@ -0,0 +1,280 @@
...
+
+// resolveKogitoImageName resolves the ImageName to be used in the given build, e.g. kogito-quarkus-ubi8-s2i
+func resolveKogitoImageStreamName(build *v1alpha1.KogitoBuild, isBuilder bool) string {
+	imageName := resolveKogitoImageName(build, isBuilder)
+	image := &build.Spec.RuntimeImage
+	if isBuilder {
+		image = &build.Spec.BuildImage
+	}
+	if len(image.Name) > 0 { // custom image
+		return strings.Join([]string{customKogitoImagePrefix, imageName}, "")
```
spolti, line 161: what is the purpose to add custom prefix here?:

ricardozanini:
> @ricardozanini Custom image tag is still broken. If I create:

Sorry about that, @sutaakar, I'm taking a look at it, it seems that I forget to change one line of code in the previous fix. Deeply sorry. I'm sending the fix and also the increase for the coverage report.

pkg/controller/kogitobuild/build/image_stream_kogito.go
ricardozanini: See the example @sutaakar gave: when an user creates their own s2i images, we will create custom ImageStreams to handle those instead of using our own. We need a new name in case the user sets their image name being the same of ours.

ricardozanini:
@sutaakar take a look:

```
  strategy:
    sourceStrategy:
      from:
        kind: ImageStreamTag
        name: custom-kogito-springboot-ubi8-s2i-nightly:latest
        namespace: kogito-1999
    type: Source
  triggers:
  - imageChange:
      from:
        kind: ImageStreamTag
        name: process-springboot-example-builder:latest
      lastTriggeredImageID: image-registry.openshift-image-registry.svc:5000/kogito-1999/process-springboot-example-builder@sha256:d9905bc54c6ea2ff2b98d651c5fd6f3f6cfdfff773f43ef068fef483b616e82f
    type: ImageChange
```

I used the same YAML file, so the images are switched, but I believe it's corrected by now. The build worked in my end.

pkg/apis/app/v1alpha1/types.go
```diff
@@ -26,6 +28,22 @@ type Image struct {
...
+	return len(i.Namespace) == 0 &&
+		len(i.Domain) == 0 &&
+		len(i.Name) == 0 &&
+		len(i.Tag) == 0
```
Sgitario, line 36: Maybe checking the Tag is too restricted? If not provided, tag should be "latest".

sutaakar:
@ricardozanini Works as expected, thanks

pkg/apis/app/v1alpha1/kogitobuild_types.go
```diff
@@ -0,0 +1,236 @@
...
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:resourceRequirements"
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Maven Mirror URL to be used during source-to-image builds (Local and Remote) to considerably increase build speed
+	// +optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Maven Mirror URL"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:label"
+	MavenMirrorURL string `json:"mavenMirrorURL,omitempty"`
```
Sgitario, line 100: I don't think this setting copes with private Maven repositories, are we planning to support using private repositories in the future? It would be nice to create a ticket for this feature.

pkg/apis/app/v1alpha1/kogitobuild_types.go
```diff
@@ -0,0 +1,236 @@
...
+	// +listType=atomic
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Build Env Variables"
+	// +optional
+	Envs []corev1.EnvVar `json:"envs,omitempty"`
+
+	// Information about the git repository where the Kogito Service source code resides.
+	// Ignored for binary builds
+	// +optional
+	GitSource GitSource `json:"gitSource,omitempty"`
```
Sgitario, line 63: Does GitSource support private GIT repositories? In this case, we would need to provide the SSH credentials. If not, we should raise a ticket for this.

test/framework/util.go
```diff
@@ -39,7 +40,7 @@ const (
 // GenerateNamespaceName generates a namespace name, taking configuration into account (local or not)
 func GenerateNamespaceName(prefix string) string {
 	rand.Seed(time.Now().UnixNano())
-	ns := fmt.Sprintf("%s-%s", prefix, RandSeq(4))
+	ns := fmt.Sprintf("%s-%s", prefix, test.GenerateShortUID(4))
```
sutaakar, line 43: This change breaks the Cucumber tests when running in parallel.
In case of parallel tests there is a chance that tests will start in exactly same time (happens in the beginning of test execution). In such case same id is provided resulting in same namespace name, leading to undesired behaviour.

Can you please return RandSeq back for Cucumber tests?

pkg/apis/app/v1alpha1/types.go
ricardozanini: We need this since any parameter is filled, the image would be a custom one.

ricardozanini:
> Good work! I just spotted some minor things, however I'm missing a section in the README.md about how to run this using CRD. But not sure if you were planning to add this when adding the support of this via CLI.

This is a "secret" feature now. I should document on [KOGITO-1998](https://issues.redhat.com/browse/KOGITO-2064) and [KOGITO-2064](https://issues.redhat.com/browse/KOGITO-2064).

Thanks for the heads up!

pkg/apis/app/v1alpha1/kogitobuild_types.go
ricardozanini: Why not? You would set the public Maven URL to your private "group" repository, which groups all your enterprise repos. Not sure if I understood your question. Maybe elaborate a little bit more in format of story and open a JIRA?

pkg/apis/app/v1alpha1/kogitobuild_types.go
ricardozanini: Yes, any kind of git repos. Well put, we can definitely add this support if BuildConfig supports it. Could you please open a JIRA for me?

test/framework/util.go
ricardozanini: Of course! But I'm wondering why the UID is generated with the same characters? Could you please explain a little bit more (I'll ping you)?

pkg/apis/app/v1alpha1/kogitobuild_types.go
Sgitario: Using git@github.com:kiegroup/kogito-examples.git is not working, it fails with:

{code}
Error: stat git@github.com:Sgitario/kogito-examples.git: no such file or directory
{code}

But without SSH, it works fine setting the username and password like this: https://username:password@github.com/username/repo_name.git.

Should I raise an issue to support SSH?

ricardozanini:
>  In case of parallel tests there is a chance that tests will start in exactly same time (happens in the beginning of test execution). In such case same id is provided resulting in same namespace name, leading to undesired behaviour.

@sutaakar fixed by using Random instead. Could you please give it a try? We do have a chance of conflict on 1 in 17 billion though. :smile:

sutaakar:
Works as expected, thanks :)

pkg/apis/app/v1alpha1/kogitobuild_types.go
ricardozanini: Please, raise and I'll take a look. Even if it's just an update in the docs to use the format with `username:password@`.

ricardozanini:
@spolti can you please take a final look before merging it?

PR #366, opened 2020-06-02
[KOGITO-2330] Unify parsing of configuration table for Kogito services

sutaakar:
Signed-off-by: Karel Suta <ksuta@redhat.com>

https://issues.redhat.com/browse/KOGITO-2330

Many thanks for submiting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.MD#sending-a-pull-request)
- [x] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket
- [ ] Your feature/bug fix has a unit test that verifies it
- [x] You've tested the new feature/bug fix in an actual OpenShift cluster

PR #370, opened 2020-06-02
[KOGITO-2358] Add a scenario for optaplanner-quarkus on OpenShift

rsynek:
https://issues.redhat.com/browse/KOGITO-2358

Adds a new scenario that deploys the optaplanner-quarkus example to OpenShift via the operator and checks the application is responding. 

Motivation: we have similar scenarios for a couple of other examples, but until now any automated end-to-end check for OptaPlanner has been missing.

Sorry for the extra changes due to formatting (automatic removal of trailing spaces done by IDE).

### DIFF
--- a/test/features/deploy_kogito_runtime.feature
+++ b/test/features/deploy_kogito_runtime.feature
@@ -14,7 +14,7 @@ Feature: Deploy Kogito Runtime
 
     When Deploy <runtime> example service using image in variable "built-image" with configuration:
       | config | persistence | disabled |
-    
+
     Then Kogito Runtime "<example-service>" has 1 pods running within 10 minutes
     And Service "<example-service>" with process name "orders" is available within 2 minutes
 
@@ -49,14 +49,14 @@ Feature: Deploy Kogito Runtime
     And Start "orders" process on service "<example-service>" within 3 minutes with body:
       """json
       {
-        "approver" : "john", 
+        "approver" : "john",
         "order" : {
-          "orderNumber" : "12345", 
+          "orderNumber" : "12345",
           "shipped" : false
         }
       }
       """
-    
+
     Then Service "<example-service>" contains 1 instances of process with name "orders"
 
     When Scale Kogito Runtime "<example-service>" to 0 pods within 2 minutes
@@ -97,14 +97,14 @@ Feature: Deploy Kogito Runtime
     And Start "orders" process on service "<example-service>" within 3 minutes with body:
       """json
       {
-        "approver" : "john", 
+        "approver" : "john",
         "order" : {
-          "orderNumber" : "12345", 
+          "orderNumber" : "12345",
           "shipped" : false
         }
       }
       """
-    
+
     Then GraphQL request on Data Index service returns ProcessInstances processName "orders" within 2 minutes
 
     @springboot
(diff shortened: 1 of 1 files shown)


test/features/deploy_quarkus_service.feature
```diff
@@ -51,6 +51,31 @@ Feature: Deploy quarkus service
...
+          "origin" : "A",
+          "destination" : "B",
+          "departureDateTime" : "2020-05-30T17:30:43.873968",
+          "seatRowSize" : 6,
+          "seatColumnSize" : 10
+        }
+      }
+      """
+
+    @native
```
radtriste, line 74: missing non-native examples here

test/features/deploy_quarkus_service.feature
```diff
@@ -51,6 +51,31 @@ Feature: Deploy quarkus service
       | native  | minutes |
       | enabled | 20      |
 
+#####
+
+  Scenario Outline: Deploy process-optaplanner-quarkus service without persistence and native <native>
```
radtriste, line 56: are you planning springboot in another PR ?

radtriste:
Also, we usually use KOGITO issues for the changes. Can you create one ?

test/features/deploy_quarkus_service.feature
rsynek: I was thinking of doing just a single combination in order not to prolong the execution time too much. Is it required to include the non-native variant? If so, I will add it, no problem.

test/features/deploy_quarkus_service.feature
rsynek: I am not - for the very same reason I stated in the previous comment.

> If we are going to test all examples, I think we should create a new testing subset of features that should be executed on demand. Or should we test them all in full testing ?
> @sutaakar @Sgitario @MarianMacik what do you think ?

We test examples that incorporate processes and rules, but the OptaPlanner is completely missing here right now.

After we talked about testing all examples and you guys didn't recommend it, I decided to add only a single test - to be sure that OptaPlanner is tested in combination with Quarkus and OptaPlanner. I don't plan to add anything else.

rsynek:
@radtriste Jira reference updated, non-native variant added.

test/features/deploy_quarkus_service.feature
```diff
@@ -41,7 +41,37 @@ Feature: Deploy quarkus service
...
+          "origin" : "A",
+          "destination" : "B",
+          "departureDateTime" : "2020-05-30T17:30:43.873968",
+          "seatRowSize" : 6,
+          "seatColumnSize" : 10
+        }
+      }
+      """
+
+    @smoke
```
radtriste, line 74: not sure we should have that one as a smoke test. I would remove this tag.

radtriste:
> @radtriste Jira reference updated, non-native variant added.

thank you

rsynek:
@radtriste `smoke` tag removed, that should be it. Thanks for your review!

sutaakar:
I don't think we should test all examples, examples could be tested in example repository. Here we should rather focus on integration between components.

rsynek:
> I don't think we should test all examples, examples could be tested in example repository. Here we should rather focus on integration between components.

I've already answered in a different, but related comment, but unfortunately it got folded. Let me copy the answer here just to make it more visible to everybody:

We test examples that incorporate processes and rules, but the OptaPlanner is completely missing here right now.

After we talked about testing all examples and you guys didn't recommend it, I decided to add only a single test - to be sure that OptaPlanner is tested in combination with Quarkus and OptaPlanner. I don't plan to add anything else.

sutaakar:
@rsynek That is ok for me. We can have a sample test for every "capability" we have, I just want to avoid having hundreds of tests which mostly just duplicate another tests.

A side note, maybe it would be better to move/transform the test to https://github.com/kiegroup/kogito-cloud-operator/blob/master/test/features/deploy_kogito_runtime.feature . The point is that KogitoApp will become deprecated as building and running Kogito application will get separated. 
Also the KogitoRuntime tests have implemented image cache.

rsynek:
@sutaakar ok, I will move the test later today.

sutaakar:
@rsynek When you will test the moved test please define also `runtime_application_image_registry` and `runtime_application_image_namespace` in your command. It serves as storage for built images. You can use your quay account for it if you have some (in that case you need to make sure that pushed images are public).
Also comment out `@disabled` tag in the beginning of feature file when running tests.

rsynek:
@sutaakar reworked to deploy_kogito_runtime.feature. Please review it again.

sutaakar:
Needs to be coordinated with https://github.com/kiegroup/kogito-cloud-operator/pull/366 as it changes the step syntax and configuration options.
Depends on what will merge first :)

rsynek:
Thanks for your help & guidance guys!

PR #373, opened 2020-06-04
[KOGITO-2278] Add option to mark Kogito image streams as insecure

sutaakar:
https://issues.redhat.com/browse/KOGITO-2278

### DIFF
--- a/cmd/kogito/command/install/data_index.go
+++ b/cmd/kogito/command/install/data_index.go
@@ -16,6 +16,7 @@ package install
 
 import (
 	"fmt"
+
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/common"
 
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
@@ -39,7 +40,6 @@ const (
 
 type installDataIndexFlags struct {
 	deploy.CommonFlags
-	image              string
 	kafka              v1alpha1.KafkaConnectionProperties
 	infinispan         v1alpha1.InfinispanConnectionProperties
 	infinispanSasl     string
@@ -124,7 +124,7 @@ For more information on Kogito Data Index Service see: https://github.com/kiegro
 			if err := deploy.CheckDeployArgs(&i.flags.CommonFlags); err != nil {
 				return err
 			}
-			if err := deploy.CheckImageTag(i.flags.image); err != nil {
+			if err := deploy.CheckImageTag(i.flags.Image); err != nil {
 				return err
 			}
 			return nil
@@ -143,7 +143,6 @@ func (i *installDataIndexCommand) InitHook() {
 	i.Parent.AddCommand(i.command)
 	deploy.AddDeployFlags(i.command, &i.flags.CommonFlags)
 
-	i.command.Flags().StringVarP(&i.flags.image, "image", "i", "", "Image tag for the Data Index Service, example: quay.io/kiegroup/kogito-data-index:latest")
 	i.command.Flags().StringVar(&i.flags.kafka.ExternalURI, "kafka-url", "", "The Kafka cluster external URI, example: my-kafka-cluster:9092")
 	i.command.Flags().StringVar(&i.flags.kafka.Instance, "kafka-instance", "", "The Kafka cluster external URI, example: my-kafka-cluster")
 	i.command.Flags().StringVar(&i.flags.infinispan.URI, "infinispan-url", "", "The Infinispan Server URI, example: infinispan-server:11222")
@@ -195,12 +194,13 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error 
 			KogitoServiceSpec: v1alpha1.KogitoServiceSpec{
 				Replicas: &i.flags.Replicas,
 				Envs:     shared.FromStringArrayToEnvs(i.flags.Env),
-				Image:    framework.ConvertImageTagToImage(i.flags.image),
+				Image:    framework.ConvertImageTagToImage(i.flags.Image),
 				Resources: v1.ResourceRequirements{
 					Limits:   shared.FromStringArrayToResources(i.flags.Limits),
 					Requests: shared.FromStringArrayToResources(i.flags.Requests),
 				},
-				HTTPPort: i.flags.HTTPPort,
+				HTTPPort:              i.flags.HTTPPort,
+				InsecureImageRegistry: i.flags.InsecureImageRegistry,
 			},
 			InfinispanMeta: v1alpha1.InfinispanMeta{InfinispanProperties: i.flags.infinispan},
 			KafkaMeta:      v1alpha1.KafkaMeta{KafkaProperties: i.flags.kafka},
--- a/cmd/kogito/command/install/jobs_service.go
+++ b/cmd/kogito/command/install/jobs_service.go
@@ -16,6 +16,7 @@ package install
 
 import (
 	"fmt"
+
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/common"
 
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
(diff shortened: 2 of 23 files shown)


sutaakar:
@ricardozanini @spolti This should fix the issue with insecure registry.

Makes me think, do we want to distinguish when the image tag should be marked as insecure and when not? Or should we mark all image streams as insecure?

ricardozanini:
@sutaakar I don't think we should do that. Do you have any reference explaining this approach? And why should we mark them as unsecure? To me, this is a OpenShift Infrastructure topic that would vary depending on the environment.

sutaakar:
@ricardozanini The issue which I try to solve is this:
User builds a Kogito application and upload the result image into some private registry without valid certificate.
How should user create KogitoRuntime application in OpenShift using the image stored in insecure registry? Because Kogito operator creates image stream without this insecure flag.

Should we support possibility to use such insecure registry? Or user must adjust OpenShift infrastructure to trust such registry?

ricardozanini:
In this case, we should add an attribute in the CR to enable unsecure imagestream, documenting this problem, not forcing it in the code like we are doing in the PR.

sutaakar:
Thanks, my intention was to initiate a discussion and find a solution (therefore just a draft PR).
@spolti What do you think about the suggestion above?

spolti:
I think that a property in the CR +  a flag on CLI would be enough :)

sutaakar:
Ready for review

ricardozanini:
@sutaakar great work, I'd suggest to freeze 0.11.0 by now since we are approaching the release. Should be ready to merge by the end of the week.

spolti:
+1 
let's reschedule it to 0.12.

pkg/apis/app/v1alpha1/types.go
```diff
@@ -20,7 +20,10 @@ Collection of common types among CRDs
 
 // Image is a definition of a Docker image
 type Image struct {
-	Domain    string `json:"domain,omitempty"`
+	Domain string `json:"domain,omitempty"`
+	// +optional
+	// A flag used to indicate the specified image registry is insecure. Defaults to 'false'.
+	Insecure  bool   `json:"insecure,omitempty"`
```
ricardozanini, line 26: I have to give more thoughts on this. Not sure if this would be the correct place to add this flag. I'll get back to it once I finish the work on #368

pkg/apis/app/v1alpha1/types.go
```diff
@@ -20,7 +20,10 @@ Collection of common types among CRDs
 
 // Image is a definition of a Docker image
 type Image struct {
-	Domain    string `json:"domain,omitempty"`
+	Domain string `json:"domain,omitempty"`
```
Sgitario, line 23: The format is wrong here?

sutaakar: The format change is affected by adding a comment below.

cmd/kogito/command/install/data_index.go
```diff
@@ -16,6 +16,7 @@ package install
```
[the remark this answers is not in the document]
sutaakar: Well, I was considering it.
I haven't added it there as KogitoApp becomes deprecated. Though I can add it if it has sense.

pkg/apis/app/v1alpha1/types.go
spolti: IMHO image and imagestream are different things.
Image in this case defines the fields of a Docker image, not a imagestream. Probably create a new struct would be better, or refactor this Image struct

cmd/kogito/command/install/data_index.go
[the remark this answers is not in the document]
ricardozanini: Please avoid adding more features to `KogitoApp` since it will be just a waste of time. Mainly for this feature, won't make sense now IMO.

pkg/apis/app/v1alpha1/types.go
ricardozanini: Yeah, @spolti. This is not an attribute for the image, thus to the registry. A `registry` structure or just a flag `insecureRegistry` or something like this would make more sense. Avoid using OpenShift related names in the interface since it will serve both clusters.

spolti: what about here? 
https://github.com/kiegroup/kogito-cloud-operator/pull/368/files#diff-4007d5c40c32998beb650053c8cb3045R53

:)

ricardozanini: oh `ImageHandler` is an internal structure to handle image naming/resolution. We need to figure out a structure in the `KogitoService` interface. :cry:

radtriste: Why not then having just a `insecureRegistry` as part of the KogitoServiceSpec ?

ricardozanini: ^ that was my suggestion up there. But it's out of context being in the root of the structure... I'm just nitpicking actually, we can go in this route

ricardozanini:
@sutaakar like we talked before, let's stick with `insecureImageRegistry` in the main `KogitoServiceSpec` interface with a comprehensive description for this flag. Default should be `false` since it's the default value for Go `bool`.

sutaakar:
Still needs a bit more manual testing, but brief check looks good.
Can be reviewed.

sutaakar:
Ready for review.

pkg/infrastructure/services/image.go
```diff
@@ -175,7 +177,7 @@ func NewImageHandlerForBuiltServices(image *v1alpha1.Image, namespace string, cl
 	}
 	if cli.IsOpenshift() {
 		// creates the empty tag reference in the ImageStream since this handler is for services being built
-		handler.createImageStream(namespace, false)
+		handler.createImageStream(namespace, false, false)
```
ricardozanini, line 180: @sutaakar I believe we would need this information here as well since this is the target `ImageStream` for the `KogitoRuntime` when the image is being built in the cluster. You might need to add this flag in the `KogitoBuild` CR as well to have it here.

ricardozanini: Ignore if the flag is only needed when pulling from an external registry, in the use cases where the image is built and pushing to a third party registry instead.

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -112,6 +112,7 @@ type KogitoServiceSpecInterface interface {
 	GetRuntime() RuntimeType
 	//Define port on which service will listen internally
 	GetHTTPPort() int32
+	IsInsecureImageRegistry() bool
```
ricardozanini, line 115: If we add this flag here, the `KogitoRuntime` will also inherit it. Is that right?

cmd/kogito/command/install/jobs_service.go
```diff
@@ -37,6 +38,7 @@ const (
 type installJobsServiceFlags struct {
 	deploy.CommonFlags
 	image                         string
+	insecureImageRegistry         bool
```
ricardozanini, line 41: Maybe having `image` and `insecureImageRegistry` on their own structure? This way we can reuse some code. Once we have implemented the Runtime and Build in the CLI we will be able to reuse it as well for the `deploy` command.

pkg/apis/app/v1alpha1/kogitoservices_types.go
sutaakar: Yes

pkg/infrastructure/services/image.go
sutaakar: Yes, the flag is used just for pulling.
Here we directly push image to image stream so there is no need for the flag.

cmd/kogito/command/install/jobs_service.go
sutaakar: Hmm, does it have sense to put `image` and `insecureImageRegistry` into one structure here? In `kogitoservices_types.go` we keep them separate.

ricardozanini: They do not follow the same context, IMO. Those are flags reusable across many commands, adding them in the same structure would improve reusablity.

sutaakar: ok, will try...

sutaakar:
@ricardozanini Moved `Image` and `InsecureImageRegistry` to `CommonFlags`. Please take another look.

PR #380, opened 2020-06-09
[KOGITO-2408] Correct Spring boot kafka bootstrap

radtriste:
https://issues.redhat.com/browse/KOGITO-2408

### DIFF
--- a/pkg/infrastructure/services/kafka.go
+++ b/pkg/infrastructure/services/kafka.go
@@ -16,6 +16,7 @@ package services
 
 import (
 	"fmt"
+
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
 	kafkabetav1 "github.com/kiegroup/kogito-cloud-operator/pkg/apis/kafka/v1beta1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/infrastructure/services/kafka.go
```diff
@@ -29,7 +30,7 @@ const (
 	QuarkusBootstrapAppProp = "kafka.bootstrap.servers"
 
 	// SpringBootstrapAppProp spring boot application property for setting kafka server
-	SpringBootstrapAppProp = "spring.kafka.bootstrap-servers"
+	SpringBootstrapAppProp = "kafka.bootstrap-servers"
```
Sgitario, line 33: Out of curiosity, why is "kafka.bootstrap-servers" for spring boot and "kafka.bootstrap.servers" for quarkus? Is this defined in the kogito examples or is set by the operator? I guess we should set the same parameter in both.

radtriste:
As discussed, we should keep default from Spring Boot and update the examples

pkg/infrastructure/services/kafka.go
ricardozanini: This property depends on the runtime being used. @radtriste are you removing the `spring` prefix because won't work on this runtime? The correct property is `kafka.boostrap-servers`? I remember seeing the Spring Docs `spring.kafka.boostrap-servers`. But I assume you ran some tests and that worked. :p

radtriste: the correct property is `spring.kafka.boostrap-servers`
Thing is that we introduce some specific behavior in our examples. I am correcting them :)

PR #386, opened 2020-06-15
[KOGITO-2451] Migrate nightly pipelines to RHBA

radtriste:
https://issues.redhat.com/browse/KOGITO-2435

and added skip tests parameter

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -1,10 +1,22 @@
+@Library('jenkins-pipeline-shared-libraries')_
+
 pipeline {
-    agent { label 'operator-slave' }
+    agent {
+        label 'kogito-operator-slave && !master'
+    }
+
+    // Needed for local build
+    tools {
+        jdk 'kie-jdk11'
+    }
 
     parameters {
         string(name: 'DISPLAY_NAME', defaultValue: '', description: 'Setup a specific build display name')
+        
         string(name: 'BUILD_BRANCH_NAME', defaultValue: 'master', description: 'Which branch to build ? Set if you are not on a multibranch pipeline.')
+        string(name: 'GIT_AUTHOR', defaultValue: 'kiegroup', description: 'Which Git author repository ?')
 
+        booleanParam(name: 'IMAGE_USE_OPENSHIFT_REGISTRY', defaultValue: false, description: 'Set to true if image should be deployed in Openshift registry.In this case, IMAGE_REGISTRY_CREDENTIALS, IMAGE_REGISTRY and IMAGE_NAMESPACE parameters will be ignored')
         string(name: 'IMAGE_REGISTRY_CREDENTIALS', defaultValue: '', description: 'Image registry credentials to use to deploy images. Will be ignored if no IMAGE_REGISTRY is given')
         string(name: 'IMAGE_REGISTRY', defaultValue: '', description: 'Image registry to use to deploy images')
         string(name: 'IMAGE_NAMESPACE', defaultValue: 'kiegroup', description: 'Image namespace to use to deploy images')
@@ -20,6 +32,8 @@ pipeline {
         string(name: 'EXAMPLES_REF', defaultValue: '', description: 'Git reference (branch/tag) to the kogito-examples repository to use for tests.')
 
         string(name: 'MAVEN_ARTIFACT_REPOSITORY', defaultValue: '', description: 'Maven repository where the build artifacts are present')
+        booleanParam(name: 'SMOKE_TESTS_ONLY', defaultValue: false, description: 'If only smoke tests should be run. Default is full testing.')
+        booleanParam(name: 'SKIP_TESTS', defaultValue: false, description: 'Skip tests')
     }
 
     options {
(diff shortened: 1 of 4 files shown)


Jenkinsfile.deploy
```diff
@@ -1,5 +1,15 @@
...
+
 pipeline {
-    agent { label 'operator-slave' }
+    agent {
+        label 'kogito-operator-slave && !master'
+    }
+
+    // Needed for local build
+    tools {
+        maven 'kie-maven-3.6.2'
```
Kaitou786, line 10: I think this can be removed as 3.6.3 is default for all kogito now.

Jenkinsfile.deploy
```diff
@@ -1,5 +1,15 @@
+@Library('jenkins-pipeline-shared-libraries')_
+
 pipeline {
-    agent { label 'operator-slave' }
+    agent {
+        label 'kogito-operator-slave && !master'
+    }
+
```
Kaitou786, line 7: Can't we put label in a single line?

Jenkinsfile.deploy
```diff
@@ -77,59 +83,58 @@ pipeline {
                     if (env.BRANCH_NAME != "") {
                         // Switch to branch if not on a multibranch pipeline
                         env.BRANCH_NAME = params.BUILD_BRANCH_NAME
-                        dir("${WORKING_DIR}"){
-                            checkout([$class: 'GitSCM', branches: [[name: env.BRANCH_NAME]], browser: [$class: 'GithubWeb', repoUrl: "${GIT_URL}"], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: '']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'user-kie-ci10', url: "${GIT_URL}"]]])
-                        }
+                        // TODO change author
+                        checkout(githubScm.resolveRepository("kogito-cloud-operator", "radtriste", env.BRANCH_NAME, false))
                     }
```
Kaitou786, line 88: Isn't the deploy file will always run on master branch? and author will be kiegroup right?

radtriste: It is supposed to run on kiegroup, but I can add a `BUILD_AUTHOR` parameter
But if we have release branches that are still enabled, it needs to be configurable.

Jenkinsfile.deploy
radtriste: I don't like much implicit dependencies in fact.
For now it is 3.6.3 in the definition, but if the definition gets a new default version and that breaks everything, that will be a lot of maintenance to update all ...

radtriste: just thinking that it would be the same if we need to upgrade the maven version.
btw, why is not jdk 11 by default ?

Kaitou786: We can let the maven version stay in place, apparently, if you try `kie-maven-3.6.3` it doesn't work, needs to be configured on the tools plugin I guess, Though I don't think it is going to be downgraded as all the kogito is now working with 3.6.2+. Regarding Java 11, I'm not really sure It has been the default since the beginning with all java versions available, pipelines uses whichever version they require.

Jenkinsfile.deploy
Kaitou786: Okay

Jenkinsfile.deploy
Kaitou786: Yeah, tool will now work with 3.6.3, so if you want we keep using 3.6.2 or upgrade it with 3.6.3

hack/go-build.sh
```diff
@@ -80,7 +81,7 @@ if [[ -z ${CI} ]]; then
     ./hack/go-test.sh
     BUILD_PARAMS=""
     if [[ ! -z ${BUILDER} ]]; then BUILD_PARAMS="${BUILD_PARAMS} --image-builder ${BUILDER}"; fi
-    operator-sdk build ${REGISTRY}/${IMAGE}:${TAG} ${BUILD_PARAMS}
+    operator-sdk build ${REGISTRY}/${IMAGE}:${TAG} ${BUILD_PARAMS} --go-build-args -o=${BINARY_OUTPUT}
```
radtriste, line 84: this will force the output name to `kogito-cloud-operator` because default, operator-sdk is taking the folder name for the binary and it may not run in a `kogito-cloud-operator` named folder

Jenkinsfile.deploy
radtriste: added the GIT_AUTHOR

Jenkinsfile.deploy
spolti: +1 for remove it if the current maven version is 3.6.3

Jenkinsfile.deploy
```diff
@@ -1,9 +1,21 @@
...
+
+    // Needed for local build
+    tools {
+        maven 'kie-maven-3.6.2'
+        jdk 'kie-jdk11'
+    }
 
     parameters {
         string(name: 'DISPLAY_NAME', defaultValue: '', description: 'Setup a specific build display name')
+
```
spolti, line 16: can be removed?

Jenkinsfile.deploy
```diff
@@ -1,9 +1,21 @@
...
+    tools {
+        maven 'kie-maven-3.6.2'
+        jdk 'kie-jdk11'
+    }
 
     parameters {
         string(name: 'DISPLAY_NAME', defaultValue: '', description: 'Setup a specific build display name')
+        
         string(name: 'BUILD_BRANCH_NAME', defaultValue: 'master', description: 'Which branch to build ? Set if you are not on a multibranch pipeline.')
+        string(name: 'GIT_AUTHOR', defaultValue: 'kiegroup', description: 'Which Git author repository ?')
```
spolti, line 18: is it really necessary? can't this info be picked up automatically?

Jenkinsfile.deploy
```diff
@@ -28,104 +41,86 @@ pipeline {
     }
 
     environment {
+        JAVA_HOME = "${GRAALVM_HOME}"
```
spolti, line 44: not sure if use graalvm as JAVA home is a good idea.
probably define JAVA_home with the openjdk and GRAALVM_HOME would be better,

Jenkinsfile.deploy
```diff
@@ -28,104 +41,86 @@ pipeline {
...
-                        touch /etc/sub{u,g}id
-                        usermod --add-subuids 10000-75535 \$(whoami)
-                        usermod --add-subgids 10000-75535 \$(whoami)
-                        cat /etc/subuid
-                        cat /etc/subgid
-                        make image_builder=buildah
-                    """
-                }
+                sh """
+                    go get -u golang.org/x/lint/golint
```
spolti, line 91: is the rhba env running on a container?
if not probably there is no need to do this step everytime. unless the directory the dependencies are saved is cleaned up after the build.

Jenkinsfile.deploy
```diff
@@ -28,104 +41,86 @@ pipeline {
...
             }
             
         }
         stage('Build Kogito CLI') {
             steps {
-                dir ("${WORKING_DIR}") {
-                    sh "make build-cli"
-                }
+                sh """
+                    go get -u github.com/gobuffalo/packr/v2/packr2
```
spolti, line 100: same here

Jenkinsfile.promote
```diff
@@ -23,35 +25,37 @@ pipeline {
...
+    }
 
     stages {
         stage('Initialization') {
             steps {
                 script {
-                    sh "docker rm -f \$(docker ps -a -q) || docker rmi -f \$(docker images -q) || date"
-
+                    cleanImages()
+
```
spolti, line 58: extra line

Jenkinsfile.promote
```diff
@@ -152,20 +163,23 @@ boolean isDeployLatestTag(){
...
-        sh "set +x && docker login -u admin -p \$(oc whoami -t) ${OPENSHIFT_REGISTRY_ROUTE}"
+void loginOpenshift(){
+    withCredentials([usernamePassword(credentialsId: "OPENSHIFT_CREDS", usernameVariable: 'OC_USER', passwordVariable: 'OC_PWD')]){
+        sh "oc login --username=${OC_USER} --password=${OC_PWD} --server=${OPENSHIFT_API} --insecure-skip-tls-verify"
     }
 }
 
+void loginOpenshiftRegistry(){
+    loginOpenshift()
+    sh "set +x && podman login -u jenkins -p \$(oc whoami -t) --tls-verify=false ${OPENSHIFT_REGISTRY}"
```
spolti, line 174: we probably should not expose any user information here, can we use envs instead?

Jenkinsfile.deploy
Kaitou786: The jobs run on VM which is destroyed after the job is complete, for every PR a new VM is created, so yes it'll be a clean run

Jenkinsfile.deploy
```diff
@@ -28,104 +41,86 @@ pipeline {
...
 
-                    if (env.BRANCH_NAME != "") {
-                        // Switch to branch if not on a multibranch pipeline
-                        env.BRANCH_NAME = params.BUILD_BRANCH_NAME
-                        dir("${WORKING_DIR}"){
-                            checkout([$class: 'GitSCM', branches: [[name: env.BRANCH_NAME]], browser: [$class: 'GithubWeb', repoUrl: "${GIT_URL}"], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: '']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'user-kie-ci10', url: "${GIT_URL}"]]])
-                        }
-                    }
+                    checkout(githubscm.resolveRepository("kogito-cloud-operator", params.GIT_AUTHOR, params.BUILD_BRANCH_NAME, false))
                 }
```
Kaitou786, line 85: Can't we use `githubscm.checkoutIfExists` function over here from shared library

Jenkinsfile.promote
Kaitou786: +1 we can store this in Jenkins credentials

Jenkinsfile.deploy
radtriste: ok will remove it

Jenkinsfile.deploy
radtriste: I'd rather keep parameters in "groups" for readability of it

Jenkinsfile.deploy
radtriste: nope we don't have that information. It is useful for anybody wanted to build on its own repository

Jenkinsfile.deploy
radtriste: we need to build native example locally with Maven so Graalvm needs to be default for the tests (it is only used there)...

Jenkinsfile.promote
radtriste: as for images, username is useless but needed.
https://github.com/kiegroup/kogito-images/pull/182#discussion_r447716950

Jenkinsfile.deploy
radtriste: https://github.com/kiegroup/kogito-images/pull/182#discussion_r447714401

Jenkinsfile.deploy
radtriste: removed

Jenkinsfile.promote
radtriste: updated

Jenkinsfile.promote
```diff
@@ -152,20 +163,24 @@ boolean isDeployLatestTag(){
...
+void loginOpenshift(){
+    withCredentials([usernamePassword(credentialsId: "OPENSHIFT_CREDS", usernameVariable: 'OC_USER', passwordVariable: 'OC_PWD')]){
+        sh "oc login --username=${OC_USER} --password=${OC_PWD} --server=${OPENSHIFT_API} --insecure-skip-tls-verify"
     }
 }
 
+void loginOpenshiftRegistry(){
+    loginOpenshift()
+    // username can be anything. See https://docs.openshift.com/container-platform/4.4/registry/accessing-the-registry.html#registry-accessing-directly_accessing-the-registry
+    sh "set +x && podman login -u openshift -p \$(oc whoami -t) --tls-verify=false ${OPENSHIFT_REGISTRY}"
```
spolti, line 175: can we remove this user from here and use envs?

radtriste: Like I wrote into the comment just before, the username can be anything it won't be taken into account.
If you want I can put `anything` instead of `openshift` as username but I don't see the point of overloading the env with an information which is useless in the end.

radtriste:
new image not yet ready so putting on hold here as the image set is a test image

test/features/deploy_quarkus_service.feature
```diff
@@ -150,6 +150,9 @@ Feature: Deploy quarkus service
       | native   | minutes |
       | disabled | 10      |
 
+    # Disabled until Quarkus 1.6+ is used due to https://github.com/quarkusio/quarkus/issues/9954 
+    # Follow-up issue: https://issues.redhat.com/browse/KOGITO-2616
+    @disabled
```
radtriste, line 155: cherry-pick from https://github.com/kiegroup/kogito-cloud-operator/pull/418 for the tests

radtriste:
I changed a bit the pipeline to handle correctly the parameters. I find this way much more readable and prone to easier modification.

- test parameters -> exported to single method that creates a map of params and then print as `key=value` string
- The `DEPLOY_` are no longer used in environment. Instead I used some methods that will calculate the correct value. It results in a more readable code IMO

Jenkinsfile.deploy
```diff
@@ -213,76 +138,133 @@ String getOperatorVersion(){
...
+    if (params.MAVEN_ARTIFACT_REPOSITORY != '') {
+        // No mirror if we set directly the Maven repository
+        // Tests will be slower but we need to test against specific artifacts
+        testParamsMap["custom_maven_repo"] = params.MAVEN_ARTIFACT_REPOSITORY
+    } else if(env.MAVEN_MIRROR_REPOSITORY != null
+                && env.MAVEN_MIRROR_REPOSITORY != '') {
+        testParamsMap["maven_mirror"] = env.MAVEN_MIRROR_REPOSITORY
+    }
+
+    if (params.KOGITO_IMAGES_IN_OPENSHIFT_REGISTRY) {
```
sutaakar, line 237: Please add runtime_application_* properties here too.

radtriste: good point. will add them

radtriste: added. let me know

sutaakar: Looks good

sutaakar:
@radtriste One more thing, can you add also `image_cache_mode`?

spolti:
Is this PR ready to merge?

radtriste:
@sutaakar are you ok with the latest changes ?

sutaakar:
yes, looks good

PR #389, opened 2020-06-16
[KOGITO-2406] - Add Test Coverage to the Operator PR check CI

Kaitou786:
SEE: https://issues.redhat.com/browse/KOGITO-2406


spolti:
> That's a well done job! Many thanks Tarun! :)
Indeed :D

Kaitou786:
Guys, I was thinking as `bats test` takes alot of time to run (Compared to other tests). So do we add a require field to only run `bats test` when `lint` and `unit test` are passed?

ricardozanini:
> Guys, I was thinking as `bats test` takes alot of time to run (Compared to other tests). So do we add a require field to only run `bats test` when `lint` and `unit test` jobs are passed?

You mean to run `bats` only when unit test and lint have passed? It's not possible to run `bats` in parallel and have the `unit test` to be dependent to `lint`?

So:

1. `lint` and `bats` at the same time
2. `unit test` if `lint` passes

b/c in the vast majority of time, unit and lint will pass, running `bats` simultaneously will decrease the CI process time, no?

ricardozanini:
Putting on hold to finish this discussion. DO NOT MERGE.

Kaitou786:
@ricardozanini  Yes, we can set the CI to run `lint` and `bats` at same time and hold `unit test` till lint passes, but my point is if we start in the beginning  `bats` and let say there was a small linting error, that would just leave bats to run till they complete, even when in between the developer has fixed the linting error and pushed another commit, on circleCI the bats from previous commit will also kept running till they pass or fail.
So I suggest we run `lint` in the start and if it passes run `unit test` and `bats` in parallel, wdyt?

spolti:
sounds good to me @Kaitou786.
If lint fails we could save time and resources failing the pipeline right away.

ricardozanini:
> So I suggest we run `lint` in the start and if it passes run `unit test` and `bats` in parallel, wdyt?

Lint shouldn't fail in the most cases since `make test` always runs it. But I understand your point, yeah, we can do this way. Are you going to update this PR?

Kaitou786:
> Lint shouldn't fail in the most cases since `make test` always runs it. But I understand your point, yeah, we can do this way. Are you going to update this PR?

Yes, I'll update this PR tomorrow morning :)

hack/go-lint.sh
```diff
@@ -20,6 +20,6 @@ fi
 rm -f golint_errors
 # The command in or will fetch the latest tag available for golangci-lint and install in $GOPATH/bin/
 which golangci-lint > /dev/null || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
-golangci-lint run ./... --enable golint
+golangci-lint run ./... --enable golint --timeout 2m0s
```
radtriste, line 23: this seems to be a really short timeout or ?

Kaitou786: By default it was 1m0s, I think 2m0s is enough

radtriste: oki

PR #397, opened 2020-06-22
[KOGITO-2090] Integrate KogitoBuild in tests

sutaakar:
https://issues.redhat.com/browse/KOGITO-2090

### DIFF
--- a/test/features/deploy_kogito_build.feature
+++ b/test/features/deploy_kogito_build.feature
@@ -1,0 +1,107 @@
+# Disabled until CLI (used in PR check) supports KogitoRuntime and KogitoBuild - KOGITO-2064
+@disabled
+Feature: Deploy Kogito Build
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+
+  Scenario Outline: Build <runtime> remote S2I with native <native> using KogitoBuild
+    When Build <runtime> example service "<example-service>" with configuration:
+      | config | native | <native> |
+
+    Then Build "<example-service>-builder" is complete after <minutes> minutes
+    And Build "<example-service>" is complete after 5 minutes
+    And Deploy runtime <runtime> example service "<example-service>" with configuration:
+      | infinispan | useKogitoInfra | disabled |
+    And Kogito Runtime "<example-service>" has 1 pods running within 5 minutes
+    And Service "<example-service>" with process name "orders" is available within 2 minutes
+
+    @springboot
+    Examples:
+      | runtime    | example-service            | native   | minutes |
+      | springboot | process-springboot-example | disabled | 10      |
+
+    @smoke
+    @quarkus
+    Examples:
+      | runtime    | example-service         | native   | minutes |
+      | quarkus    | process-quarkus-example | disabled | 10      |
+
+    @quarkus
+    @native
+    Examples:
+      | runtime    | example-service         | native  | minutes |
+      | quarkus    | process-quarkus-example | enabled | 20      |
+
+#####
+
+  Scenario Outline: Build <runtime> binary build with native <native> using KogitoBuild
+    Given Clone Kogito examples into local directory
+    And Local example service "<example-service>" is built by Maven using profile "<profile>"
+
+    When Build binary <runtime> service "<example-service>" with configuration:
+      | config | native | <native> |
+
+    Then BuildConfig "<example-service>" is created after 1 minutes
+    # Once https://issues.redhat.com/browse/KOGITO-2161 is implemented then we need to refactor this scenario to CLI implementation (getting rid of manual build trigger)
+    And Start build with name "<example-service>" from local example service path "<example-service>/target"
+    And Deploy runtime <runtime> example service "<example-service>" with configuration:
+      | infinispan | useKogitoInfra | disabled |
+    And Kogito Runtime "<example-service>" has 1 pods running within 5 minutes
+    And Service "<example-service>" with process name "orders" is available within 2 minutes
+
+    @springboot
+    Examples:
+      | runtime    | example-service            | native   | profile |
+      | springboot | process-springboot-example | disabled | default |
+
+    @quarkus
+    Examples:
+      | runtime    | example-service         | native   | profile |
+      | quarkus    | process-quarkus-example | disabled | default |
+
+    @quarkus
+    @native
+    Examples:
+      | runtime    | example-service         | native  | profile |
+      | quarkus    | process-quarkus-example | enabled | native  |
+
+#####
+
+  Scenario Outline: Build <runtime> local source build with native <native> using KogitoBuild
+    Given Clone Kogito examples into local directory
+
+    When Build <runtime> example service "process-scripts-quarkus" from local file with configuration:
+      | config | native | <native> |
+
+    Then BuildConfig "process-scripts-quarkus-builder" is created after 1 minutes
+    # Once https://issues.redhat.com/browse/KOGITO-2161 is implemented then we need to refactor this scenario to CLI implementation (getting rid of manual build trigger)
+    And Start build with name "process-scripts-quarkus-builder" from local example service file "process-scripts-quarkus/src/main/resources/org/acme/travels/scripts.bpmn"
+    And Build "process-scripts-quarkus-builder" is complete after <minutes> minutes
+    And Build "process-scripts-quarkus" is complete after 5 minutes
+    And Deploy runtime <runtime> example service "process-scripts-quarkus" with configuration:
+      | infinispan | useKogitoInfra | disabled |
+    And Kogito Runtime "process-scripts-quarkus" has 1 pods running within 5 minutes
+    And Start "scripts" process on service "process-scripts-quarkus" within 2 minutes with body:
+      """json
+      {
+          "name" : "john"
+      }
+      """
+
+    @springboot
+    Examples:
+      | runtime    | native   | minutes |
+      | springboot | disabled | 10      |
+
+    @quarkus
+    Examples:
+      | runtime    | native   | minutes |
+      | quarkus    | disabled | 10      |
+
+    @quarkus
+    @native
+    Examples:
+      | runtime    | native  | minutes |
+      | quarkus    | enabled | 20      |
--- a/test/framework/kogitobuild.go
+++ b/test/framework/kogitobuild.go
@@ -1,0 +1,105 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


sutaakar:
Depends on https://github.com/kiegroup/kogito-cloud-operator/pull/368

ricardozanini:
This is @sutaakar 's revenge. I'll take a look once we are good to go with #368, many thanks! Seems a real great job!

spolti:
will review when  #368 is merged

sutaakar:
Should be ready for review now.
Tests are currently disabled as they are available just for CR run and CI pipeline use CLI.

test/framework/kogitobuild.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
+	"github.com/kiegroup/kogito-cloud-operator/test/config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DeployKogitoBuild deploy a KogitoBuild
+func DeployKogitoBuild(namespace string, installerType InstallerType, kogitoBuild *v1alpha1.KogitoBuild) error {
+	GetLogger(namespace).Infof("%s deploy %s example %s with name %s and native %v", installerType, kogitoBuild.Spec.Runtime, kogitoBuild.Spec.GitSource.ContextDir, kogitoBuild.Name, kogitoBuild.Spec.Native)
+
+	switch installerType {
```
ricardozanini, line 32: Just a nitpick, but I'd change this to a simple `if` instead.

test/features/deploy_kogito_build.feature
```diff
@@ -0,0 +1,70 @@
...
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+
+  Scenario Outline: Build <runtime> remote S2I with native <native> using KogitoBuild
+    When Build <runtime> example service "<example-service>" with configuration:
+      | config | native | <native> |
+
+    Then Build "<example-service>-builder" is complete after <minutes> minutes
+    And Build "<example-service>" is complete after 5 minutes
```
Sgitario, line 14: Do we need both or only with "And Build "<example-service>" is complete after 5 minutes" is enough? Also, I guess we should use the minutes here: <minutes>

test/features/deploy_kogito_build.feature
```diff
@@ -0,0 +1,70 @@
...
+  @springboot
+  Scenario Outline: Build <runtime> binary build with native <native> using KogitoBuild
+    Given Clone Kogito examples into local directory
+    And Local example service "<example-service>" is built by Maven
+
+    When Build binary <runtime> service "<example-service>" with configuration:
+      | config | native | <native> |
+
+    Then BuildConfig "<example-service>" is created after 1 minutes
+    And Start build with name "<example-service>" from local example service path "<example-service>/target"
```
Sgitario, line 49: I'm confused about this. Why do we need to start the build ourselves in the binary scenario? Using RemoteSourceBuildType is not needed. I would have expected to behave the same after running the Build binary ....

test/framework/kogitobuild.go
```diff
@@ -0,0 +1,107 @@
...
+	return kogitoBuild
+}
+
+// SetupKogitoBuildImageStreams sets the correct images for the KogitoBuild
+func SetupKogitoBuildImageStreams(kogitoBuild *v1alpha1.KogitoBuild) {
+	kogitoBuild.Spec.BuildImage = getKogitoBuildS2IImage(kogitoBuild)
+	kogitoBuild.Spec.RuntimeImage = getKogitoBuildRuntimeImage(kogitoBuild)
+}
+
+func getKogitoBuildS2IImage(kogitoBuild *v1alpha1.KogitoBuild) v1alpha1.Image {
```
Sgitario, line 72: This method is almost duplicated in [here](https://github.com/kiegroup/kogito-cloud-operator/blob/2a15042e1717c1af2f6d0e8b08803b69d01f316f/test/framework/kogitoapp.go#L216). Maybe, we could refactor it to make use of it also here.

test/framework/kogitobuild.go
```diff
@@ -0,0 +1,107 @@
...
+	if len(config.GetBuildS2IImageStreamTag()) > 0 {
+		return framework.ConvertImageTagToImage(config.GetBuildS2IImageStreamTag())
+	}
+
+	return getKogitoBuildImage(resource.BuildImageStreams[resource.BuildTypeS2I][kogitoBuild.Spec.Runtime])
+}
+
+func getKogitoBuildRuntimeImage(kogitoBuild *v1alpha1.KogitoBuild) v1alpha1.Image {
```
Sgitario, line 80: This method is almost duplicated in [here](https://github.com/kiegroup/kogito-cloud-operator/blob/2a15042e1717c1af2f6d0e8b08803b69d01f316f/test/framework/kogitoapp.go#L199). Maybe we can make use of it after a tiny refactor.

test/framework/kogitoserviceutils.go
```diff
@@ -105,7 +106,6 @@ func newImageOrDefault(fullImage string, defaultImageName string) v1alpha1.Image
 
 		image.Domain = config.GetServicesImageRegistry()
 		image.Namespace = config.GetServicesImageNamespace()
-		image.Name = defaultImageName
```
Sgitario, line 108: Is this correct? If the GetServicesImageNameSuffic is not provided, then the name won't be set.

test/steps/kogitoruntime.go
```diff
@@ -49,10 +50,18 @@ func registerKogitoRuntimeSteps(s *godog.Suite, data *Data) {
...
-
 func (data *Data) deployExampleServiceFromRuntimeRegistryWithConfiguration(runtimeType, kogitoApplicationName string, table *messages.PickleStepArgument_PickleTable) error {
 	imageTag := data.ScenarioContext[getBuiltRuntimeImageTagContextKey(kogitoApplicationName)]
+	return data.deployExampleServiceFromImageWithConfiguration(runtimeType, kogitoApplicationName, imageTag, table)
+}
+
+// Can be renamed to deployExampleServiceWithConfiguration once KogitoApp is removed
+func (data *Data) deployRuntimeExampleServiceWithConfiguration(runtimeType, kogitoApplicationName string, table *messages.PickleStepArgument_PickleTable) error {
+	// Passing empty image tag so image values are not filled
+	return data.deployExampleServiceFromImageWithConfiguration(runtimeType, kogitoApplicationName, "", table)
```
Sgitario, line 61: Question: if the image is empty, the operator will try to match the image stream with same name? The KogitoBuild will have created the image stream previously. 
Is this correct?

test/features/deploy_kogito_build.feature
```diff
@@ -0,0 +1,68 @@
+Feature: Deploy Kogito Build
```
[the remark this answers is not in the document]
sutaakar: Yes, because local source is currently available just through CLI which is not available yet for KogitoBuild

test/features/deploy_kogito_build.feature
sutaakar: Well, I am checking here that both builds are executed.
It depends on what we actually want to check here.

Do we want to make sure that operator creates both builds and executes them successfully? Or we consider it an implementation detail and are interested just in running instance?

I am inclined to check the builds too.

About the minutes - the second build is moreless same for all scenario options, so I kept there hardcoded value. The first build is specific for each build, so I keep the timeout flexible there.

test/features/deploy_kogito_build.feature
sutaakar: Yes, we need to start the build by ourselves.
It is because we upload the binary using OC client, which is the separate step from KogitoBuild creation. Once we will have support for binary upload in CLI we can unify the steps (for CLI).

test/framework/kogitobuild.go
sutaakar: Yes, you are right.
But once KogitoApp is removed then the referenced implementation is removed too and this one will become the only one.
So I just didn't want to refactor something which will be removed in future anyway.

test/framework/kogitobuild.go
sutaakar: Same as above.

test/framework/kogitoserviceutils.go
sutaakar: Yes, if the name is not set then default name is applied by Kogito operator.
So we don't have to specify it by ourselves but rather leverage operator functionality where possible.

test/steps/kogitoruntime.go
sutaakar: Yes, correct.

test/features/deploy_kogito_build.feature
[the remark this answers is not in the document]
ricardozanini: Actually you can use `oc start-build..` and give a file to it instead of relying on CLI. :stuck_out_tongue_closed_eyes:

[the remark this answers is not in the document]
Sgitario: So, is it possible? If not, I think we should raise a ticket for covering when can be done.

test/features/deploy_kogito_build.feature
Sgitario: Ok

test/features/deploy_kogito_build.feature
Sgitario: Ok, can we add a comment here with a link to the ticket when this will be refactored?

test/framework/kogitobuild.go
Sgitario: Ok, it makes sense.

test/framework/kogitobuild.go
Sgitario: Ok, it makes sense.

test/framework/kogitoserviceutils.go
Sgitario: Ok, thanks for explaining!

test/steps/kogitoruntime.go
Sgitario: Ok

test/features/deploy_kogito_build.feature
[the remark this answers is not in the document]
sutaakar: Will try...

sutaakar:
@ricardozanini @Sgitario Can you please take a look again?
I have added there build from local source and addressed comments.

sutaakar:
@spolti Can you please take a look?

test/framework/kogitoserviceutils.go
sutaakar: it turns out that this is needed for CLI installer, returning it back

spolti:
is sonarcloud live again?

PR #404, opened 2020-06-25
[KOGITO-2119] - Kogito CLI support for setting variable values from s…

vaibhavjainwiz:
…ecret

Jira issue : https://issues.redhat.com/browse/KOGITO-2119


cmd/kogito/command/shared/converters.go
```diff
@@ -19,14 +19,32 @@ import (
...
+	"strings"
 )
 
+const secretKeyNameSeparator = "#"
+
 // FromStringArrayToEnvs converts a string array in the format of key=value pairs to the required type for the Kubernetes EnvVar type
-func FromStringArrayToEnvs(strings []string) []v1.EnvVar {
-	if strings == nil {
+func FromStringArrayToEnvs(keyPairStrings []string) []v1.EnvVar {
+	if keyPairStrings == nil {
```
ricardozanini, line 29:
```suggestion
	if len(keyPairStrings) == 0 {
```

vaibhavjainwiz: done

ricardozanini:
We merged some CI changes, could you please rebase?

cmd/kogito/command/shared/converters.go
```diff
@@ -19,14 +19,32 @@ import (
...
-	if strings == nil {
+func FromStringArrayToEnvs(keyPairStrings []string) []v1.EnvVar {
+	if len(keyPairStrings) == 0 {
 		return nil
 	}
-	return framework.MapToEnvVar(util.FromStringsKeyPairToMap(strings))
+	var envVars []v1.EnvVar
+	keyPairMap := util.FromStringsKeyPairToMap(keyPairStrings)
+	for key, value := range keyPairMap {
+		if strings.Contains(value, secretNameKeySeparator) {
```
radtriste, line 35: one question: what happens if the value of an env var contains a `#` and is not supposed to be a secret, like a password for example ?
the value could contain a `#` where as the key should not as it should follow the env var naming convention.

Shouldn't we instead use a specific "equality" like for example `#=` (that can be something else) so we are sure that it is a secret name/key that is set for the variable ?
that would give us `MY_VAR#=secretName#secretKey`

sutaakar: Hmm, wouldn't it rather have sense to create a separate flag for variable from secret?
That way we can keep a flexibility in case we will need to add another variable source, for example config map.

ricardozanini: @vaibhavjainwiz have you discussed this feature with @MarianMacik? Since he asked for this, I believe he could weight in. @radtriste has a good point tho.

vaibhavjainwiz: > Hmm, wouldn't it rather have sense to create a separate flag for variable from secret?
> That way we can keep a flexibility in case we will need to add another variable source, for example config map.

@sutaakar I came to thought to introduce seperate variable to capture secrets but after analysis of code I found that we already have two different flag to capture envs(--env & --build-env) and user can provide secret in either of them. So if I go with this approch then we have add two more flag(`--env-secret & --build-env-secret`) which looks like too many flags just to capture env variable.

vaibhavjainwiz: > @vaibhavjainwiz have you discussed this feature with @MarianMacik? Since he asked for this, I believe he could weight in. @radtriste has a good point tho.

@ricardozanini Yes, I represent my approach to @MarianMacik on Jira and took his confirmation.

vaibhavjainwiz:
> We merged some CI changes, could you please rebase?

done

cmd/kogito/command/shared/converters.go
vaibhavjainwiz: @sutaakar @ricardozanini @radtriste Please suggest How we should processed further with this. Should I created new seperate flag to capture secret, chage secret seperator or leave it as it is?

ricardozanini: @vaibhavjainwiz I'd say to let it as is and open a new JIRA to dive into the discussion, but since @radtriste brought this up, I'll give him the final word.

radtriste: Honestly, I would not let it like that as this can be problematic.
If we need a solution for now, I would go for @sutaakar idea, with many options. This can be refactored later and is more visible to not forget.
It makes more options for sure but at least, we know exactly what it does and there is no specific convention to write a secret in the `--env` option.

vaibhavjainwiz: ok then, I go with @sutaakar idea to introduce new variables

vaibhavjainwiz: done

cmd/kogito/command/shared/converters.go
```diff
@@ -16,17 +16,42 @@ package shared
 
 import (
 	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
-	"github.com/kiegroup/kogito-cloud-operator/pkg/util"
```
ricardozanini, line 19: Question: you removed from the util package because these functions were only used in here? Many thanks for it!

cmd/kogito/command/util/common_util_test.go
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+package util
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func Test_ValidateKeyPair_CorrectInput(t *testing.T) {
```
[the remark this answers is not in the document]
vaibhavjainwiz: Thanks for for new learning. Table test driven is new to me, I had never used it before. From next time I will try to implement it

cmd/kogito/command/shared/converters.go
vaibhavjainwiz: Previous `FromStringsKeyPairToMap()` is written `map.go` file under kogito-cloud-operator package but it is only used in kogito-cli -> `converters.go` file, so I moved this function from util package to converter file.

radtriste:
@vaibhavjainwiz can you confirm I can merge that one ?

ricardozanini:
Adding `onhold` to wait for @vaibhavjainwiz confirmation.
@radtriste I'll send a PR to bump master to 0.13, please do not merge any PRs that impacts CRDs/CRs.

vaibhavjainwiz:
> @vaibhavjainwiz can you confirm I can merge that one ?

@ricardozanini @radtriste Please merge this PR

PR #412, opened 2020-06-30
[KOGITO-2527] - Upgrade Operator Golang to 1.14

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-2527

Signed-off-by: Tarun Khandelwal <tarkhand@redhat.com>

Many thanks for submiting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.MD#sending-a-pull-request)
- [x] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket
- [ ] Your feature/bug fix has a unit test that verifies it
- [ ] You've tested the new feature/bug fix in an actual OpenShift cluster

PR #413, opened 2020-06-30
[KOGITO-2550] - Fixed vet test

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-2550

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,27 +9,37 @@ jobs:
         environment:
           - GO111MODULE=on
           - GOLANG_LINT_VERSION=v1.27.0
+          - OPERATOR_SDK_VERSION=v0.18.2
     working_directory: /go/src/github.com/kiegroup/kogito-cloud-operator
     steps:
       - checkout
       - run:
-          name: Check Formatting
+          name: Install operator-sdk
           command: |
-            ./hack/go-fmt.sh
-            changed_files=$(git status -s | grep -v 'go.mod\|go.sum' || :)
-            [[ -z "$changed_files" ]] ||  (printf "Some files are not formatted properly: \n$changed_files\n Please make sure to run "make test" before sending the PR" && exit 1)
-      - run:
-          name: Check Headers
-          command: |
-            ./hack/addheaders.sh
-            changed_files=$(git status -s | grep -v 'go.mod\|go.sum' || :)
-            [[ -z "$changed_files" ]] ||  (printf "Some files are missing the headers: \n$changed_files\n Please make sure to run "make test" before sending the PR" && exit 1)
+            sudo curl -L https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/operator-sdk-$OPERATOR_SDK_VERSION-x86_64-linux-gnu -o /usr/bin/operator-sdk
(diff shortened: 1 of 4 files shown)


hack/go-vet.sh
```diff
@@ -14,19 +14,16 @@
 # limitations under the License.
 
 
-if [[ -z ${CI} ]]; then
-    ./hack/go-mod.sh
```
Kaitou786, line 18: Because of this condition the script was not executing. In circle CI this is an enviornment variable set to `true`

Kaitou786:
Because the steps were missed(due to that if condition) there are some changes in the files. I changed those in https://github.com/kiegroup/kogito-cloud-operator/pull/412. After that is merged I can rebase and this one will succeed

hack/go-vet.sh
ricardozanini: You want this to be executed to catch when someone send a PR without running `make mod`?

Kaitou786: Not just this step, but all the steps were in this condition like operator-sdk generate k8s and crds, which were not executed on this CircleCI

Kaitou786: Because of this condition the script `go-vet.sh` was only running a single command `go vet ./...`

ricardozanini:
@Kaitou786 this one depends on #412?

Kaitou786:
@ricardozanini on the master branch itself if you run `make test` you'll see the changes on the CRDs, I first noticed this when sending #412 . So after the #412 is merged the lint stage here will pass (After Rebase)

spolti:
can you verify why lint in failing?

Kaitou786:
@spolti yes, the lint is failing because the change is detected in the crds. Previously this check was not able to detect the changes as the operator-sdk command was not executing,  because of that some commit got merged in the master which should have also changed the crds. I changed added those crds in #412 After the rebase the lint over here will succeed.

Kaitou786:
@ricardozanini I made some changes can you review again?

.circleci/config.yml
```diff
@@ -9,27 +9,37 @@ jobs:
...
+            sudo curl -L https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/operator-sdk-$OPERATOR_SDK_VERSION-x86_64-linux-gnu -o /usr/bin/operator-sdk
+            sudo chmod +x /usr/bin/operator-sdk
+            # For debugging
+            operator-sdk version
+      - run:
+          name: Check Vet
+          command: |
+            ./hack/go-vet.sh
+            changed_files=$(git status -s | grep -v 'go.mod\|go.sum' || :)
+            openapi_changed_files=$(git status -s | grep zz_generated || :)
```
ricardozanini, line 28: Just a nitpick in here. But `zz_generated` is also a prefix for the controller-gen object deepcopy generation. So the variable name could be `generated_changed_files`. :D

Kaitou786: What do you suggest should we handle only the openapi files or for both generated files?

ricardozanini: I think both. DeepCopy also changes if have a `*_types` change. Your script already will caught that, just need to change the var name.

Kaitou786: Updated

PR #414, opened 2020-06-30
[KOGITO-2594] Implement script to generate Kogito example images

sutaakar:

https://issues.redhat.com/browse/KOGITO-2594

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,7 @@ You can set those optional keys:
 - `container_engine` engine used to interact with images and local containers.
   *Default is docker.*
 - `domain_suffix` domain suffix used for exposed services. Ignored when running tests on Openshift.
+- `image_cache_mode` Use this option to specify whether you want to use image cache for runtime images. Available options are 'always', 'never' or 'if-available'(default).
 <!--- operator information -->
 - `operator_image` is the Operator image full name.  
   *Default: operator_image=quay.io/kiegroup/kogito-cloud-operator*.
@@ -253,6 +254,8 @@ You can set those optional keys:
 <!--- build runtime applications -->
 - `runtime_application_image_registry` sets the registry for built runtime applications.
 - `runtime_application_image_namespace` sets the namespace for built runtime applications.
+- `runtime_application_image_name_suffix` sets the image name suffix to append to usual image names for built runtime applications.
+- `runtime_application_image_version` sets the version for built runtime applications.
 <!--- development options -->
 - `show_scenarios` sets to true to display scenarios which will be executed.  
   *Default is false.*
--- a/hack/run-tests.sh
+++ b/hack/run-tests.sh
@@ -49,6 +49,7 @@ function usage(){
   printf "\n--load_default_config\n\tTo be used if you want to directly use the default test config contained into ${SCRIPT_DIR}/../test/.default_config"
   printf "\n--container_engine\n\tTo be used if you want to specify engine to interact with images and containers. Default is docker."
   printf "\n--domain_suffix\n\tTo be used if you want to set a domain suffix for exposed services. Ignored when running tests on Openshift."
+  printf "\n--image_cache_mode\n\tUse this option to specify whether you want to use image cache for runtime images. Available options are 'always', 'never' or 'if-available'(default)."
 
   # operator information
   printf "\n--operator_image {NAME}\n\tOperator image name. Default is 'quay.io/kiegroup/kogito-cloud-operator' one."
@@ -69,6 +70,8 @@ function usage(){
   printf "\n--management_console_image_tag {IMAGE_TAG}\n\tSet the Kogito Management Console image tag ('services_image_version' is ignored)"
   printf "\n--runtime_application_image_registry {REGISTRY}\n\tSet the registry for built runtime applications."
   printf "\n--runtime_application_image_namespace {NAMESPACE}\n\tSet the namespace for built runtime applications."
+  printf "\n--runtime_application_image_name_suffix {NAME_SUFFIX}\n\tSet the image name suffix to append to usual image names for built runtime applications."
+  printf "\n--runtime_application_image_version {VERSION}\n\tSet the version for built runtime applications."
 
   # build
   printf "\n--custom_maven_repo {URI}\n\tSet a custom Maven repository url for S2I builds, in case your artifacts are in a specific repository. See https://github.com/kiegroup/kogito-images/README.md for more information."
@@ -211,6 +214,10 @@ case $1 in
     shift
     if addParamKeyValueIfAccepted "--tests.domain-suffix" ${1}; then shift; fi
   ;;
+  --image_cache_mode)
+    shift
+    if addParamKeyValueIfAccepted "--tests.image-cache-mode" ${1}; then shift; fi
+  ;;
 
   # operator information
   --operator_image)
@@ -268,6 +275,14 @@ case $1 in
   --runtime_application_image_namespace)
     shift
     if addParamKeyValueIfAccepted "--tests.runtime-application-image-namespace" ${1}; then shift; fi
+  ;;
+  --runtime_application_image_name_suffix)
+    shift
+    if addParamKeyValueIfAccepted "--tests.runtime-application-image-name-suffix" ${1}; then shift; fi
+  ;;
+  --runtime_application_image_version)
+    shift
+    if addParamKeyValueIfAccepted "--tests.runtime-application-image-version" ${1}; then shift; fi
   ;;
 
   # build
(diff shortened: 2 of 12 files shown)


sutaakar:
Some statistics related to the implementation:
```
  <testsuite name="Build process-quarkus-example images" tests="2" skipped="0" failures="0" errors="0" time="683.446678797">
    <testcase name="Build process-quarkus-example images" status="passed" time="352.952519068"></testcase>
    <testcase name="Build native process-quarkus-example images" status="passed" time="330.346043232"></testcase>
  </testsuite>
  <testsuite name="Build process-springboot-example images" tests="1" skipped="0" failures="0" errors="0" time="229.565909194">
```
`Build process-quarkus-example images` also pulled all the maven artifacts which affected overall execution time. So in case of Maven proxy we can expect 3 minutes less there.

So, to build all profile combinations for one example service takes roughly 3-4 minutes. One native build takes more than 5 minutes. Images were pushed to registry in same location, in case target registry is external (for example quay) the execution time will extend.

Currently there are roughly 48 examples available in examples repo. In case we would want to build all of them (without native images) it could take two and half hours to build them.
One option would be to split jobs to build springboot and quarkus in parallel, third job for native images.

README.md
```diff
@@ -219,6 +219,7 @@ You can set those optional keys:
 - `container_engine` engine used to interact with images and local containers.
   *Default is docker.*
 - `domain_suffix` domain suffix used for exposed services. Ignored when running tests on Openshift.
+- `image_cache_mode` Use this option to specify whether you want to use image cache for runtime images. Available options are always, never of if-available(default).
```
radtriste, line 222:
```suggestion
- `image_cache_mode` Use this option to specify whether you want to use image cache for runtime images. Available options are 'always', 'never' or 'if-available'(default).
```

hack/run-tests.sh
```diff
@@ -49,6 +49,7 @@ function usage(){
   printf "\n--load_default_config\n\tTo be used if you want to directly use the default test config contained into ${SCRIPT_DIR}/../test/.default_config"
   printf "\n--container_engine\n\tTo be used if you want to specify engine to interact with images and containers. Default is docker."
   printf "\n--domain_suffix\n\tTo be used if you want to set a domain suffix for exposed services. Ignored when running tests on Openshift."
+  printf "\n--image_cache_mode\n\tUse this option to specify whether you want to use image cache for runtime images. Available options are always, never of if-available(default)."
```
radtriste, line 52:
```suggestion
  printf "\n--image_cache_mode\n\tUse this option to specify whether you want to use image cache for runtime images. Available options are 'always', 'never' or 'if-available'(default)."
```

test/config/config.go
```diff
@@ -108,6 +110,7 @@ func BindFlags(set *flag.FlagSet) {
 	set.BoolVar(&env.crDeploymentOnly, prefix+"cr-deployment-only", false, "Use this option if you have no CLI to test against. It will use only direct CR deployments.")
 	set.StringVar(&env.containerEngine, prefix+"container-engine", defaultContainerEngine, "Engine used to interact with images and local containers.")
 	set.StringVar(&env.domainSuffix, prefix+"domain-suffix", "", "Set the domain suffix for exposed services. Ignored when running tests on Openshift.")
+	set.StringVar(&env.imageCacheMode, prefix+"image-cache-mode", "if-available", "Use this option to specify whether you want to use image cache for runtime images. Available options are always, never of if-available(default).")
```
radtriste, line 113:
```suggestion
	set.StringVar(&env.imageCacheMode, prefix+"image-cache-mode", "if-available", "Use this option to specify whether you want to use image cache for runtime images. Available options are 'always', 'never' or 'if-available'(default).")
```

test/scripts/examples/process-quarkus-example.feature
```diff
@@ -0,0 +1,17 @@
+@quarkus
+Feature: Build process-quarkus-example images
+
+  Background:
+    Given Clone Kogito examples into local directory
+
+  Scenario: Build process-quarkus-example images
+    Then Local example service "process-quarkus-example" is built by Maven using profile "default" and deployed to runtime registry
+    And Local example service "process-quarkus-example" is built by Maven using profile "persistence" and deployed to runtime registry
+    And Local example service "process-quarkus-example" is built by Maven using profile "events" and deployed to runtime registry
```
radtriste, line 10: do we need "only-events" example ?

test/scripts/examples/process-springboot-example.feature
```diff
@@ -0,0 +1,10 @@
+@springboot
+Feature: Build process-springboot-example images
+
+  Scenario: Build process-springboot-example images
+    Given Clone Kogito examples into local directory
+
+    Then Local example service "process-springboot-example" is built by Maven using profile "default" and deployed to runtime registry
+    And Local example service "process-springboot-example" is built by Maven using profile "persistence" and deployed to runtime registry
+    And Local example service "process-springboot-example" is built by Maven using profile "events" and deployed to runtime registry
```
radtriste, line 9: Same here

test/config/config.go
```diff
@@ -193,6 +196,15 @@ func GetDomainSuffix() string {
 	return env.domainSuffix
 }
 
+// GetImageCacheMode returns image cache mode
+func GetImageCacheMode() ImageCacheMode {
+	imageCacheMode := ImageCacheMode(env.imageCacheMode)
+	if imageCacheMode.isValid() {
+		return imageCacheMode
+	}
+	panic(fmt.Errorf("Invalid image cache mode: %s", imageCacheMode))
```
Sgitario, line 205: I think this will fail when the test is running. In order to make it fail the sooner the better, we could check this in the [check_setup.go](https://github.com/kiegroup/kogito-cloud-operator/blob/master/test/framework/check_setup.go)

test/steps/image_registry.go
```diff
@@ -70,6 +76,23 @@ func (data *Data) localServiceBuiltByMavenWithProfileAndDeployedToRuntimeRegistr
 	return nil
 }
 
+// Returns true if the image has to be built
+func needToBuildImage(namespace, imageTag string) bool {
+	switch config.GetImageCacheMode() {
+	case config.UseImageCacheAlways:
+		return false
```
Sgitario, line 83: So, if we set image cache mode to always and the image is not built, it will fail. However, this is fine and expected, it would be a wrong setup rather.

test/steps/image_registry.go
```diff
@@ -40,7 +40,7 @@ func (data *Data) localServiceBuiltByMavenWithProfileAndDeployedToRuntimeRegistr
 		return err
 	}
 
-	if err := framework.GetContainerEngine(data.Namespace).PullImage(imageTag).GetError(); err != nil {
+	if needToBuildImage(data.Namespace, imageTag) {
 		// Not found in registry, so we need to build and push the application
 		// Build the application
 		err = data.localServiceBuiltByMavenWithProfile(contextDir, profile)
```
Sgitario, line 46: This might be unrelated to these changes, but I think the image built does not take into account the parameters: **build_image_version** and **build_image_name_suffix**. 

These parameters are used to take the docker image base from, but using the image cache this issue might be worse as we will be reusing an image with a base image, when I asked to use another base image. Probably, we could amend [this line](https://github.com/kiegroup/kogito-cloud-operator/blob/25122c0e2a0236357a8b6d8dd2f940db5465cf31/test/steps/image_registry.go#L109) to use the value from [GetBuildImage](https://github.com/kiegroup/kogito-cloud-operator/blob/25122c0e2a0236357a8b6d8dd2f940db5465cf31/test/framework/util.go#L222) instead of latest.

test/scripts/examples/process-quarkus-example.feature
sutaakar: Currently no, I put it there to have all the combinations created.

test/steps/image_registry.go
sutaakar: Right.

test/scripts/examples/process-quarkus-example.feature
sutaakar: I can remove it

test/steps/image_registry.go
sutaakar: good point, will apply

sutaakar: Well, when I think about it, it may have more sense to create a separate `runtime_application_image_version` and `runtime_application_image_name_suffix` to keep consistency of the configuration (as we already have `runtime_application_image_registry` and `runtime_application_image_namespace`)

sutaakar:
@Sgitario @radtriste Please take a look again, the comments should be addressed.

PR #423, opened 2020-07-06
[KOGITO-2650] Setup retry HTTP calls for tests

radtriste:
https://issues.redhat.com/browse/KOGITO-2650

Mainly to avoid Openshift and network stabilities.


test/framework/http.go
```diff
@@ -72,7 +74,27 @@ func WaitForSuccessfulHTTPRequest(namespace string, requestInfo HTTPRequestInfo,
...
-	return ExecuteHTTPRequestC(&http.Client{}, namespace, requestInfo)
+	// Setup a retry in case the first time it did not work
+	retry := 0
+	var resp *http.Response
+	var err error
+	for retry < config.GetHTTPRetryNumber() {
+		resp, err := ExecuteHTTPRequestC(&http.Client{}, namespace, requestInfo)
+		if err == nil && checkHTTPResponseSuccessful(namespace, resp) {
+			return resp, err
+		} else if resp.StatusCode == 500 {
```
Sgitario, line 85: Looking in ExecuteHTTPRequestC, it seems that the response could return null in the response with an error happened.   (Scenario when err is not null and resp is null). So I guess, we should either do the else if err != nil first, or check whether resp is not nil here.

Makefile
```diff
@@ -180,6 +181,7 @@ run-tests:
 		--container_engine ${container_engine} \
 		--domain_suffix ${domain_suffix} \
 		--image_cache_mode ${image_cache_mode} \
+		--http_retry_nb ${http_retry_nb} \
```
Sgitario, line 184: I guess this should be covered in the bats tests too.

test/framework/http.go
radtriste: good catch. will update

test/framework/http.go
```diff
@@ -72,7 +74,30 @@ func WaitForSuccessfulHTTPRequest(namespace string, requestInfo HTTPRequestInfo,
...
+		if err == nil && checkHTTPResponseSuccessful(namespace, resp) {
+			return resp, err
+		} else if err != nil {
+			GetLogger(namespace).Warnf("Error while making http call: %v", err)
+		} else if resp != nil {
+			if resp.StatusCode == 500 {
+				// In case of 500, it is server error and we don't need to call again
+				return resp, err
+			}
+			GetLogger(namespace).Warnf("Http call was not successful. Response code: %d", resp.StatusCode)
```
MarianMacik, line 92: Maybe we could first log the message and then in case it is 500, return. Now it won't log the code if it is 500.

radtriste: updated

PR #432, opened 2020-07-07
[KOGITO-2688] Removing listType=map from non-required objects

ricardozanini:

See:
https://issues.redhat.com/browse/KOGITO-2688

This needs to be on 0.12.0, otherwise users will fail to install all the CRDs in Kubernetes 1.18, like the newer minikube.

### DIFF
--- a/deploy/crds/app.kiegroup.org_kogitoapps_crd.yaml
+++ b/deploy/crds/app.kiegroup.org_kogitoapps_crd.yaml
@@ -273,9 +273,7 @@ spec:
                         type: string
                     type: object
                   type: array
-                  x-kubernetes-list-map-keys:
-                  - type
-                  x-kubernetes-list-type: map
+                  x-kubernetes-list-type: atomic
               type: object
             deploymentLabels:
               additionalProperties:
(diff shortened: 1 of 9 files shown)


ricardozanini:
> I always have one question in mind, Why we are commiting auto-generated file in Github? Generally we don't do so

Those files are needed for the whole project to work since they implement necessary functions for our inner types to work with Kubernetes. In Golang this is pretty common to generate boilerplate functions to extend some functionality.

vaibhavjainwiz:
> > I always have one question in mind, Why we are commiting auto-generated file in Github? Generally we don't do so
> 
> Those files are needed for the whole project to work since they implement necessary functions for our inner types to work with Kubernetes. In Golang this is pretty common to generate boilerplate functions to extend some functionality.

Sorry If I am not clear with my last statement, I am not asking to remove these files from project. I just checking that because these are auto-generated files and ideally developer should not suppose to make any change in these files directly so there is no need to checkin these files in GitHub repo. These will be auto generates when Jenkins perform its build.

With this approach we prevent lots of issue where developer miss to compile/commit these files. Or Develeloper have different version of Operator-sdk on his system which cause him to generate wrong files

ricardozanini:
We need those files in the repository, otherwise our types won't work and will fail in case of anyone uses the package. It's not a matter of generating those files for runtime only.

PR #446, opened 2020-07-14
KOGITO-2738 - Stop codecov bot when there is change in files which do…

Kaitou786:
…esn't change the coverage
See: https://issues.redhat.com/browse/KOGITO-2738

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,5 @@
+comment:
+  require_changes: true
 flags:
   operator:
     paths:


ricardozanini:
@Kaitou786 it's better to elect which pattern will get the unit test run:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns

If we only change the bdd tests, documentation or examples yaml files, we might not have to run it.

Kaitou786:
Yes, I was thinking that, but in that case we'll have to use a separate workflow for the unit-tests As they currently don't have it for per job or steps. https://github.community/t/path-filtering-for-jobs-and-steps/16447. If you think this is good idea then I can send a PR for this @ricardozanini

ricardozanini:
Yeah, I believe we need to separate them like we talked yesterday. Go ahead! :)

We will save lots of resources and speed up the bdd-tests PRs.

PR #454, opened 2020-07-15
[KOGITO-2784] - Bump to 0.13.0

ricardozanini:

Let's try again.


ricardozanini:
:shrug: looks like GH is having problems.

Kaitou786:
Not just them but our Jenkins CI is also not working, maybe something wrong with github again =(

README.md
```diff
@@ -109,7 +109,7 @@ account to create an application repository.
 
 Follow the steps below:
 
-1. Run `make prepare-olm version=0.12.0`. Bear in mind that if there's different versions
+1. Run `make prepare-olm version=0.13.0`. Bear in mind that if there's different versions
 in the `deploy/olm-catalog/kogito-operator/kogito-operator.package.yaml` file, every CSV must
 be included in the output folder. At this time, the script did not copy previous CSV versions to the
 output folder, so it must be copied manually.
```
Kaitou786, line 115: Meybe we can remove this from README, as now all the manifests will be present after running the `make prepare-olm`?

ricardozanini: +1

ricardozanini: I'll change once GH gets back.

ricardozanini: Actually this is needed to verify the CSV in a local cluster. I'll keep it.

Kaitou786: Yes, the command should be there but what I meant to be remove is this:
 ```
 every CSV must
be included in the output folder. At this time, the script did not copy previous CSV versions to the
output folder, so it must be copied manually.
```
As after running this we won't need to copy anything manually.

File: hack/bump-version.sh [hunk 2]
```diff
@@ -29,10 +29,8 @@ if [ -z "$new_version" ]; then
 fi
 
 sed -i "s/$old_version/$new_version/g" cmd/kogito/version/version.go README.md version/version.go deploy/operator.yaml deploy/olm-catalog/kogito-operator/kogito-operator.package.yaml hack/go-build.sh hack/go-vet.sh .osdk-scorecard.yaml
-operator-sdk generate csv --csv-version "$new_version" --from-version "$old_version"  --update-crds --operator-name kogito-operator
+operator-sdk generate csv --apis-dir ./pkg/apis/apps/v1alpha1 --verbose --csv-version "$new_version" --from-version "$old_version"  --update-crds --operator-name kogito-operator
 
-# rewrite test default config, all other configuration into the file will be overridden
-test_config_file="test/.default_config"
```
sutaakar, line 35: Why is this code removed?

ricardozanini: because was updating the config file. I'm assuming we don't need that anymore since it's pointing to `latest`. Should we update it every time? If so, I think we could review this approach.

sutaakar: ok, thanks
I guess we can remove it

radtriste: This code is important if we use release branches ...
anyway updated in for the future in https://github.com/kiegroup/kogito-cloud-operator/pull/420/files#diff-7a9d206eb5b9303a3c9c293020c798daR47-R50

deploy/olm-catalog/kogito-operator/0.13.0/kogito-operator.v0.13.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,957 @@
...
+  - cloud
+  - kogito
+  - quarkus
+  - springboot
+  - operator
+  - native
+  - automation
+  labels:
+    alm-owner-kogito: kogito-operator
+    operated-by: kogito-operator.0.11.0
```
radtriste, line 942: This needs to be updated as well ...
can have a look here: https://github.com/kiegroup/kogito-cloud-operator/pull/420/files#diff-7a9d206eb5b9303a3c9c293020c798daR33

ricardozanini: I've updated this in the other PR, thanks GitHub for messing with my commits.

hack/bump-version.sh
ricardozanini: Ok, but aren't you guys changing this configuration all the time? That's what I meant. Should I put it back or not?

ricardozanini:
@radtriste updated. We just need to resolve this test configuration thing. :D

hack/bump-version.sh
radtriste: the `test/.default_config` needs to be there as it is used by the PR checks for example.
it allows also to not set everything each time.
For a release and/or a release branch, we need to change to the correct versions so we are sure the default config is pointing to the right images.

ricardozanini: Ok, I'll put it back. The `bump-script` changed it to 0.13.0, was latest. Should I keep it as 0.13.0 or latest? Is this the final solution? I mean, we should update this config file every time?

radtriste: if you are on `master`, it should set to `latest` automatically: https://github.com/kiegroup/kogito-cloud-operator/pull/454/files#diff-7a9d206eb5b9303a3c9c293020c798daL39

radtriste: ah but you are not ... that is why XD
need to update the script accordingly then. will do that.
`latest` should be set, also some parameters are missing there. I thought I modified the script already ...
will PR today for the update

radtriste: https://issues.redhat.com/browse/KOGITO-2795

ricardozanini: that's why I removed these lines. The test config is correctly set to latest. So I believe I have to do any thing here?

radtriste: let it like the default_config is and do not remove the lines. I will update the script.

ricardozanini:
Waiting for the CI.

PR #469, opened 2020-07-24
Mocking approach for testing

vaibhavjainwiz:

As discussed in last meeting, I am sharing mocking approach with testity framework.
We are already using `testify` framework in our project to write assert in unit test cases. I had implement mocking services of `testify` framework to write test cases with mocking of another service.

With this approach we can cover every scenerio in our project. 

### DIFF
--- a/cmd/kogito/command/deploy/delete_service_test.go
+++ b/cmd/kogito/command/deploy/delete_service_test.go
@@ -17,9 +17,13 @@ package deploy
 import (
 	"fmt"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/context"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/service"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/shared"
 	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/test"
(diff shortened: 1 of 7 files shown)


sutaakar:
I can see a meaning for mocking in case we want to simulate some specific conditions/issues like failure in Kubernetes API call, something what is hard to simulate with current functionality. Not much for "common" usage.

Mocking would allow us to run unit tests for "high level" structures, by isolating it from lower levels. But do we want it?
IMHO such approach would bind the tests with the implementation, in case of refactoring we would have to basically rewrite the tests. 
We would still have to keep some functional tests to make sure that application layers can work together, making sure that upper layer is able to process response from lower layer.

Unit tests have sense in case of simple functions without much dependencies and integrations. But for complex functionality it is from my point of view important to make sure it follows contract and correctly integrates with all components.

ricardozanini:
Agree with @sutaakar, IMO I don't see why mocking these internals on command side. The main problem that we'd like to test is actually to see if the flags are been converted into the correct CR structure. Mocking this behavior would not test what we really need.

ricardozanini:
@vaibhavjainwiz can we close this one?

vaibhavjainwiz:
> @vaibhavjainwiz can we close this one?

Please close it

PR #490, opened 2020-08-04
[KOGITO-2932] Install operators one after the other

radtriste:
https://issues.redhat.com/browse/KOGITO-2932

Due to https://github.com/operator-framework/operator-lifecycle-manager/issues/1704

### DIFF
--- a/test/steps/operator.go
+++ b/test/steps/operator.go
@@ -52,7 +52,7 @@ func (data *Data) kogitoOperatorIsDeployed() error {
 }
 
 func (data *Data) kogitoOperatorIsDeployedWithDependencies(dependencies string) error {
-	// Install kogito operator
+	// First install and wait for kogito operator
 	if exists, err := framework.IsKogitoOperatorRunning(data.Namespace); err != nil {
 		return fmt.Errorf("Error while trying to retrieve the operator: %v ", err)
 	} else if !exists {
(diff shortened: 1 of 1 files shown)


radtriste:
Jenkins test this

radtriste:
also tested the full testing, no more `Timeout waiting for operator`

radtriste:
Jenkins retest this

radtriste:
@Sgitario hard to tell because some other tests failed because of other reasons ... but timeout for Strimzi operator is 30 min (load_factor is 3 by default) so each time there is a problem on operator, then it waits 30 min before failing.
I would expect the time installing operators taking less than 30 min overall ;)

But I can make it configurable if you prefer, with sequential by default for now. wdyt ?

radtriste:
Jenkins retest this

radtriste:
Jenkins test this

Sgitario:
> @Sgitario hard to tell because some other tests failed because of other reasons ... but timeout for Strimzi operator is 30 min (load_factor is 3 by default) so each time there is a problem on operator, then it waits 30 min before failing.
> I would expect the time installing operators taking less than 30 min overall ;)
> 
> But I can make it configurable if you prefer, with sequential by default for now. wdyt ?

If it would only work sequentially, then it's fine to not being configurable. But if the all at once installation works for some scenario, then I would expect to be configurable.

radtriste:
well this an olm issue and at schrödinger installation: it may or may not work ...
it is a problem if number of dependent operators >= 2. I could change to check that.

test/steps/operator.go
```diff
@@ -61,32 +61,30 @@ func (data *Data) kogitoOperatorIsDeployedWithDependencies(dependencies string)
...
-			installedOperators = append(installedOperators, dependentOperator)
+			if err := framework.WaitForKogitoOperatorDependencyRunning(data.Namespace, dependentOperator, operatorSource); err != nil {
+				return err
+			}
 		}
 	}
 
-	// Wait for kogito operator to be running
+	// Wait for Kogito operator running
+	// This is put at the end because there should not be any racing condition as install is done via yaml and not OLM
```
radtriste, line 84: @Sgitario I let this one at the end as there should not be any conflict. no need to wait first for Kogito Operator running

radtriste:
Jenkins test this

PR #507, opened 2020-08-10
[KOGITO-3030] Tests: Retry image build in case of failure

sutaakar:

https://issues.redhat.com/browse/KOGITO-3030

### DIFF
--- a/test/framework/container_engine.go
+++ b/test/framework/container_engine.go
@@ -103,7 +103,7 @@ func (containerEngine *containerEngineStruct) PushImage(imageTag string) Contain
 
 func (containerEngine *containerEngineStruct) BuildImage(projectLocation, imageTag string) ContainerEngine {
 	if containerEngine.err == nil {
-		_, containerEngine.err = CreateCommand(containerEngine.engine, containerEngine.buildCommand, "--tag", imageTag, ".").InDirectory(projectLocation).WithLoggerContext(containerEngine.namespace).Execute()
+		_, containerEngine.err = CreateCommand(containerEngine.engine, containerEngine.buildCommand, "--tag", imageTag, ".").InDirectory(projectLocation).WithLoggerContext(containerEngine.namespace).WithRetry(NumberOfRetries(3)).Execute()
 	}
 	return containerEngine
 }
--- a/test/framework/command.go
+++ b/test/framework/command.go
@@ -16,6 +16,7 @@ package framework
 
 import (
 	"os/exec"
+	"time"
 
 	"go.uber.org/zap"
 )
@@ -31,6 +32,8 @@ type Command interface {
 	WithLoggerContext(loggerContext string) Command
 	// InDirectory method sets the directory where the command will be executed.
 	InDirectory(directory string) Command
+	// WithRetry method defines retry options to be applied to the command.
+	WithRetry(opts ...RetryOption) Command
 	// Execute command and returns the outputs.
 	Execute() (string, error)
 }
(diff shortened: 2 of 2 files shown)


sutaakar:
Jenkins test this

test/framework/container_engine.go
```diff
@@ -103,7 +103,13 @@ func (containerEngine *containerEngineStruct) PushImage(imageTag string) Contain
...
 func (containerEngine *containerEngineStruct) BuildImage(projectLocation, imageTag string) ContainerEngine {
 	if containerEngine.err == nil {
-		_, containerEngine.err = CreateCommand(containerEngine.engine, containerEngine.buildCommand, "--tag", imageTag, ".").InDirectory(projectLocation).WithLoggerContext(containerEngine.namespace).Execute()
+		retries := 3
+		for i := 0; i < retries; i++ {
+			_, containerEngine.err = CreateCommand(containerEngine.engine, containerEngine.buildCommand, "--tag", imageTag, ".").InDirectory(projectLocation).WithLoggerContext(containerEngine.namespace).Execute()
+			if containerEngine.err == nil {
+				return containerEngine
+			}
+		}
```
Sgitario, line 112: If this is a long term solution, I would implement the "Retry" function as part of the CreateCommand framework (see line [here](https://github.com/kiegroup/kogito-cloud-operator/blob/3613717d8111cb5f696f9235ee3e5501aebea3f9/test/framework/command.go#L67)). This way we can reuse this feature in other commands in the future. The usage would be simply:

```java
CreateCommand(...).InDirectory(...).WithRetry(3 times)
```

What do you think?
```

Comment (sutaakar) [reply test/framework/container_engine.go:112]: hmm, good point.
will extend the command executor.

**Comment by sutaakar** [comment]
I have separated retries functionality configuration into separate interface to keep Command interface not polluted by various retry configuration options.

File: test/framework/container_engine.go [hunk 2]
```diff
@@ -103,7 +103,7 @@ func (containerEngine *containerEngineStruct) PushImage(imageTag string) Contain
 
 func (containerEngine *containerEngineStruct) BuildImage(projectLocation, imageTag string) ContainerEngine {
 	if containerEngine.err == nil {
-		_, containerEngine.err = CreateCommand(containerEngine.engine, containerEngine.buildCommand, "--tag", imageTag, ".").InDirectory(projectLocation).WithLoggerContext(containerEngine.namespace).Execute()
+		_, containerEngine.err = CreateCommand(containerEngine.engine, containerEngine.buildCommand, "--tag", imageTag, ".").InDirectory(projectLocation).WithLoggerContext(containerEngine.namespace).WithRetry(NumberOfRetries(3)).Execute()
```
radtriste, line 106: Shouldn't we set also RetryDelay ?

sutaakar: IMHO it is not needed in this case. In case of some temporary issue I think that we can retry immediately.
RetryDelay was introduced as an additional functionality for command, not needed in this case.

radtriste: ok so if no value then the sleep is not happening right ?

sutaakar: Yes

ricardozanini:
@Kaitou786 since this is a change only on `test` path, should Filippe and I also be included in the review process?

Kaitou786:
I noticed this, the current deployment should not have this, but this got triggered as the change in the bot image did not automatically triggered the deployment with latest code in image(I thought it would) so I manually triggered the deployment which now has the latest code and this thing solved i.e. default reviewers won't be requested in case path reviewers exist.

PR #515, opened 2020-08-12
[KOGITO-3050] - Fixing from file build trigger

ricardozanini:

See: https://issues.redhat.com/browse/KOGITO-3050

In this PR we fix a bug when the server does not have a BC ready for a build to start, also hacked around the `FakeBuildV1` to have a REST client call.


ricardozanini:
> Change detected in the PR, requesting reviews and running pipeline(if required) again

How dare you? You don't have permissions. Let me do it.

ricardozanini:
@sutaakar looks like 1 minute to wait for the BC to be there in the server was too low on a real server. I increased the timeout to 5 min and never was able to replicate the problem again, but it looks like that can only be replicated on environments which connections between client and server are somewhat slow. I don't have  an env like this to replicate, could you please take a look if this last PR fixed the issue? 

Besides, there was a bug in the build_service.go file that was swallowing this error. At least now we will have the right output.

Also, I'll merge https://issues.redhat.com/browse/KOGITO-3063 EOD with @spolti so you would have more data to look into.

ricardozanini:
depends on #516 for debug reasons.

PR #519, opened 2020-08-13
KOGITO-3084 - Add first version to deploy a prometheus and a grafana operators

r00ta:
Still work in progress, not ready to be reviewed. Description to be added

- [ ] You have read the [contributors' guide](https://github.com/kiegroup/kogito-cloud-operator/blob/master/README.md#contributing-to-the-kogito-operator)


ricardozanini:
@r00ta  I know that it's a draft, but  I'm a little concerned with this use case. I don't think we will deploy any operators. This should be a design discussion with the runtimes team as well. We are about to actually **drop** any dependant operators. 

Also, the platform side will be handled in an agnostic way on: https://issues.redhat.com/browse/KOGITO-3039 (this is just for persistence for now, but we will follow for others in the future) So the infra controller could understand which resource type to deploy. The operator itself should be provisioned and deployed by the user.

Please bring this UC to the runtime team, let's have a wider conversation about it. cc @spolti

PR #541, opened 2020-08-31
KOGITO-3058: Improve DEBUG option

Sgitario:
JIRA: https://issues.redhat.com/browse/KOGITO-3058
Description: Debug log of test run is not appearing

### DIFF
--- a/test/framework/logging.go
+++ b/test/framework/logging.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/logger"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/util"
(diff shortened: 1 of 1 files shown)


Sgitario:
I've tested these changes and now the debug traces from the steps are appearing. But when running in debug mode, the output is not in JSON, I can easily change this but it would imply changes in the operator framework as well and maybe this was done without json on purpose. What do you think @sutaakar ?

Sgitario:
@sutaakar another question, the JIRA says that the command output should be displayed too: "+ set command execution in debug mode as well", but I think it was already working fine (the only problem was that the debug traces were hidden). Am I missing something here?

sutaakar (approved):

sutaakar:
Debug output in plaintext is ok for me.

sutaakar:
About command output - it seems to be implemented already, I think we can consider it as implemented.

PR #550, opened 2020-09-10
[KOGITO-3039] - Agnostic infrastructure layer configuration on Operator

vaibhavjainwiz:
Jira issue : https://issues.redhat.com/browse/KOGITO-3039

As a part of KOGITO-3039, we had seperated infra details from KogitoService CR. Previously infra(Infinispan/kafka) was initialize at the time of KogitoService creation but now user have to setup kogitoInfra before KogitoService initialization.

Previous Implementation:
 ```
kind: KogitoDataIndex
metadata:
  name: data-index
spec:
  replicas: 1
  image:
    tag: "0.12"
  kafka:
    useKogitoInfra: true
  infinispan:
    useKogitoInfra: true
```
New implementation:
```
kind: KogitoInfra
metadata:
  name: kogito-infinispan-infra
spec:
  resource:
    apiVersion: infinispan.org/v1 
    kind: Infinispan
---
kind: KogitoInfra
metadata:
  name: kogito-kafka-infra
spec:
  resource:
    apiVersion: kafka.strimzi.io/v1beta1
    kind: Kafka
---
kind: KogitoDataIndex
metadata:
  name: kogitodataindex
spec:
  replicas: 1
  image:
    domain: quay.io
    namespace: kiegroup
    name: kogito-data-index
    tag: "0.12"
  infra:
    - kogito-infinispan-infra
    - kogito-kafka-infra
```

User could also initialize KogitoInfra using Kogito CLI:
```
kogito install infra kogito-infinispan-infra --kind Infinispan --apiVersion infinispan.org/v1
kogito install infra kogito-kafka-infra --kind Kafka --apiVersion kafka.strimzi.io/v1beta1
kogito install data-index --image quay.io/kiegroup/kogito-data-index:0.12 --infra kogito-infinispan-infra --infra kogito-kafka-infra
```

User could leverage option to allow kogito-operator to setup Inifinispan and kafka instance for him. To allow it user should pre-install Inifinispan-operator and Strimzi-operator in same context as kogito-operator. Or User could could provide its references of already install infra. If user don't provide name and namespace of resource then kogito operator will try to setup infra for user otherwise user could provide name and namespace of custom infra and kogito-operator will use these details to connect with user provides infra like below :
```
kind: KogitoInfra
metadata:
  name: kogito-infinispan-infra
spec:
  resource:
    apiVersion: infinispan.org/v1 
    kind: Infinispan
    name: custom-infinispan-instance
    namespace: custom-namespace  
```

Same options are also provided in Kogito-cli:
```
kogito install infra kogito-infinispan-infra --kind Infinispan --apiVersion infinispan.org/v1 --resource-name custom-infinispan-instance --resource-namespace custom-namespace
```

### CODE REVIEW

**Comment by ricardozanini** [comment]
@vaibhavjainwiz do you mind taking a look in the PR checks failures?

File: pkg/apis/app/v1alpha1/kogitoinfra_types.go [hunk 1]
```diff
@@ -19,6 +19,19 @@ import (
...
+
+	// Kind describe the kind of referred resource for example, app.infinispan.org/v1
+	Kind string `json:"kind,omitempty"`
+
+	// Namespace where referred resource exists
+	Namespace string `json:"namespace,omitempty"`
+
+	// Name of referred resource
+	Name string `json:"name,omitempty"`
+}
```
ricardozanini, line 33: @vaibhavjainwiz I believe we would need the `apiVersion` attribute here as well, with is the `app.infinispan.org/v1`. So the interface should be something like:

```suggestion
// Resource provide reference infra resource
type Resource struct {
	 // ApiVersion describes the API Version of referred Kubernetes resource for example, app.infinispan.org/v1
   	ApiVersion string `json:"apiVersion"`

	// Kind describes the kind of referred Kubernetes resource for example, Infinispan
	Kind string `json:"kind"`

	// Namespace where referred resource exists, default to the current namespace
	Namespace string `json:"namespace,omitempty"`

	// Name of referred resource
	Name string `json:"name"`
}
```

Also I've removed the `omitempty` attribute from the required properties.

pkg/apis/app/v1alpha1/kogitoinfra_types.go
```diff
@@ -56,7 +74,7 @@ type KogitoInfraSpec struct {
 type KogitoInfraStatus struct {
 	Condition  KogitoInfraCondition            `json:"condition,omitempty"`
 	Infinispan InfinispanInstallStatus         `json:"infinispan,omitempty"`
-	Kafka      InfraComponentInstallStatusType `json:"kafka,omitempty"`
+	Kafka      KafkaInstallStatus              `json:"kafka,omitempty"`
```
ricardozanini, line 77: We should remove any mention to implementation from our interfaces. The status should inform the user if the `KogitoInfra` resource was able to fetch the given third party resource and use that information to configure the linked KogitoRuntimes.

pkg/controller/kogitoinfra/infinispan/infinispan_properties.go
```diff
@@ -0,0 +1,108 @@
...
+		AppPropInfinispanUseAuth:       "infinispan.remote.use-auth",
+		AppPropInfinispanSaslMechanism: "infinispan.remote.sasl-mechanism",
+		AppPropInfinispanAuthRealm:     "infinispan.remote.auth-realm",
+
+		EnvVarInfinispanUser:     "INFINISPAN_REMOTE_AUTH_USERNAME",
+		EnvVarInfinispanPassword: "INFINISPAN_REMOTE_AUTH_PASSWORD",
+	}
+)
+
+// FetchInfraProperties provide application/env properties of infra that need to be ser in kogitoRuntime object
```
ricardozanini, line 75:
```suggestion
// FetchInfraProperties provide application/env properties of infra that need to be set in the KogitoRuntime object
```

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+func (i *InfinispanResource) Reconcile(client *client.Client, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) (requeue bool, resultErr error) {
+
+	var infinispanInstance *infinispanv1.Infinispan
+
+	// Step 1: check whether user has provided custom infinispan instance reference
+	isCustomReferenceProvided := len(instance.Spec.Resource.Name) > 0
+	if isCustomReferenceProvided {
+		log.Debugf("Custom infinispan instance reference is provided")
+		resourceName := instance.Spec.Resource.Name
+		resourceNameSpace := instance.Spec.Resource.Namespace
```
ricardozanini, line 57: Why we need these temps? `instance.Spec.Resource.Name` won't work?

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+			return false, fmt.Errorf("custom Infinispan instance(%s) not found in namespace %s", resourceName, resourceNameSpace)
+		}
+	} else {
+		// create/refer kogito-infinispan instance
+		log.Debugf("Custom infinispan instance reference is not provided")
+		// if resource name is not provided then Infinispan instance should be created with default name = kogito-infinispan
+		resourceName := InstanceName
+
+		// if resource name is not provided then Infinispan instance should be created with default name = kogito-infinispan
+		resourceNameSpace := instance.Namespace
```
ricardozanini, line 73: I think we can use the provided variables instead of creating temps.

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	infinispanProperties := &instance.Status.Infinispan.InfinispanProperties
+	infinispanProperties.URI = infinispanURI
+	infinispanProperties.Credentials.SecretName = customInfinispanSecret.Name
+	infinispanProperties.Credentials.UsernameKey = infrastructure.InfinispanSecretUsernameKey
+	infinispanProperties.Credentials.PasswordKey = infrastructure.InfinispanSecretPasswordKey
+
+	log.Debugf("Updating kogitoInfra(%s) value with new properties", instance.Name)
+	if resultErr = kubernetes.ResourceC(client).Update(instance); resultErr != nil {
```
ricardozanini, line 127: Is this updating the status of the instance, right? Why we need this here? I didn't see a change in the `instance` attributes in this function.

pkg/controller/kogitoinfra/kafka/kafka_reconciler.go
```diff
@@ -0,0 +1,108 @@
...
+func (k *KafkaResource) Reconcile(client *client.Client, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) (requeue bool, resultErr error) {
+
+	var kafkaInstance *kafkav1beta1.Kafka
+
+	// Step 1: check whether user has provided custom infinispan instance reference
+	isCustomReferenceProvided := len(instance.Spec.Resource.Name) > 0
+	if isCustomReferenceProvided {
+		log.Debugf("Custom kafka instance reference is provided")
+		resourceName := instance.Spec.Resource.Name
+		resourceNameSpace := instance.Spec.Resource.Namespace
```
ricardozanini, line 53: Same here about the temps

pkg/controller/kogitoinfra/kafka/kafka_reconciler.go
```diff
@@ -0,0 +1,108 @@
...
+			return false, fmt.Errorf("custom kafka instance(%s) not found in namespace %s", resourceName, resourceNameSpace)
+		}
+	} else {
+		// create/refer kogito-kafka instance
+		log.Debugf("Custom kafka instance reference is not provided")
+		// if resource name is not provided then Infinispan instance should be created with default name = kogito-infinispan
+		resourceName := InstanceName
+
+		// if resource name is not provided then Infinispan instance should be created with default name = kogito-infinispan
+		resourceNameSpace := instance.Namespace
```
ricardozanini, line 69: Same here regarding the temp vars.

pkg/controller/kogitoinfra/kafka/kafka_reconciler.go
```diff
@@ -0,0 +1,108 @@
...
+	uri, resultErr := infrastructure.GetKafkaServerURI(kafkaInstance.Name, kafkaInstance.Namespace, client)
+	if resultErr != nil {
+		return false, resultErr
+	}
+
+	kafkaProperties := &instance.Status.Kafka.KafkaProperties
+	kafkaProperties.ExternalURI = uri
+
+	log.Debugf("Updating kogitoInfra(%s) value with new properties", instance.Name)
+	if resultErr = kubernetes.ResourceC(client).Update(instance); resultErr != nil {
```
ricardozanini, line 102: Same question regarding the need to update the instance here. I believe you can perform this update in the main controller after updating the instance status and returning a reference of it to the main controller.

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Reconcile reconcile Kogito infra object
+func (k *KeycloakResource) Reconcile(client *client.Client, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) (requeue bool, resultErr error) {
+
+	var keycloakInstance *keycloakv1alpha1.Keycloak
+
+	// Step 1: check whether user has provided custom keycloak instance reference
+	isCustomReferenceProvided := len(instance.Spec.Resource.Name) > 0
+	if isCustomReferenceProvided {
+		log.Debugf("Custom kafka instance reference is provided")
```
ricardozanini, line 51:
```suggestion
		log.Debugf("Custom Keycloak instance reference is provided")
```

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
```diff
@@ -0,0 +1,93 @@
...
+
+		keycloakInstance, resultErr = loadDeployedKeycloakInstance(client, resourceName, resourceNameSpace)
+		if resultErr != nil {
+			return false, resultErr
+		}
+		if keycloakInstance == nil {
+			return false, fmt.Errorf("custom Keycloak instance(%s) not found in namespace %s", resourceName, resourceNameSpace)
+		}
+	} else {
+		log.Debugf("Custom kafka instance reference is not provided")
```
ricardozanini, line 63:
```suggestion
		log.Debugf("Custom keycloak instance reference is not provided")
```

ricardozanini (changes requested):
Great Job clean up the interface! I have some observations, though:

1. We need kind + group/version in the interface, sorry I missed that in my explanation. Is the way Kuberentes uses to identify a custom resource
2. We should not reference the implementation in our interface even in the status. I'd say to add those properties in a secret (for credentials) or configMap (for general properties), reference them in the status and then make the Kogito services fetch and use/mount them.

I liked the approach to use interfaces to handle the different reconciliation jobs, would be easier to add more infra components in the future.

Many thanks for this PR!

pkg/apis/app/v1alpha1/kogitoinfra_types.go
spolti: +1

pkg/controller/kogitoinfra/infinispan/infinispan_properties.go
```diff
@@ -0,0 +1,108 @@
...
+	// EnvVarInfinispanPassword environment variable for setting infinispan password
+	EnvVarInfinispanPassword
+
+	infinispanEnvKeyCredSecret string = "INFINISPAN_CREDENTIAL_SECRET"
+	enablePersistenceEnvKey           = "ENABLE_PERSISTENCE"
+)
+
+var (
+	/*
+		Infinispan variables for the KogitoInfra deployed infrastructure.
```
spolti, line 48: indentation.

pkg/controller/kogitoinfra/infinispan/resource.go
```diff
@@ -146,3 +150,117 @@ func newInfinispanLinkedSecret(kogitoInfra *v1alpha1.KogitoInfra, cli *client.Cl
...
+	log.Debugf("Fetching %s ", name)
+	secret := &v1.Secret{}
+	if exits, err := kubernetes.ResourceC(cli).FetchWithKey(types.NamespacedName{Name: name, Namespace: namespace}, secret); err != nil {
+		log.Errorf("Error occurs while fetching %s", name)
+		return nil, err
+	} else if !exits {
+		log.Errorf("%s not found", name)
+		return nil, nil
+	} else {
+		log.Errorf("%s fetch successfully", name)
```
spolti, line 213: %s  successfully fetched

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+		if resultErr != nil {
+			return false, resultErr
+		}
+
+		if infinispanInstance == nil {
+			// if not exist then create new Infinispan instance. Infinispan operator creates Infinispan instance, secret & service resource
+			infinispanInstance, resultErr = createNewInfinispanInstance(client, resourceName, resourceNameSpace, instance, scheme)
+			if resultErr != nil {
+				return false, resultErr
+			} // provisioning
```
spolti, line 96: is this comment here correct?

pkg/controller/kogitoinfra/keycloak/keyclock_properties.go
```diff
@@ -0,0 +1,25 @@
...
+
+package keycloak
+
+import (
+	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// FetchInfraProperties provide application/env properties of infra that need to be ser in kogitoRuntime object
+func (k *KeycloakResource) FetchInfraProperties(instance *v1alpha1.KogitoInfra, runtimeType v1alpha1.RuntimeType) (appProps map[string]string, envProps []corev1.EnvVar) {
```
spolti, line 23: not supported yet?

pkg/controller/kogitoinfra/keycloak/resource.go
```diff
@@ -65,3 +71,40 @@ func CreateRequiredResources(kogitoInfra *v1alpha1.KogitoInfra) (resources map[r
...
+			Extensions:     []string{keycloakMetricsExtension},
+			Instances:      1,
+			ExternalAccess: keycloakv1alpha1.KeycloakExternalAccess{Enabled: true},
+		},
+	}
+	if err := controllerutil.SetOwnerReference(instance, keycloakInstance, scheme); err != nil {
+		return nil, err
+	}
+	if err := kubernetes.ResourceC(cli).Create(keycloakInstance); err != nil {
+		log.Error("Error occurs while creating kogito Kafka instance")
```
spolti, line 105: Keycloak

pkg/controller/kogitoinfra/keycloak/resource.go
```diff
@@ -65,3 +71,40 @@ func CreateRequiredResources(kogitoInfra *v1alpha1.KogitoInfra) (resources map[r
...
+		return nil, err
+	}
+	log.Debug("Kogito Kafka instance created successfully")
```
spolti, line 108: Keycloak

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
```diff
@@ -0,0 +1,93 @@
...
+
+// KeycloakResource implementation of KogitoInfraResource
+type KeycloakResource struct {
+}
+
+// GetWatchedObjects provide list of object that needs to be watched to maintain Keycloak kogitoInfra resource
+func GetWatchedObjects() []framework.WatchedObjects {
+	return []framework.WatchedObjects{
+		{
+			GroupVersion: kafkav1beta1.SchemeGroupVersion,
```
spolti, line 36: Keycloak?

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
```diff
@@ -0,0 +1,93 @@
...
+			AddToScheme:  kafkav1beta1.SchemeBuilder.AddToScheme,
+			Objects:      []runtime.Object{&kafkav1beta1.Kafka{}},
```
spolti, line 38: Keycloak?

pkg/infrastructure/infinispan.go
```diff
@@ -22,6 +22,7 @@ import (
 	"k8s.io/api/apps/v1"
```
spolti, line 22: why not group infinispan and other infra on the same package?
pkg/infinispan/{controller,infra}? 
does it make sense?

spolti (changes requested):
Nice work man and sry for the late review.
just a few comments for now.

vaibhavjainwiz:
> Great job cleanin up the interface! I have some observations, though:
> 
> 1. We need kind + group/version in the interface, sorry I missed that in my explanation. Is the way Kuberentes uses to identify a custom resource
> 2. We should not reference the implementation in our interface even in the status. I'd say to add those properties in a secret (for credentials) or configMap (for general properties), reference them in the status and then make the Kogito services fetch and use/mount them.
> 
> I liked the approach to use interfaces to handle the different reconciliation jobs, would be easier to add more infra components in the future.
> 
> Many thanks for this PR!

@ricardozanini 
Property value in config map varies based on `RuntimeType`. For example if runtime is `springboot` then following properties needs to be configure :
```infinispan.remote.server-list
infinispan.remote.use-auth
infinispan.remote.sasl-mechanism```

whereas if runtime is `quarkus` then we need to configure following properties :
```quarkus.infinispan-client.server-list
quarkus.infinispan-client.use-auth
quarkus.infinispan-client.sasl-mechanism```

To tackle this secenerio we can create follow below approaches :
**Approach 1:**
both configmap(for springboot, quarkus) and KogitoRuntime service can decide which config map it want use.

**Approach 2:** 
We can create one config map which contains properties for both runtime type. At runtime application only pick its relative properties.

I would prefer **Approach 2** because I could be applied in future in more broader scenerios. 

If we follow **Approach 1** then we have to maintain collection of `map[RuntimeType]string` and also need to maintain additional code which identify correct config map on bases of RuntimeType.

File: pkg/apis/app/v1alpha1/kogitoinfra_types.go (hunk 1 above)
Comment (vaibhavjainwiz) [reply pkg/apis/app/v1alpha1/kogitoinfra_types.go:33]: done

File: pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go (hunk 6 above)
Comment (vaibhavjainwiz) [reply pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go:127]: We are setting properties in instance attribute:
```
infinispanProperties := &instance.Status.Infinispan.InfinispanProperties
	infinispanProperties.URI = infinispanURI
	infinispanProperties.Credentials.SecretName = customInfinispanSecret.Name
	infinispanProperties.Credentials.UsernameKey = infrastructure.InfinispanSecretUsernameKey
	infinispanProperties.Credentials.PasswordKey = infrastructure.InfinispanSecretPasswordKey
```

File: pkg/controller/kogitoinfra/kafka/kafka_reconciler.go (hunk 7 above)
Comment (vaibhavjainwiz) [reply pkg/controller/kogitoinfra/kafka/kafka_reconciler.go:53]: Similarly we are setting below properties for kafka :
```
        kafkaProperties := &instance.Status.Kafka.KafkaProperties
	kafkaProperties.ExternalURI = uri
```

File: pkg/controller/kogitoinfra/kafka/kafka_reconciler.go (hunk 9 above)
Comment (vaibhavjainwiz) [reply pkg/controller/kogitoinfra/kafka/kafka_reconciler.go:102]: Similarly we are setting below properties for kafka :
```
        kafkaProperties := &instance.Status.Kafka.KafkaProperties
	kafkaProperties.ExternalURI = uri
```

File: pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go (hunk 10 above)
Comment (vaibhavjainwiz) [reply pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go:51]: done

File: pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go (hunk 11 above)
Comment (vaibhavjainwiz) [reply pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go:63]: done

File: pkg/controller/kogitoinfra/kogitoinfra_service.go [hunk 21]
```diff
@@ -0,0 +1,92 @@
...
+func isInfinispanResource(instance *v1alpha1.KogitoInfra) bool {
+	return instance.Spec.Resource.Kind == infinispanv1.SchemeGroupVersion.String()
+}
+
+func isKafkaResource(instance *v1alpha1.KogitoInfra) bool {
+	return instance.Spec.Resource.Kind == kafkabetav1.SchemeGroupVersion.String()
+}
+
+func isKeycloakResource(instance *v1alpha1.KogitoInfra) bool {
+	return instance.Spec.Resource.Kind == keycloakv1alpha1.SchemeGroupVersion.String()
```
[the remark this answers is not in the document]
vaibhavjainwiz: done

pkg/controller/kogitoinfra/infinispan/infinispan_properties.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/infinispan/resource.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/keycloak/keyclock_properties.go
vaibhavjainwiz: For keycloak we don't set any properties

pkg/controller/kogitoinfra/keycloak/resource.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
vaibhavjainwiz: done

pkg/infrastructure/infinispan.go
vaibhavjainwiz: Thanks for pointing it out. There are lots of pck structure changes needs to be done so we have created seperate Jira for it(KOGITO-3322). We could incorporate your suggested changes with that Jira

pkg/controller/kogitoinfra/keycloak/resource.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
vaibhavjainwiz: I had removed this comment

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
vaibhavjainwiz: I created these temp for code readablity purpose only

pkg/controller/kogitoinfra/infinispan/infinispan_properties.go
vaibhavjainwiz: done

ricardozanini:
@vaibhavjainwiz +1 for the **Approach 2**, many thanks for handling this feature! <3

pkg/apis/app/v1alpha1/kogitoinfra_types.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
vaibhavjainwiz: I had removed these temp

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
vaibhavjainwiz: I had removed these temp

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
```diff
@@ -0,0 +1,93 @@
...
+
+	var keycloakInstance *keycloakv1alpha1.Keycloak
+
+	// Step 1: check whether user has provided custom keycloak instance reference
+	isCustomReferenceProvided := len(instance.Spec.Resource.Name) > 0
+	if isCustomReferenceProvided {
+		log.Debugf("Custom kafka instance reference is provided")
+		resourceName := instance.Spec.Resource.Name
+		resourceNameSpace := instance.Spec.Resource.Namespace
+
```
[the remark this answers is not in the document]
vaibhavjainwiz: I had removed these temp

pkg/controller/kogitoinfra/keycloak/keyclock_reconciler.go
```diff
@@ -0,0 +1,93 @@
...
+			return false, fmt.Errorf("custom Keycloak instance(%s) not found in namespace %s", resourceName, resourceNameSpace)
+		}
+	} else {
+		log.Debugf("Custom kafka instance reference is not provided")
+		// if resource name is not provided then Infinispan instance should be created with default name = kogito-infinispan
+		resourceName := InstanceName
+
+		// if resource name is not provided then Infinispan instance should be created with default name = kogito-infinispan
+		resourceNameSpace := instance.Namespace
+
```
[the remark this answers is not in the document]
vaibhavjainwiz: I had removed these temp

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
vaibhavjainwiz: I had removed this update

pkg/controller/kogitoinfra/kafka/kafka_reconciler.go
vaibhavjainwiz: I had removed these temp

pkg/controller/kogitoinfra/kafka/kafka_reconciler.go
vaibhavjainwiz: I had removed these temp

pkg/controller/kogitoinfra/kafka/kafka_reconciler.go
vaibhavjainwiz: I had removed this function

pkg/controller/kogitoinfra/kogitoinfra_controller.go
```diff
@@ -116,121 +92,28 @@ type ReconcileKogitoInfra struct {
...
-	} else if !exists {
-		return reconcile.Result{}, nil
-	}
-
-	// watcher will be nil on test env
-	if controllerWatcher != nil && !controllerWatcher.AreAllObjectsWatched() {
-		if (instance.Spec.InstallKafka && !controllerWatcher.IsGroupWatched(kafkav1beta1.SchemeGroupVersion.Group)) ||
-			(instance.Spec.InstallInfinispan && !controllerWatcher.IsGroupWatched(infinispanv1.SchemeGroupVersion.Group)) ||
-			(instance.Spec.InstallKeycloak && !controllerWatcher.IsGroupWatched(keycloakv1alpha1.SchemeGroupVersion.Group)) {
-			// try to add them
```
ricardozanini, line 142: Not sure if we need to remove this check. This is important since if we add a Watch to a cluster that does not have the Infinispan, Kafka, Keycloak CRDs installed, a lot of errors will be seeing in the logs, might not even work. Maybe adding in the reconcilers you created?

pkg/controller/kogitoinfra/kogitoinfra_controller.go
```diff
@@ -116,121 +92,28 @@ type ReconcileKogitoInfra struct {
...
-			if err := controllerWatcher.Watch(watchedObjects...); err != nil {
-				return reconcile.Result{}, err
-			}
-			if !controllerWatcher.AreAllObjectsWatched() {
-				log.Warn("Dependencies not found for KogitoInfra, please install them.")
-				return reconcile.Result{RequeueAfter: time.Minute * 5}, nil
```
ricardozanini, line 148: This check can be added in your reconcilers. :)

pkg/controller/kogitoinfra/kogitoinfra_resource.go
```diff
@@ -0,0 +1,26 @@
...
+package kogitoinfra
+
+import (
+	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1alpha1"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// InfraResource Interface to represent type of supported kogito infra resources like Infinispan, kafka & keycloak
+type InfraResource interface {
```
ricardozanini, line 24: You don't need to provide an interface if you can hide the functions in the packages and you will only have one implementation of it.

pkg/controller/kogitojobsservice/kogitojobsservice_controller.go
```diff
@@ -141,13 +140,8 @@ const (
 	maxIntervalLimitRetryEnvKey       = "MAX_INTERVAL_LIMIT_RETRY"
 	backOffRetryDefaultValue          = 1000
 	maxIntervalLimitRetryDefaultValue = 60000
-	kafkaTopicNameJobsEvents          = "kogito-job-service-job-status-events"
 )
 
-var kafkaTopics = []services.KafkaTopicDefinition{
-	{TopicName: kafkaTopicNameJobsEvents, MessagingType: services.KafkaTopicOutgoing},
-}
-
```
ricardozanini, line 150: Same thing here related to my comment on Data Index

pkg/controller/kogitotrusty/kogitotrusty_controller.go
```diff
@@ -145,18 +144,3 @@ func (r *ReconcileKogitoTrusty) Reconcile(request reconcile.Request) (reconcile.
...
-	kafkaTopicNameExplainabilityResultEvents  string = "trusty-explainability-result"
-	kafkaTopicNameExplainabilityRequestEvents string = "trusty-explainability-request"
-)
-
-var kafkaTopics = []services.KafkaTopicDefinition{
-	{TopicName: kafkaTopicNameTraceEvents, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameModelEvents, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameExplainabilityResultEvents, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameExplainabilityRequestEvents, MessagingType: services.KafkaTopicOutgoing},
-}
```
ricardozanini, line 162: Same thing about creating `KafkaTopics` if the `KogitoInfra` kind is from Strimzi operator.

ricardozanini (changes requested):
Looks great! Pretty nice to see all this code going away. Please just take a look in the comments are left mentioning that we need `KafkaTopics` to be created for supporting services if `KogitoInfra` relates to a Kafka kind. This way we guarantee a smooth integration with third parties, but not providing too much details. This resources should not be owned or reconciled by us, but on Strimzi. So users could update any information they would need theirs.

Can you please wait for #567 to be merged so we rebase? I believe they have some entwined code. Thanks again for this!

go.mod
```diff
@@ -2,6 +2,7 @@ module github.com/kiegroup/kogito-cloud-operator
 
 require (
 	github.com/RHsyseng/operator-utils v0.0.0-20200304191317-2425bf382482
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
```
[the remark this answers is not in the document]
vaibhavjainwiz: yes, we can remove it. I had just add new testcase and IDE automatically fetched it to support `assert` statement.

pkg/controller/kogitodataindex/kogitodataindex_controller.go
```diff
@@ -149,23 +148,8 @@ const (
...
 )
 
-var kafkaTopics = []services.KafkaTopicDefinition{
-	{TopicName: kafkaTopicNameProcessInstances, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameUserTaskInstances, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameProcessDomain, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameUserTaskDomain, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameJobsEvents, MessagingType: services.KafkaTopicIncoming},
-}
-
```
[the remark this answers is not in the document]
vaibhavjainwiz: only `kafka.bootstrap.servers` property need to be define. User don't need to explicitly define these topics even for data-index

pkg/controller/kogitoexplainability/kogitoexplainability_controller.go
```diff
@@ -140,14 +139,3 @@ func (r *ReconcileKogitoExplainability) Reconcile(request reconcile.Request) (re
...
-const (
-	// Collection of kafka topics that should be handled by the Explainability service
-	kafkaTopicNameExplainabilityRequest string = "trusty-explainability-request"
-	kafkaTopicNameExplainabilityResult  string = "trusty-explainability-result"
-)
-
-var kafkaTopics = []services.KafkaTopicDefinition{
-	{TopicName: kafkaTopicNameExplainabilityRequest, MessagingType: services.KafkaTopicIncoming},
-	{TopicName: kafkaTopicNameExplainabilityResult, MessagingType: services.KafkaTopicOutgoing},
-}
```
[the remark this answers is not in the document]
vaibhavjainwiz: only kafka.bootstrap.servers property need to be define. User don't need to explicitly define these topics even for data-index

pkg/controller/kogitoinfra/infinispan/manage_status.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func updateEnvVarsInStatus(cli *client.Client, infinispanInstance *infinispan.Infinispan, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) error {
+	log.Debugf("going to Update Infinispan env properties in kogito infra instance status")
+	envVars, err := getInfinispanSecretEnvVars(cli, infinispanInstance, instance, scheme)
+	if err != nil {
+		return err
+	}
+	instance.Status.EnvVars = envVars
+	log.Debugf("Following env properties are set infra status : %s", envVars)
```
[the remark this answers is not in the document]
vaibhavjainwiz: not the value. It shows only Secret Name and key

pkg/controller/kogitoinfra/kogitoinfra_controller.go
vaibhavjainwiz: I had already included these check in their respective reconciler. 
https://github.com/vaibhavjainwiz/kogito-cloud-operator/blob/25ffe8183c9dfa325fa814790b8f41ab1c442834/pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go#L68
https://github.com/vaibhavjainwiz/kogito-cloud-operator/blob/25ffe8183c9dfa325fa814790b8f41ab1c442834/pkg/controller/kogitoinfra/kafka/kafka_reconciler.go#L64

pkg/controller/kogitoinfra/kogitoinfra_controller.go
vaibhavjainwiz: already did

pkg/controller/kogitoinfra/kogitoinfra_resource.go
vaibhavjainwiz: `kogitoinfra_reconciler ` uses this interface reference to call reconcile method on different resources. 
https://github.com/vaibhavjainwiz/kogito-cloud-operator/blob/25ffe8183c9dfa325fa814790b8f41ab1c442834/pkg/controller/kogitoinfra/kogitoinfra_controller.go#L108

pkg/controller/kogitojobsservice/kogitojobsservice_controller.go
vaibhavjainwiz: I could double verify it but user only need to define `kafka.bootstrap.servers` property and these topics will be created automatically

pkg/controller/kogitodataindex/kogitodataindex_controller.go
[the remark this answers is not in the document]
ricardozanini: Hi! Actually the Topics are required if the cluster is not set to auto create topics. 

Since we do know which topics we require for the supporting services (like Data Index and Trusty), we should create them if KogitoInfra is associated with a Kafka CR.

pkg/controller/kogitoinfra/kogitoinfra_resource.go
ricardozanini: Hi! Yes, but what I'm saying is that we do not need an interface of it's not implemented by more than once, which is the case. Plus this feature is in a package which you can hide the implementation with unexported functions.

ricardozanini:
Hi @vaibhavjainwiz  could you also please update the https://github.com/kiegroup/kogito-cloud-operator/blob/master/RELEASE_NOTES.md?

pkg/controller/kogitodataindex/kogitodataindex_controller.go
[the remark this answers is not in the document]
vaibhavjainwiz: done

pkg/controller/kogitoexplainability/kogitoexplainability_controller.go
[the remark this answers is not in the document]
vaibhavjainwiz: done

pkg/controller/kogitojobsservice/kogitojobsservice_controller.go
vaibhavjainwiz: done

pkg/controller/kogitotrusty/kogitotrusty_controller.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/infinispan/infinispan_properties.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	appProps[propertiesInfinispanSpring[appPropInfinispanUseAuth]] = "true"
+	appProps[propertiesInfinispanQuarkus[appPropInfinispanUseAuth]] = "true"
+	if len(infinispanURI) > 0 {
+		appProps[propertiesInfinispanSpring[appPropInfinispanServerList]] = infinispanURI
+		appProps[propertiesInfinispanQuarkus[appPropInfinispanServerList]] = infinispanURI
+	}
+
+	// TODO : Check requirement -> How operator could configure these properties as user don't provide any infinispan spec input
```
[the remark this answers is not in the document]
vaibhavjainwiz: done

vaibhavjainwiz:
> Hi @vaibhavjainwiz could you also please update the https://github.com/kiegroup/kogito-cloud-operator/blob/master/RELEASE_NOTES.md?

@ricardozanini I had updated RELEASE_NOTES

pkg/apis/app/v1alpha1/kogitoinfra_types.go
```diff
@@ -19,45 +19,49 @@ import (
...
-	Infinispan InfinispanInstallStatus         `json:"infinispan,omitempty"`
-	Kafka      InfraComponentInstallStatusType `json:"kafka,omitempty"`
-	Keycloak   InfraComponentInstallStatusType `json:"keycloak,omitempty"`
+	Condition KogitoInfraCondition `json:"condition,omitempty"`
+
+	// +optional
+	// +mapType=atomic
+	// Application properties to be added to the runtime container.
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	AppProps map[string]string `json:"appProps,omitempty"`
```
ricardozanini, line 58:
```suggestion
	// Application properties extracted from the linked resource that will be added to the deployed Kogito service.
	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
	AppProps map[string]string `json:"appProps,omitempty"`
```

pkg/apis/app/v1alpha1/kogitoinfra_types.go
```diff
@@ -19,45 +19,49 @@ import (
...
+
+	// +optional
+	// +listType=atomic
+	// Environment variables to be added to the runtime container.
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	EnvVars []v1.EnvVar `json:"envVars,omitempty"`
```
ricardozanini, line 64:
```suggestion
	// Environment variables extracted from the linked resource that will be added to the deployed Kogito service.
	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
	Env []v1.EnvVar `json:"env,omitempty"`
```

My suggestion to change to `Env` is to have coherence with #586

pkg/apis/app/v1alpha1/kogitoinfra_types.go
```diff
@@ -116,11 +94,9 @@ const (
 // infrastructure dependency (such as Infinispan) to a third party operator.
 // It holds the deployment status of each infrastructure dependency and custom
 // resources needed to run Kogito Runtime and Kogito Data Index services.
+// KafkaTopics to be created for supporting services if KogitoInfra relates to a Kafka kind
```
ricardozanini, line 97: I believe we need to rethink this description, what about:

> KogitoInfra is the resource to bind a Custom Resource (CR) not managed by Kogito Operator to a given deployed Kogito service. It holds the reference of a CR managed by another operator such as Strimzi. For example: one can create a Kafka CR via Strimzi and link this resource using KogitoInfra to a given Kogito service (custom or supporting, such as Data Index). Please refer to the Kogito Operator documentation (https://docs.jboss.org/kogito/release/latest/html_single/) for more information.

ricardozanini (approved):
Looks good! I just left some minor comments regarding description/interface naming.

sutaakar:
old Jenkinsfile was applied, I guess the PR needs to be rebased to apply correct Jenkinsfile

pkg/apis/app/v1alpha1/kogitoinfra_types.go
vaibhavjainwiz: done

pkg/apis/app/v1alpha1/kogitoinfra_types.go
vaibhavjainwiz: done

pkg/apis/app/v1alpha1/kogitoinfra_types.go
vaibhavjainwiz: yes..its more clear

sutaakar:
@vaibhavjainwiz In case resource name is defined in KogitoInfra then namespace is required too. Would it have sense to use current namespace by default and don't require user to always specify namespace?

sutaakar:
Also by default the user doesn't have rights to access CRs from different namespaces. So in case a different namespace is specified then user have to perform additional steps.

sutaakar:
Should we support using CRs from different namespaces than the ones where Kogito operator is running?

vaibhavjainwiz:
> Would it have sense to us

vaibhavjainwiz:
> @vaibhavjainwiz In case resource name is defined in KogitoInfra then namespace is required too. Would it have sense to use current namespace by default and don't require user to always specify namespace?

yes sure, we could do so.

ricardozanini:
> Should we support using CRs from different namespaces than the ones where Kogito operator is running? 

Great catch! Yes, we should be able to access CRs from different namespaces. In this case our operator user will need access there. We will change the operator to multi namespace, thus the user will have cluster role. 

But for now we should guide the user to add these permissions manually vai documentation.

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	// APIVersion refers to infinispan APIVersion
+	APIVersion = "app.infinispan.org/v1"
```
sutaakar, line 30: Correct APIVersion is `infinispan.org/v1`

pkg/apis/app/v1alpha1/kogitoinfra_types.go
```diff
@@ -19,45 +19,49 @@ import (
...
 
-// InfraComponentInstallStatusType is the base structure to define the status for an actor in the infrastructure.
-type InfraComponentInstallStatusType struct {
-	Service   string             `json:"service,omitempty"`
-	Name      string             `json:"name,omitempty"`
-	Condition []InstallCondition `json:"condition,omitempty"`
+// Resource provide reference infra resource
+type Resource struct {
+
+	// APIVersion describes the API Version of referred Kubernetes resource for example, app.infinispan.org/v1
```
sutaakar, line 25: same here

pkg/controller/kogitoinfra/infinispan/infinispan_reconciler.go
vaibhavjainwiz: fixed it

pkg/apis/app/v1alpha1/kogitoinfra_types.go
vaibhavjainwiz: done

vaibhavjainwiz:
> Also by default the user doesn't have rights to access CRs from different namespaces. So in case a different namespace is specified then user have to perform additional steps.

ricardozanini (approved):
Code looks ok, but the BDD tests are failing.

ricardozanini:
@vaibhavjainwiz one last question: I did not see the creation of `KafkaTopic` CR in the Data Index, Jobs Service, Trusty services. I might be missing something in the PR. :(

vaibhavjainwiz:
> @vaibhavjainwiz one last question: I did not see the creation of `KafkaTopic` CR in the Data Index, Jobs Service, Trusty services. I might be missing something in the PR. :(

Sorry @ricardozanini Its my mistake. I had misunderstood your kafkaTopic comment. Previously I had only append kafka topics entries in application.properties but not create them if they don't exists. I had fixed this gap now.
Code to implement that logic is present at below location:

https://github.com/vaibhavjainwiz/kogito-cloud-operator/blob/a4dbdb3520d7667ca42f2eee35008d136c848f48/pkg/infrastructure/services/kafka.go#L32

ricardozanini:
> **Note:**
> 
> Kogito operator will create kafka topics only if infra reference is provided through kogitoInfra CR. 
> 
> If user has created its custom kafka instance and provide its reference directly through application.properties file then kogito operator will not check/create kafka topics
> 
> 

Perfect! Makes totally sense to me and it's perfect for a first interaction and 1.0. 

We can improve this later with Crossplane. 

Many thanks!

ricardozanini (approved):
Looks good! As soon as @sutaakar finish his checks and BDD update, I believe we will be ready to go. We really need this feature now to unlock various JIRAs for 1.0v 

Thanks again for the hard work.

cmd/kogito/command/flag/runtime_flag.go
```diff
@@ -38,8 +36,6 @@ type RuntimeFlags struct {
 // AddRuntimeFlags adds the RuntimeFlags to the given command
 func AddRuntimeFlags(command *cobra.Command, flags *RuntimeFlags) {
 	AddInstallFlags(command, &flags.InstallFlags)
-	AddInfinispanFlags(command, &flags.InfinispanFlags)
-	AddKafkaFlags(command, &flags.KafkaFlags)
 	AddMonitoringFlags(command, &flags.MonitoringFlags)
 	command.Flags().BoolVar(&flags.EnableIstio, "enable-istio", false, "Enable Istio integration by annotating the Kogito service pods with the right value for Istio controller to inject sidecars on it. Defaults to false")
 	command.Flags().BoolVar(&flags.EnablePersistence, "enable-persistence", false, "If set to true, deployed Kogito service will support integration with Infinispan server for persistence. Default to false")
```
sutaakar, line 41: @vaibhavjainwiz I suppose that `enable-persistence` and `enable-events` can be removed from KogitoRuntime CLI flags, right?

ricardozanini: I think we can remove since he removed the Infinispan and Kafka creation. We won't remove that feature on this PR, but since it's done, yeah, we can get rid of these flags.

sutaakar:
This PR is causing an issue with Jobs service, reported as https://issues.redhat.com/browse/KOGITO-3446

vaibhavjainwiz:
> This PR is causing an issue with Jobs service, reported as https://issues.redhat.com/browse/KOGITO-3446. Jobs service scripts need to be adjusted.

@sutaakar 
As I check, `events-support.quarkus.kafka.bootstrap-servers` property is define for job-service but I didn't found any reference for this properties in Kogito Docs(https://docs.jboss.org/kogito/release/latest/html_single/#proc-jobs-service-messaging_kogito-configuring). Does this property is still required? or Should I remove it?

https://github.com/kiegroup/kogito-images/blob/b4c202559a6c0ce862f72b18e41c0dd4bac6616e/modules/kogito-jobs-service/added/launch/kogito-jobs-service.sh#L36

test/steps/data.go
```diff
@@ -48,7 +48,7 @@ func (data *Data) RegisterAllSteps(ctx *godog.ScenarioContext) {
 	registerKogitoDataIndexServiceSteps(ctx, data)
 	registerKogitoExplainabilityServiceSteps(ctx, data)
 	registerKogitoTrustyServiceSteps(ctx, data)
-	registerKogitoInfraSteps(ctx, data)
+	//registerKogitoInfraSteps(ctx, data)
```
spolti, line 51: remove?

test/steps/kogitodataindex.go
```diff
@@ -59,9 +58,9 @@ func (data *Data) installKogitoDataIndexServiceWithReplicasWithConfiguration(rep
 		if err := framework.CreateSecret(data.Namespace, kogitoExternalInfinispanSecret, map[string]string{usernameSecretKey: dataIndex.Infinispan.Username, passwordSecretKey: dataIndex.Infinispan.Password}); err != nil {
 			return err
 		}
-		dataIndex.KogitoService.(*v1alpha1.KogitoDataIndex).Spec.InfinispanProperties.Credentials.SecretName = kogitoExternalInfinispanSecret
-		dataIndex.KogitoService.(*v1alpha1.KogitoDataIndex).Spec.InfinispanProperties.Credentials.UsernameKey = usernameSecretKey
-		dataIndex.KogitoService.(*v1alpha1.KogitoDataIndex).Spec.InfinispanProperties.Credentials.PasswordKey = passwordSecretKey
+		//dataIndex.KogitoService.(*v1alpha1.KogitoDataIndex).Spec.InfinispanProperties.Credentials.SecretName = kogitoExternalInfinispanSecret
```
spolti, line 61: remove?

test/steps/kogitoinfra.go
```diff
@@ -14,7 +14,7 @@
 
 package steps
 
-import (
+/*import (
```
spolti, line 17: remove it?

test/framework/kogitoinfra.go
```diff
@@ -14,6 +14,7 @@
 
 package framework
 
+/*
```
spolti, line 17: remove?

pkg/infrastructure/services/deployer_test.go
```diff
@@ -53,7 +53,7 @@ func Test_serviceDeployer_Deploy(t *testing.T) {
 	deployer := NewSingletonServiceDeployer(definition, serviceList, cli, meta.GetRegisteredSchema())
 	reconcileAfter, err := deployer.Deploy()
 	assert.NoError(t, err)
-	assert.True(t, reconcileAfter > 0) // we just deployed Infinispan
+	assert.Equal(t, time.Duration(0), reconcileAfter) // we just deployed Infinispan
```
spolti, line 56: maybe rephrase this comment? didn't see infinispan getting deployed.

spolti (changes requested):
very nice, just a minor comments.

test/steps/data.go
vaibhavjainwiz: I keep them commented because BDD tests needs to be updated for changes in KOGITO-3039

pkg/infrastructure/services/deployer_test.go
vaibhavjainwiz: done

test/steps/kogitoinfra.go
vaibhavjainwiz: done

test/steps/kogitodataindex.go
vaibhavjainwiz: done

test/steps/data.go
vaibhavjainwiz: done

test/framework/kogitoinfra.go
vaibhavjainwiz: done

pkg/controller/kogitoinfra/kogitoinfra_service.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoinfra/infinispan"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoinfra/kafka"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/controller/kogitoinfra/keycloak"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// FetchKogitoInfraInstance load infra instance
+func FetchKogitoInfraInstance(client *client.Client, name string, namespace string) (*v1alpha1.KogitoInfra, error) {
```
radtriste, line 29: small suggestion: can't we reuse this one or https://github.com/kiegroup/kogito-cloud-operator/pull/550/files#diff-b60bb571f748158bb6df328282f4380eR172 and put in `infrastructure` package ? seems doing twice the same or ?

radtriste (approved):
nice job ! 
one minor suggestion :)

pkg/controller/kogitoinfra/kogitoinfra_service.go
vaibhavjainwiz: done

sutaakar (approved):
I didn't find any other issues apart from the reported Jobs service one, other components seem to work well.
BDD tests are adjusted, I will create a PR with adjustments once this PR is merged, to keep PR size reasonable.

spolti (approved):

PR #567, opened 2020-09-15
[KOGITO-3270] - Accepting external ConfigMaps to be mounted in a give…

ricardozanini:
…n KogitoService

See: https://issues.redhat.com/browse/KOGITO-3270

In this PR we introduce a new attribute in the Kogito services interface, which can be informed to tell the operator to not create a configMap, but to use the one given in advance. This way users can configure the application upfront using an external process (like a CI) before deploying a given Kogito service in the namespace.

For example, one can create a configMap with:

```shell
$ oc create configmap mycustom-configmap --from-file=application.properties 
```

And then deploy the KogitoRuntime with:

```yaml
kind: KogitoRuntime
metadata:
  name: process-quarkus-example
spec:
  propertiesConfigMap: mycustom-configmap
  replicas: 1
  image:
    domain: quay.io
    namespace: kiegroup
    name: process-quarkus-example
```

The same applies for any other supporting service like Management Console or Data Index.

Every property from the given CM, manually inserted properties, or properties manipulated by the operator will be in this CM and managed by the operator.

Note: I'm adding a RELEASE_NOTES.md file to be used later by our CI to create the Release Notes in the GH page. Please update this file from now on when sending a PR that have a feature/bug fix that needs to be included in the release notes.

### DIFF
--- a/cmd/kogito/command/service/config.go
+++ b/cmd/kogito/command/service/config.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	configMapPrefix     = "custom-properties"
+	configMapSuffix     = "custom-properties"
 	createdByAnnonKey   = "createdBy"
 	createdByAnnonValue = "Kogito CLI"
 )
@@ -38,7 +38,7 @@ func createConfigMapFromFile(cli *client.Client, flags *flag.RuntimeFlags) (cmNa
 		return "", nil
 	}
 	cm := &v1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", flags.Name, configMapPrefix), Namespace: flags.Project},
+		ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", flags.Name, configMapSuffix), Namespace: flags.Project},
 	}
 	fileContent, err := ioutil.ReadFile(flags.ConfigFile)
 	if err != nil {
--- a/cmd/kogito/command/service/config.go
+++ b/cmd/kogito/command/service/config.go
@@ -1,0 +1,69 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package service
+
+import (
+	"fmt"
+	"github.com/kiegroup/kogito-cloud-operator/cmd/kogito/command/flag"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure/services"
+	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	configMapPrefix     = "-custom-properties"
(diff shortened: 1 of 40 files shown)


spolti (approved):

vaibhavjainwiz:
@ricardozanini 

I have some doubts, could you please help me to clarify them

- User could provide more then one config map.

- User could creates config map from file as well as from literal.

- With new approch in KOGITO-3039, kogito operator will create seperate config map for each kogitoInfra object. I think its better to create seperate config map for kogito-operator managed properties. With this we could easy maintain config map ownership and we would not made any changes with user managed config map.

- Could we let user to provide its volume/volume mount if he is creating config map from file or we have to take extra input from user whether he create config map from file or literal? I think it will be better from user usablity point of view if we could provide two seperate property like `configMapFromLiteral` and `configMapFromFile`. With this user don't need to provide volume/volume mount details.

ricardozanini:
> @ricardozanini
> 
> I have some doubts, could you please help me to clarify them
> 
>     * User could provide more then one config map.

No, they can't. This approach is to just configure their application based on Quarkus or Spring runtimes. The configMap is mounted in the `config` directory, which is read by the application in runtime.

>     * User could creates config map from file as well as from literal.

This configMap supposed to be mounted as an `application.properties` file. Whatever the way they create the CM, if it has the `application.properties` key will work.

>     * With new approch in KOGITO-3039, kogito operator will create seperate config map for each kogitoInfra object. I think its better to create seperate config map for kogito-operator managed properties. With this we could easy maintain config map ownership and we would not made any changes with user managed config map.

We shouldn't create separated CMs since all the properties should be in the `application.properties` file mounted in the pod. We can however, create one CM for each infra for our control, in our format. Then we can merge all the properties in the one which will be mounted in the `config` directory.

>     * Could we let user to provide its volume/volume mount if he is creating config map from file or we have to take extra input from user whether he create config map from file or literal? I think it will be better from user usablity point of view if we could provide two seperate property like `configMapFromLiteral` and `configMapFromFile`. With this user don't need to provide volume/volume mount details.

I don't think we need to open that much, users can create/map their on CMs/keys using the `envs` attribute. About volume/volume mounts, this can be another JIRA/conversation if they need additional files being mounted. We can let this discussion to the future, since all files needed by the application are handled: protobuf and properties. This can change of course.

The main motivation here is just for the user to inform a custom application.properties to the operator before the deployment. This will cover CI/CD process where you have a Consul/Netflix/Spring properties vault which you mount then during this process and set that CM to the Kogito CR.

vaibhavjainwiz (approved):
LGTM

sutaakar (approved):
Looks good to me, good job.

sutaakar:
@ricardozanini Looks like the property is currently not provided in CLI. DO you plan to add CLI support?

ricardozanini:
@sutaakar yes indeed I have to add a flag in the deployment command. Will do it today.

ricardozanini:
@sutaakar please take a look again. I've added a new feature to the CLI where an user could give a local file to be set as a property file to the application. Do you mind verifying it? :-)

Many thanks!

cmd/kogito/command/service/config.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure/services"
+	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	configMapPrefix     = "-custom-properties"
```
sutaakar, line 29:
```suggestion
	configMapPrefix     = "custom-properties"
```
Current implementation creates double dash: `example-quarkus--custom-properties`

sutaakar (approved):
One minor suggestion above.

cmd/kogito/command/service/config.go
sutaakar: Also could be called `configMapSuffix` to be precise :)

ricardozanini: LOL! I'm pretty dumb. Gonna fix ASAP. Many thanks @sutaakar 

Have you had the opportunity to verify it?

sutaakar: yes, works as expected

spolti (approved):

ricardozanini:
@spolti AWS is down

PR #464, opened 2020-07-21
KOGITO-2812 - add explainability service

r00ta:
Jira: https://issues.redhat.com/browse/KOGITO-2812

The aim of this pr is to add the support of the explainability service in the operator. The architecture looks like the following: 

Screenshot from 2020-08-25 10-32-12

In particular we have 
1) the trusty service that will generate some explainability requests
2) the explainability service takes the requests, perturbs the inputs of the decision, makes an http call to the kogito runtime service to retrieve the results and then use the explainability library to calculate the features importance.
3) the explainability service generates explainability result events that are consumed by the trusty service. 
4) the explainability results are exposed via an endpoint.

In order to reach the kogito runtime service, the route of that service is injected in the explainability service instances. 

Many thanks for submiting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.MD#sending-a-pull-request)
- [x] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Your feature/bug fix has a unit test that verifies it
- [x] You've tested the new feature/bug fix in an actual OpenShift cluster

PR #568, opened 2020-09-15
[0.15.x] KOGITO-2812 - cherry pick explainability pr

r00ta:
Cherry pick of https://github.com/kiegroup/kogito-cloud-operator/pull/464


r00ta:
Since I had to rebase my PR on master, it contains also all the previous commits that were not included when the branch 0.15.x was created. 
Let me know if it's ok for you or if I should do some magic to remove such commits..

test/features/install_explainability.feature
```diff
@@ -0,0 +1,56 @@
+@explainability
+@kafka
+Feature: Kogito Explainability service
+
+  Background:
+    Given Namespace is created
+
+  @smoke
+  # Disabled as long as https://issues.redhat.com/browse/KOGITO-3176 is not solved
```
Sgitario, line 9: As far I can see, the https://issues.redhat.com/browse/KOGITO-3176 ticket has two PRs and both merged, maybe this is already fixed?

Sgitario (changes requested):
This PR brings other changes related to other tickets. If this needs to go to 0.15.x, we should either isolate these changes without more unrelated things or cherry-pick the unrelated changes individually to 0.15.x. 
Wdyt? cc @sutaakar

test/features/install_explainability.feature
r00ta: Hi @Sgitario , those PR are just disabling this test. They are not the fix. I'll remove them from the ticket

r00ta:
Hi @Sgitario , I'll remove the additional commits that the rebase included

r00ta:
Hi @Sgitario , I've reverted all the additional commits. I did not squash the cherry pick so that you can see the revert commits. 
Both PRs are 1469 lines long with 8 deletes: they are the same. 
Let me know if this is ok now

test/steps/data.go
```diff
@@ -46,6 +46,7 @@ func (data *Data) RegisterAllSteps(ctx *godog.ScenarioContext) {
 	registerKogitoBuildSteps(ctx, data)
 	registerKogitoRuntimeSteps(ctx, data)
 	registerKogitoDataIndexServiceSteps(ctx, data)
+	registerKogitoExplainabilityServiceSteps(ctx, data)
 	registerKogitoTrustyServiceSteps(ctx, data)
 	registerKogitoInfraSteps(ctx, data)
 	registerKogitoJobsServiceSteps(ctx, data)
```
Sgitario, line 52: Can you add the new KogitoExplainabilityList custom resource in [this line](https://github.com/kiegroup/kogito-cloud-operator/blob/3fa1d80556ac5d89f26b4be17145196576863f19/test/steps/data.go#L112) as well? This is really useful to debug purposes after running the scenarios. 

I added the same comment in the PR master.

pkg/controller/kogitoexplainability/kogitoexplainability_controller.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2019 Red Hat, Inc. and/or its affiliates
```
Sgitario, line 1: It should be 2020.

pkg/controller/kogitoexplainability/kogitoexplainability_controller_test.go
```diff
@@ -0,0 +1,166 @@
+// Copyright 2019 Red Hat, Inc. and/or its affiliates
```
Sgitario, line 1: It should be 2020.

pkg/infrastructure/explainability.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2019 Red Hat, Inc. and/or its affiliates
```
Sgitario, line 1: It should be 2020.

test/framework/kogitoexplainability.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Red Hat, Inc. and/or its affiliates
```
Sgitario, line 1: It should be 2020.

test/steps/kogitoexplainability.go
```diff
@@ -0,0 +1,128 @@
+// Copyright 2019 Red Hat, Inc. and/or its affiliates
```
Sgitario, line 1: It should be 2020.

test/features/install_explainability.feature
```diff
@@ -0,0 +1,56 @@
...
+  @disabled
+  Scenario: Install Kogito Explainability
+    Given Kogito Operator is deployed with Kafka operator
+    And Install Kogito Explainability with 1 replicas
+    Then Kogito Explainability has 1 pods running within 10 minutes
+    And HTTP GET request on service "explainability" is successful within 2 minutes with path "health/live"
```
Sgitario, line 15: I think that if this endpoint returns something is down, the pod never starts and hence the previous step should be sufficient.

test/features/install_explainability.feature
```diff
@@ -0,0 +1,56 @@
...
+
+#####
+
+  @events
+  @kafka
+  Scenario: Explainability retrieves explainability requests events using external Kafka
```
Sgitario, line 21: In this scenario, we're testing this integration using external kafka and infinispan instances. Usually, we tag these scenarios using the annotation "@externalcomponent" (see example [here](https://github.com/kiegroup/kogito-cloud-operator/blob/3fa1d80556ac5d89f26b4be17145196576863f19/test/features/operator/install_jobs_service.feature#L78)).

However, I'm also missing a scenario using the kafka and Infinispan instances that are deployed automatically.

test/steps/kogitoexplainability.go
```diff
@@ -0,0 +1,128 @@
...
+	executionsPath = data.ResolveWithScenarioContext(executionsPath)
+	requestInfo = framework.NewGETHTTPRequestInfo(uri, executionsPath)
+	responseContent = data.ResolveWithScenarioContext(responseContent)
+	return framework.WaitForOnOpenshift(data.Namespace, fmt.Sprintf("GET request on path %s to return response content '%s'", executionsPath, responseContent), timeoutInMin,
+		func() (bool, error) {
+			return framework.DoesHTTPResponseContain(data.Namespace, requestInfo, responseContent)
+		})
+}
+
+func getJSON(r *http.Response, target interface{}) error {
```
Sgitario, line 124: I think this method can be moved to framework/http.go. There are some places that could reuse this logic: keycloak.go (in master branch), and http.go too.

test/steps/kogitoexplainability.go
```diff
@@ -0,0 +1,128 @@
...
+
+type executionsResponse struct {
+	Executions []execution `json:"headers"`
+}
+
+type execution struct {
+	ExecutionID string `json:"executionId"`
+}
+
+func (data *Data) explainabilityResultIsAvailable(timeoutInMin int) error {
```
Sgitario, line 73: If I understood well the logic in this method, we're doing the following:

1. Await for call http://trusty/executions to check whether contains DECISION
2. Call http://trusty/executions and marshall to executionsResponse
3. Get the first execution from executionsResponse
4. Await for call http://trusty/executions/decisions/{first execution ID}/explanations/saliencies to check whether the content is SUCCEEDED

I think we're doing one unnecessary call (step 2) and one unnecessary await (step 4). Also, getting the first execution can be tricky, is there a way we can ensure that the decision is the one we expect? Maybe by name?

test/features/install_explainability.feature
```diff
@@ -0,0 +1,56 @@
...
+        "salary": 2000
+      },
+      "Loan": {
+        "duration": 40,
+        "installment": 1000
+      },
+      "SupremeDirector": "yes"
+      }
+      """
+    Then Explainability result is available in the Trusty service within 3 minutes
```
Sgitario, line 56: Instead of checking if the result is available, does it make sense to assert somehow the content? This way the test would be meaningful just by reading it.

Sgitario (commented):
Thanks for removing the additional comments! I could see better the changes and wrote some comments (if you agree with these comments, I guess these comments also apply to the dependant PR in the master).

ricardozanini:
@r00ta I'm sorry, but it seems that we have lots of comments in this PR yet.. I'd say that we should push to 0.16. 

@sutaakar @spolti @Sgitario need your inputs as well on this.

sutaakar:
fine for me to postpone it, to have proper time to check its functionality

test/features/install_explainability.feature
r00ta: Hi @Sgitario , it depends on what we want to test. IMO the operator is just provisioning the resources, it should not check what is the content of the service. The concept is: the operator is provisioning the resources, then the operator should check that from an infrastructural point of view everything works. 
If the operator is supposed to check end 2 end the features of all the services included, just let me know and I'll change the tests. My suggestion is that every piece should check only what it is supposed to do, The services are already unit tested in the respective repositories, I don't think we should cross check the functionalities every time we have an integration. But this is just my opinion/suggestion, I'm happy to align to the team directives in any case

test/steps/kogitoexplainability.go
r00ta: Hi @Sgitario , yes, the second step might be redundant. It's like asserting that an object is not null and then asserting that the properties have a particular value. 
The second question imo is not a problem, since we are executing just one request and there should be just one execution retrieved by the trusty service. We can improve checking that the size of the array is just one

ricardozanini:
Only failed test is the Keycloak one

PR #574, opened 2020-09-17
KOGITO-3369: Cucumber Tests - Refactor Kogito Explainability Changes

Sgitario:
JIRA: https://issues.redhat.com/browse/KOGITO-3369
Description: Apply the changes asked in the comments for https://github.com/kiegroup/kogito-cloud-operator/pull/568#event-3774194017.

### DIFF
--- a/test/features/operator/install_explainability.feature
+++ b/test/features/operator/install_explainability.feature
@@ -6,36 +6,33 @@ Feature: Kogito Explainability service
     Given Namespace is created
 
   @smoke
-  # Disabled as long as https://issues.redhat.com/browse/KOGITO-3176 is not solved
-  @disabled
-  Scenario: Install Kogito Explainability
+  Scenario Outline: Install Kogito Explainability with communication <communication>
     Given Kogito Operator is deployed with Kafka operator
-    And Install Kogito Explainability with 1 replicas
+    And Kafka topic "trusty-explainability-result" is deployed
+    When Install Kogito Explainability with 1 replicas with configuration:
+      | runtime-env  | EXPLAINABILITY_COMMUNICATION   | <communication>   |
     Then Kogito Explainability has 1 pods running within 10 minutes
-    And HTTP GET request on service "explainability" is successful within 2 minutes with path "health/live"
+
+    Examples:
+      | communication | 
+      | messaging     | 
+      | rest          | 
 
 #####
 
   @events
+  @infinispan
   @kafka
-  Scenario: Explainability retrieves explainability requests events using external Kafka
+  Scenario: Explainability retrieves explainability requests events
     Given Kogito Operator is deployed with Infinispan and Kafka operators
-    And Kafka instance "external-kafka" is deployed
-    And Infinispan instance "external-infinispan" is deployed with configuration:
-      | username | developer |
-      | password | mypass |
-    And Install Kogito Trusty with 1 replicas with configuration:
-      | infinispan | username | developer                 |
-      | infinispan | password | mypass                    |
-      | infinispan | uri      | external-infinispan:11222 |
-      | kafka | externalURI | external-kafka-kafka-bootstrap:9092 |
-    And Install Kogito Explainability with 1 replicas with configuration:
-      | kafka | externalURI | external-kafka-kafka-bootstrap:9092 |
+    And Install Kogito Trusty with 1 replicas
+
+    When Install Kogito Explainability with 1 replicas
+    And Clone Kogito examples into local directory
     And Local example service "dmn-tracing-quarkus" is built by Maven using profile "default" and deployed to runtime registry
     And Deploy quarkus example service "dmn-tracing-quarkus" from runtime registry with configuration:
-      | config      | enableEvents        | enabled                                    |
-      | runtime-env | KOGITO_SERVICE_URL  | http://dmn-tracing-quarkus:8080            |
-      | kafka       | externalURI         | external-kafka-kafka-bootstrap:9092        |
+      | config     | enablePersistence | enabled |
+      | config     | enableEvents      | enabled |
     And Kogito Runtime "dmn-tracing-quarkus" has 1 pods running within 10 minutes
     And HTTP POST request on service "dmn-tracing-quarkus" is successful within 2 minutes with path "LoanEligibility" and body:
       """json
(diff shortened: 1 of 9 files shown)


test/features/operator/install_explainability.feature
```diff
@@ -6,17 +6,55 @@ Feature: Kogito Explainability service
     Given Namespace is created
 
   @smoke
-  # Disabled as long as https://issues.redhat.com/browse/KOGITO-3176 is not solved
-  @disabled
```
Sgitario, line 10: I've enabled this test adding a workaround (manually creating the kafka topic in a step). Let me know your thoughts about it.

test/features/operator/install_explainability.feature
```diff
@@ -6,17 +6,55 @@ Feature: Kogito Explainability service
...
+    When Install Kogito Explainability with 1 replicas
     Then Kogito Explainability has 1 pods running within 10 minutes
-    And HTTP GET request on service "explainability" is successful within 2 minutes with path "health/live"
 
 #####
 
   @events
+  @infinispan
+  @kafka
+  Scenario: Explainability retrieves explainability requests events
```
Sgitario, line 21: I've created this new scenario to test the in-build integration with Kafka, Infinispan and the rest of the components.

test/framework/kogitotrusty.go
```diff
@@ -15,12 +15,41 @@
 package framework
```
Sgitario, line 15: I was not sure where to put this logic... Theoretically, it belongs to Kogito Explainability, but we get the data using the Kogito Trusty... Let me know your thoughts about this and we can discuss from here.
cc @r00ta @sutaakar

Sgitario:
I've ran all the scenario in this test (plus one about keycloak that was affected by one of the changes). All worked fine.

test/features/operator/install_explainability.feature
```diff
@@ -25,17 +63,18 @@ Feature: Kogito Explainability service
...
+      | infinispan | uri         | external-infinispan:11222           |
+      | kafka      | externalURI | external-kafka-kafka-bootstrap:9092 |
+
+    When Install Kogito Explainability with 1 replicas with configuration:
       | kafka | externalURI | external-kafka-kafka-bootstrap:9092 |
+    And Clone Kogito examples into local directory
     And Local example service "dmn-tracing-quarkus" is built by Maven using profile "default" and deployed to runtime registry
     And Deploy quarkus example service "dmn-tracing-quarkus" from runtime registry with configuration:
-      | config      | enableEvents        | enabled                                    |
-      | runtime-env | KOGITO_SERVICE_URL  | http://dmn-tracing-quarkus:8080            |
```
r00ta, line 37: I know that BDD tests on kubernetes have some bugs, for example multiple overlapping resources are created -> the tests fail. Btw the config ` | runtime-env | KOGITO_SERVICE_URL  | http://dmn-tracing-quarkus:8080            |` is needed in case somebody runs the bdd tests on kubernetes and not on openshift. 
Just fyi

test/features/operator/install_explainability.feature
```diff
@@ -6,17 +6,55 @@ Feature: Kogito Explainability service
...
+    # Workaround until https://issues.redhat.com/browse/KOGITO-3176 is solved
+    And Kafka topic "trusty-explainability-result" is deployed
+    When Install Kogito Explainability with 1 replicas
     Then Kogito Explainability has 1 pods running within 10 minutes
-    And HTTP GET request on service "explainability" is successful within 2 minutes with path "health/live"
 
 #####
 
   @events
+  @infinispan
```
r00ta, line 19: `@externalcomponent ` ?

r00ta (approved):
Thank you very much @Sgitario , looks good to me!

test/framework/kogitotrusty.go
r00ta: For me it's correct to keep the DTO in `kogitotrusty.go`, since we are interacting with the api exposed by such service

test/features/operator/install_explainability.feature
r00ta: Hi @Sgitario , my 50 cents is that this should not be intented as workaround. imo this is the proper solution to the problem: the user should provide the kafka instance + create the kafka topic resources, and this is exactly what you have implemented here.  
If the user does not want to create manually the topics, he has to enable the auto-creation of topics in the kafka broker and all the services should be deployed.

test/features/operator/install_explainability.feature
Sgitario: Then this sounds like a bigger bug that we should fix in the test framework. I will run the test on Kubernetes and provide a JIRA if I can reproduce it. 
Thanks!

test/features/operator/install_explainability.feature
Sgitario: We add this annotation only when we create the Kafka or Infinispan instances externally (without the kogito operator in between)

test/features/operator/install_explainability.feature
r00ta: Yeah, unfortunately this is what I faced when multiple services are exposed (for example data-index + kogito runtime or trusty + runtime etc..). 
Another thing is that there is an hardcoded port `80` instead of `8080`, which is the default port used by the services

r00ta: The port bug is here https://github.com/kiegroup/kogito-cloud-operator/blob/67fc59bc10abb885ae97f8a48ad146c538c48215/test/framework/kubernetes.go#L401 
And in the same file the issue is the function `ExposeServiceOnKubernetes` which creates overlapping ingress resources

Sgitario: Ok, thanks! Maybe this logic is out to dated and as we don't run the tests against Kubernetes yet, we didn't spot it yet.

test/features/operator/install_explainability.feature
```diff
@@ -6,17 +6,71 @@ Feature: Kogito Explainability service
...
+
+    Examples:
+      | communication | 
+      | messaging     | 
+
+    # Disabled due to https://issues.redhat.com/browse/KOGITO-3373: Trusty Service errors when running Explainability service with rest communication
+    @disabled
+    Examples:
+      | communication | 
+      | rest          |
```
Sgitario, line 67: Hi @r00ta , I made this test but it's not working due to some errors in the Trusty service (I'm getting 400 http error when getting the decisions data). I've raised KOGITO-3373 to solve this, but let me know I'm doing something wrong.

test/features/operator/install_explainability.feature
ricardozanini: For the tests, for now, Kogito Operator creates the Kafka CR with auto-creation topics. If the BDD uses the same approach is not a problem IMO

ricardozanini (approved):

test/features/operator/install_explainability.feature
Sgitario: Ok, I've removed the workaround comment and leave it like this. My only concern now is if the ticket KOGITO-3176 needs fixing or not :/

test/features/operator/install_explainability.feature
Sgitario: I've removed this scenario following your comment in the ticket. We can continue clarifying in the ticket directly.

spolti (approved):

test/framework/kogitotrusty.go
sutaakar: fine for me too

sutaakar (approved):
@Sgitario Thanks a lot for the improvements, looks much better now.

ricardozanini:
@Sgitario looks like a problem in the optaplanner example?

```{"level":"error","T":"2020-09-18T06:45:04.553Z","logger":"jenkins-cucumber-06ac","msg":"Error in step 'Local example service \"process-optaplanner-quarkus\" is built by Maven using profile \"default\" and deployed to runtime registry': signal: killed"}
```

I believe this error has been addressed already?

Sgitario:
> @Sgitario looks like a problem in the optaplanner example?
> 
> ```json
> {"level":"error","T":"2020-09-18T06:45:04.553Z","logger":"jenkins-cucumber-06ac","msg":"Error in step 'Local example service \"process-optaplanner-quarkus\" is built by Maven using profile \"default\" and deployed to runtime registry': signal: killed"}
> ```
> 
> I believe this error has been addressed already?

It willl be fixed as part of https://github.com/kiegroup/kogito-cloud-operator/pull/573

test/features/operator/install_explainability.feature
r00ta: Hi @Sgitario @ricardozanini , I don't think we are going to implement on the trusty service a logic to manually create the topic by the consumer. If I'm not wrong, such feature has been removed by the "official" smallrye library we are using and I don't think we should re-introduce it on our side. 
We have two options: 
1) remove the test on the standalone trusty service
2) create the topics like you proposed (I would go for this)

Sgitario: To provide further context, the ticket is https://issues.redhat.com/browse/KOGITO-3176 and is causing that:
- When deploying the trusty service, but not the explainability service - the trusty service is not starting because the kafka topic "trusty-explainability-request" does not exist (this topic is automatically created when deploying the explainability service. 
- When deploying the explainability service, but not the trusty service - the explainability service is not starting because the kafka topic "trusty-explainability-result" does not exist (this topic is created when deploying the trusty service).
- When deploying the explainability and trusty services both, everything works fine. 

I think it might be an architecture issue here as trusty and explainability services depend on each other. Some other services like Jobs Service creates the kafka topic: https://github.com/kiegroup/kogito-apps/blob/05cca086194fa14d80a725fb381feb7b586b3496/jobs-service/src/main/java/org/kie/kogito/jobs/service/stream/KafkaConfiguration.java#L86. I'm not saying that this is the right approach, but it needs further discussion and broader agreement. 

@r00ta I'm moving this discussion thread to the JIRA ticket.

ricardozanini:
The new Jenkins hasn't be configured yet, right @msmagnanijr ?

sutaakar:
Jenkins retest this please

ricardozanini:
Jenkins is out.

ricardozanini:
@Sgitario are you comfortable with these changes?

Sgitario:
> @Sgitario are you comfortable with these changes?

I tested these changes locally and it should be ok.

PR #577, opened 2020-09-18
[KOGITO-3382] Transitive dependency of Keycloak operator module missing

sutaakar:
Only keycloak-operator was updated manually, rest was updated by Go build.


Sgitario (approved):

go.mod
```diff
@@ -2,26 +2,26 @@ module github.com/kiegroup/kogito-cloud-operator
 
 require (
 	github.com/RHsyseng/operator-utils v0.0.0-20200304191317-2425bf382482
-	github.com/coreos/prometheus-operator v0.38.1-0.20200424145508-7e176fda06cc
+	github.com/coreos/prometheus-operator v0.40.0
```
ricardozanini, line 5: @sutaakar could you please double check why we need this dependency? Normally people are using their `log` package meaning to use the local `log` var

ricardozanini (commented):
I'm just curious about the prometheus dependency

go.mod
sutaakar: Actually it seems that we may not need the prometheus dependency declared. The `ServiceMonitor` seems to be created as a part of reconciliation.

sutaakar:
CI seems down again

go.mod
sutaakar: So, the dependency is needed for `ServiceMonitor` declaration.
Just it seems that we can remove `PrometheusCli` from https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/client/client.go#L62

spolti:
@msmagnanijr  ^

go.mod
sutaakar: created a separate PR for PrometheusCli removal - https://github.com/kiegroup/kogito-cloud-operator/pull/579

ricardozanini (approved):

sutaakar:
@ricardozanini I think this can be merged to fix operator build, functionality checked manually.

PR #590, opened 2020-09-29
[KOGITO-3475] Periodic reconciler error when KogitoInfra is deleted

sutaakar:

https://issues.redhat.com/browse/KOGITO-3475

### DIFF
--- a/pkg/controller/kogitoinfra/kogitoinfra_controller.go
+++ b/pkg/controller/kogitoinfra/kogitoinfra_controller.go
@@ -27,8 +27,10 @@ import (
 	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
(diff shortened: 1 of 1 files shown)


sutaakar:
related to changes done in https://github.com/kiegroup/kogito-cloud-operator/pull/550

spolti (approved):

ricardozanini (commented):
So if the KogitoInfra gets deleted, what happens with the binds in the services? We're removing the references?

vaibhavjainwiz:
@sutaakar 
This issue occurs because currently we are watching to all create/delete/update event on primary resources inside controller. In this case following thing happened : 

1. you had deleted `kogito_infra`
2. controller-runtime generates delete event for `kogito_infra`
3. KogitoInfra_controller receive delete event for `kogito_infra`
4. KogitoInfra_controller tries to fetch `kogito_infra` object and give error

There could be two ways to tackle this issue :

- Add null handling
- Don't listen for Delete event on primary resource

I am not sure which approach we should follow but we could aviod listening to delete event using code :
```
// Watch for changes to primary resource KogitoInfra
	pred := predicate.Funcs{
		DeleteFunc: func(e event.DeleteEvent) bool {
			return false
		},
	}
	err = c.Watch(&source.Kind{Type: &appv1alpha1.KogitoInfra{}}, &handler.EnqueueRequestForObject{}, pred)
```

vaibhavjainwiz:
> So if the KogitoInfra gets deleted, what happens with the binds in the services? We're removing the references? @vaibhavjainwiz

@ricardozanini 
Currently kogitoServices don't watch for `KogitoInfra` CR. We don't update env var and app prop of kogitoService if referred kogitoInfra removed.

sutaakar:
PR updated to use predicate approach

vaibhavjainwiz (approved):

ricardozanini:
We can use the predicate approach or fetch for the KogitoInfra object and return nil. Listen to the delete event is important to remove/react upon any finalizer we might have. Since we don't have one AFAIR, we can safely add the predicate for now.

ricardozanini (approved):

PR #591, opened 2020-09-30
[KOGITO-3375] - Make the operator to fetch the metrics endpoint to add Prometheus support

vaibhavjainwiz:

https://issues.redhat.com/browse/KOGITO-3375


vaibhavjainwiz:
@ricardozanini 
 Could you please review package structure. I am not clear How should I arrange these files. Code to make rest call is at below location :
https://github.com/vaibhavjainwiz/kogito-cloud-operator/blob/99eb0b61dd073e686c25521c6d15a5bba428f8b5/pkg/infrastructure/services/prometheus.go#L70

pkg/infrastructure/kogito_service.go
```diff
@@ -103,3 +105,18 @@ func getServiceEndpoints(client *client.Client, namespace string, serviceHTTPRou
...
+func FetchKogitoServiceURI(cli *client.Client, name, namespace string) (string, error) {
+	log.Debugf("Fetching kogito service instance URI.")
+	service := &v1.Service{}
+	if exits, err := kubernetes.ResourceC(cli).FetchWithKey(types.NamespacedName{Name: name, Namespace: namespace}, service); err != nil {
+		return "", err
+	} else if !exits {
+		return "", fmt.Errorf("service with name %s not exist for Kogito service instance in given namespace %s", name, namespace)
+	}
+	port := service.Spec.Ports[0]
+	uri := fmt.Sprintf("%s://%s:%d", port.Name, service.Name, port.TargetPort.IntVal)
```
ricardozanini, line 119: We should use the Kubernetes svc DNS name instead:
https://github.com/kubernetes/dns/blob/master/docs/specification.md#241---aaaaa-records

```go
// resolves to http://dataindex.mynamespace.svc.cluster.local:8080 for example
uri := fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, port.TargetPort.IntVal)
```

Also `portName` can be anything, let's stick with the standard HTTP protocol since it's an internal call.

That's because we are about to change the operator to be cluster scoped instead, meaning that the operator will be in different namespace of its operands.

pkg/infrastructure/services/prometheus.go
```diff
@@ -18,66 +18,151 @@ import (
...
+	if err != nil {
+		return false, err
+	}
+
+	url = url + getMonitoringPath(kogitoService.GetSpec().GetMonitoring())
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return false, err
+	}
+	if resp, err := client.HTTPCli.Do(req); err != nil {
```
ricardozanini, line 70: Here, instead of performing the request directly we can wait for the Deployment status of the service, if it's ready to receive connections we do it, otherwise just return without error and try in the next reconciliation loop.

The Deployment `status.availableReplicas` > 0 will tell you if it's ready to receive requests. This way we won't keep making failed requests to the deployed service.

pkg/client/client.go
```diff
@@ -59,6 +59,7 @@ type Client struct {
 	Discovery              discovery.DiscoveryInterface
 	DeploymentCli          appsv1.AppsV1Interface
 	KubernetesExtensionCli kubernetes.Interface
+	HTTPCli                restclient2.HTTPClient
```
ricardozanini, line 62: I believe we should not add this attribute to the `Client` structure since it's just a `http.Client` instance dancing around the functions. It's safer to create it once we need it, plus it's not associated with any k8s client whatsoever.

pkg/client/restclient/http_client.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package restclient
+
+import "net/http"
+
+// HTTPClient interface
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
```
ricardozanini, line 22: I believe we don't need this. Just interact with the `http` package directly. It's designed to be this way.

ricardozanini (changes requested):
Looks good! Just some comments about the internal implementation approach. I had another idea to implement this to reuse the "service is ready to receive calls / svc discovery" part, but it's better to bring this offline in a different PR and showcase to you guys. @Kevin-Mok you will likely to reuse some of this code for [KOGITO-3376](https://issues.redhat.com/browse/KOGITO-3376)

Sgitario (approved):

sutaakar:
@vaibhavjainwiz Please rebase, will do a review once Zanini's comments are addressed.

pkg/client/restclient/http_client.go
vaibhavjainwiz: `HTTPClient` interface helps me to mock http call. If we directly initialize `http.Client` in calling method then we will not be able to mock its call to run testcase. This is a same reason I have to put `HTTPCli` inside `client.go`

ricardozanini: You can use the `httptest` package directly. I understand that the Interface approach is used out there, so that might be ok for now. This will be refactored soon.

Now, about the attribute in the client package I'd rather to remove it from there. That package is meant to be a k8s API client handler only.

pkg/infrastructure/kogito_service.go
vaibhavjainwiz: done

pkg/client/client.go
vaibhavjainwiz: I had removed it from `Client` and directly call `http.Get` when required

pkg/client/restclient/http_client.go
vaibhavjainwiz: done

pkg/infrastructure/services/prometheus.go
vaibhavjainwiz: done

pkg/apis/app/v1alpha1/kogitoservices_types.go
```diff
@@ -188,6 +189,12 @@ type KogitoServiceSpec struct {
 	// Infra provides list of dependent KogitoInfra objects.
 	// +optional
 	Infra []string `json:"infra,omitempty"`
+
+	// Create Service monitor instance to connect with Monitoring service
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Monitoring"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:label"
```
ricardozanini, line 196: This won't work since it's a data structure and not a String. I need to take a look at this actually to see if something changed in the new OLM version.

pkg/infrastructure/kogito_service.go
```diff
@@ -103,3 +105,20 @@ func getServiceEndpoints(client *client.Client, namespace string, serviceHTTPRou
...
+	log.Debugf("Fetching kogito service instance URI.")
+	service := &v1.Service{}
+	if exits, err := kubernetes.ResourceC(cli).FetchWithKey(types.NamespacedName{Name: name, Namespace: namespace}, service); err != nil {
+		return "", err
+	} else if !exits {
+		return "", fmt.Errorf("service with name %s not exist for Kogito service instance in given namespace %s", name, namespace)
+	}
+	port := service.Spec.Ports[0]
+	// resolves to http://dataindex.mynamespace.svc.cluster.local:8080 for example
+	uri := fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", service.Name, service.Namespace, port.TargetPort.IntVal)
```
ricardozanini, line 120: During the weekend I made some tests around this and we can surpress the `svc.cluster.local` part. People can change this name in the kube configuration to differentiate between zones on hybrid installations, so I believe we need to remove it also.

Just this will work:

```suggestion
	uri := fmt.Sprintf("http://%s.%s:%d", service.Name, service.Namespace, port.Port)
```

Also `TargetPort` it's the port which the svc target in the container, not the exposed one. To use the exposed port we need to use `service.Port` instead. Soon we will use the default HTTP port and adding the port to the service won't be needed any more. See: https://issues.redhat.com/browse/KOGITO-3516

pkg/infrastructure/kogito_service.go
```diff
@@ -103,3 +105,20 @@ func getServiceEndpoints(client *client.Client, namespace string, serviceHTTPRou
...
+	//uri := fmt.Sprintf("%s://%s:%d", port.Name, service.Name, port.TargetPort.IntVal)
```
ricardozanini, line 121:
```suggestion

```

pkg/infrastructure/services/prometheus.go
```diff
@@ -18,66 +18,157 @@ import (
...
+}
 
-		serviceMonitorLabels := make(map[string]string)
-		serviceMonitorLabels["name"] = operator.Name
-		serviceMonitorLabels[framework.LabelAppKey] = kogitoRuntime.GetName()
+// createServiceMonitor create ServiceMonitor used for scraping by prometheus for kogito service
+func createServiceMonitor(cli *client.Client, kogitoService v1alpha1.KogitoService, scheme *runtime.Scheme) (*monv1.ServiceMonitor, error) {
+	monitoring := kogitoService.GetSpec().GetMonitoring()
+	endPoint := monv1.Endpoint{}
+	endPoint.TargetPort = &intstr.IntOrString{IntVal: getServiceHTTPPort(kogitoService)}
```
ricardozanini, line 118: I haven't found the `getServiceHTTPPort` function in this PR, but just make sure we are returning the `service.Port` and not `service.TargetPort`.

ricardozanini (approved):
Looks good, one minor thing only. Many thanks!

pkg/infrastructure/services/deployer.go
```diff
@@ -205,6 +205,7 @@ func (s *serviceDeployer) Deploy() (reconcileAfter time.Duration, err error) {
 		}
 		s.generateEventForDeltaResources("Removed", resourceType, delta.Removed)
 	}
+	reconcileAfter, err = configurePrometheus(s.client, s.instance, s.scheme)
```
sutaakar, line 208:
```suggestion
	// In case required resource needs reconciliation
	if reconcileAfter > 0 {
		return
	}
	reconcileAfter, err = configurePrometheus(s.client, s.instance, s.scheme)
```
This will make sure that Prometheus won't overwrite reconciliation period.

pkg/infrastructure/services/prometheus.go
sutaakar: IMHO the `getServiceHTTPPort` returns a port where Kogito application is running (a bit unfortunate naming) - https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/infrastructure/services/deployment.go#L85. 
Needs to be changed here to use service port.

sutaakar:
Small BDD adjustment for these changes, can be added as part of this PR - https://github.com/vaibhavjainwiz/kogito-cloud-operator/pull/1

pkg/infrastructure/services/prometheus.go
ricardozanini: Yes, this function is getting the port from the spec, getting from the service directly will be more reliable.

pkg/infrastructure/kogito_service.go
ricardozanini: Depends on: #598 to remove the port part.

sutaakar:
Checked the built operator image on OCP, works as expected.

pkg/infrastructure/services/deployment.go
```diff
@@ -103,3 +105,12 @@ func setHTTPPortInEnvVar(httpPort int32, kogitoService v1alpha1.KogitoService) {
 	modifiedEnv := framework.EnvOverride(envs, httpPortEnvVar)
 	kogitoService.GetSpec().SetEnvs(modifiedEnv)
 }
+
+func isDeploymentAvailable(cli *client.Client, name, namespace string) (bool, error) {
```
spolti, line 109: will this be called only during reconciliation?

spolti:
looks good, just left a question, and please take a look on the failing test.

pkg/infrastructure/kogito_service.go
vaibhavjainwiz: Updated to suggested change

pkg/infrastructure/kogito_service.go
vaibhavjainwiz: removed commented code

pkg/infrastructure/services/prometheus.go
vaibhavjainwiz: @ricardozanini @sutaakar 
Previously I done so because exposed port are always equal to kogitoService port but now after implementation of KOGITO-3516, exposed port always be a default port. So rather then loading Service object and fetch port from it, could I directly use default http?
We could use same approach to create kogitoServiceURI as well.
https://github.com/kiegroup/kogito-cloud-operator/blob/605cb71f5d01b94de56516aaa230ec37232fb846/pkg/infrastructure/kogito_service.go#L110

pkg/infrastructure/services/deployer.go
vaibhavjainwiz: I that case we could move this check to `createRequiredResources` itself so that if `createRequiredResources` requires reconcilation then it will return from there and don't make a call `getDeployedResources`

pkg/infrastructure/services/deployment.go
vaibhavjainwiz: yes @spolti . We call this function while configuring prometheus as pre-check. Operator goes to configure prom only when deployment is ready to serve request because operator have to application's `/metrics`  endpoint to check whether application support prom or not

https://github.com/kiegroup/kogito-cloud-operator/blob/640c94390db7cdef2f2e95f258c51013195ebae5/pkg/infrastructure/services/prometheus.go#L42

pkg/infrastructure/services/prometheus.go
ricardozanini: +1 @vaibhavjainwiz

vaibhavjainwiz: done

pkg/apis/app/v1alpha1/kogitoservices_types.go
vaibhavjainwiz: I had removed these extra comments. Just keep first line

Kaitou786 (approved):

radtriste (approved):
lgtm
Please merge PR adjustment from @sutaakar : https://github.com/vaibhavjainwiz/kogito-cloud-operator/pull/1

vaibhavjainwiz:
> lgtm
> Please merge PR adjustment from @sutaakar : [vaibhavjainwiz#1](https://github.com/vaibhavjainwiz/kogito-cloud-operator/pull/1)

done

sutaakar:
@vaibhavjainwiz Sorry for wrong information in previous comment....
Right now the ServiceMonitor created by Kogito operator points to wrong port, causing Prometheus to miss Kogito application.
There are two options to specify what port should be used 
- port (https://github.com/prometheus-operator/prometheus-operator/blob/master/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L179) pointing to the service port
- targetPort (https://github.com/prometheus-operator/prometheus-operator/blob/master/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L239) pointing to the pod's port

Right now this implementation use `targetPort`, however it is passing service port there, which is wrong.
So we can either adjust the implementation to use `port` or change the passed port to pod's one.

ricardozanini:
> Right now this implementation use `targetPort`, however it is passing service port there, which is wrong.
> So we can either adjust the implementation to use `port` or change the passed port to pod's one.

I'd rather use the service one since we can abstract the exposed port in the container.

Thinking about it now, I think we can get rid of the `spec.httpPort` attribute. There's no practical use. The default http port being served by the service addresses all the issues and make it way simpler to do svc discovery/injection.

@sutaakar @spolti @vaibhavjainwiz wdyt?

spolti:
@ricardozanini  if there's no use cas for this property anymore, +1 for removing it.

spolti:
@vaibhavjainwiz is the problem reported by @sutaakar addressed?

sutaakar:
@ricardozanini fine for me

vaibhavjainwiz:
> > Right now this implementation use `targetPort`, however it is passing service port there, which is wrong.
> > So we can either adjust the implementation to use `port` or change the passed port to pod's one.
> 
> I'd rather use the service one since we can abstract the exposed port in the container.
> 
> Thinking about it now, I think we can get rid of the `spec.httpPort` attribute. There's no practical use. The default http port being served by the service addresses all the issues and make it way simpler to do svc discovery/injection.
> 
> @sutaakar @spolti @vaibhavjainwiz wdyt?

Prometheus operator provides both options to create `Service monitor` and `Pod Monitor`. Kogito-operator creates service monitor objects to expose services for scraping so it make sence to provide service pod. 

@ricardozanini I am agree with @sutaakar point to take spec.httpPort as input.

ricardozanini:
> @ricardozanini fine for me
> 
> 
> 
> Edit: How do we recognize then what port is used by Kogito application? IMHO the `spec.httpPort` is mainly used for this, to define what port should be set in Deployment as ContainerPort

The question is: why do we need to recognize this? The port should always be defined on 8080 and exposed by the image. This should not change, there's no practical use case for this. 

The operator creates a service exposing this port **through the default HTTP one**. Other services within the cluster will just call the application. 

That's my point. There's no use to change the port, no use case. It's there just because. :)

vaibhavjainwiz:
> > @ricardozanini fine for me
> > Edit: How do we recognize then what port is used by Kogito application? IMHO the `spec.httpPort` is mainly used for this, to define what port should be set in Deployment as ContainerPort
> 
> The question is: why do we need to recognize this? The port should always be defined on 8080 and exposed by the image. This should not change, there's no practical use case for this.
> 
> The operator creates a service exposing this port **through the default HTTP one**. Other services within the cluster will just call the application.
> 
> That's my point. There's no use to change the port, no use case. It's there just because. :)

Its spring functionality which allow user to set his desired port through `service.port=9090` prop in `application.properties` file. Atfer setting this property, every time user start its service through `java -jar myapplication.jar`, its embedded tomcat server will start at user provided port. 
Currently I can't think of any specific reason to explicitly change its default port. In my previous project I used to explictly change port values because we have to run multiple java service on our local machine to test some functionality and we can't run containers on dev system due to hardware constraints. 
But in QA/Prod env we always run all services at 8080 port in containers.

ricardozanini:
> Its spring functionality which allow user to set his desired port through `service.port=9090` prop in `application.properties` file. Atfer setting this property, every time user start its service through `java -jar myapplication.jar`, its embedded tomcat server will start at user provided port.
> Currently I can't think of any specific reason to explicitly change its default port. In my previous project I used to explictly change port values because we have to run multiple java service on our local machine to test some functionality and we can't run containers on dev system due to hardware constraints.
> But in QA/Prod env we always run all services at 8080 port in containers.

Thanks for sharing your experience, @vaibhavjainwiz. But in this containers scenarios we won't clash ports since there's only one process running within the container and we control it. In case of sidecars injections to share the network from a pod, this other injection pod would require a different port, but again controlled by the operator. There's no use case to change this port any more, only on local environments where you need to share the network resources.

There's no practical usage for this and we are only adding unnecessary code/complexity to our use case. The container will be deployed and served at port 80 by the Kubernetes service. Consuming the Kogito service within the cluster will be straightforward like: `http://name.namespace`.

vaibhavjainwiz:
> Thanks for sharing your experience, @vaibhavjainwiz. But in the container scenario we won't clash ports since there's only one process running within the container and we control it. In case of sidecars injections to share the network from a pod, this other injection pod would require a different port, but again controlled by the operator. There's no use case to change this port any more, only on local environments where you need to share the network resources.
> 
> There's no practical usage for this and we are only adding unnecessary code/complexity to our use case. The container will be deployed and served at port 80 by the Kubernetes service. Consuming the Kogito service within the cluster will be straightforward like: `http://name.namespace`.
> 
> And I'd say that we should always force 8080 even if a external configuration like `service.port` is passed since we are forcing the port to be exposed on 8080 in the images:
> 
> ```dockerfile
> EXPOSE 8080
> ```

@ricardozanini Totally agree with your. Inside Image we control service port through env var `HTTP_PORT` but image default exposed port is always 8080.

https://github.com/vaibhavjainwiz/kogito-images/blob/f229b91aeeec09c12a20c3a68010a8f7d40ddde0/modules/kogito-quarkus/added/launch/kogito-quarkus-native.sh#L13-L12

ricardozanini:
Thanks for bringing the link, @vaibhavjainwiz :D

Opened: https://issues.redhat.com/browse/KOGITO-3568

vaibhavjainwiz:
> @vaibhavjainwiz Sorry for wrong information in previous comment....
> Right now the ServiceMonitor created by Kogito operator points to wrong port, causing Prometheus to miss Kogito application.
> There are two options to specify what port should be used
> 
> * port (https://github.com/prometheus-operator/prometheus-operator/blob/master/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L179) pointing to the service port
> * targetPort (https://github.com/prometheus-operator/prometheus-operator/blob/master/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L239) pointing to the pod's port
> 
> Right now this implementation use `targetPort`, however it is passing service port there, which is wrong.
> So we can either adjust the implementation to use `port` or change the passed port to pod's one.

vaibhavjainwiz:
> > > Right now this implementation use `targetPort`, however it is passing service port there, which is wrong.
> > > So we can either adjust the implementation to use `port` or change the passed port to pod's one.
> > 
> > 
> > I'd rather use the service one since we can abstract the exposed port in the container.
> > Thinking about it now, I think we can get rid of the `spec.httpPort` attribute. There's no practical use. The default http port being served by the service addresses all the issues and make it way simpler to do svc discovery/injection.
> > @sutaakar @spolti @vaibhavjainwiz wdyt?
> 
> Prometheus operator provides both options to create `Service monitor` and `Pod Monitor`. Kogito-operator creates service monitor objects to expose services for scraping so it make sence to provide service pod.
> 
> @ricardozanini I am agree with @sutaakar point to take spec.httpPort as input.

@sutaakar Now our expose port is 80 so don't need to define it explicitly inside `ServiceMonitor`. I tested it on my local machine, its generating metrics now.
image

ricardozanini:
@vaibhavjainwiz this might be a silly question, but can we change the "endpoint" field to reflect the actual port (80)? Or this is the pod IP and port?

vaibhavjainwiz:
> @vaibhavjainwiz this might be a silly question, but can we change the "endpoint" field to reflect the actual port (80)? Or this is the pod IP and port?

I had already tried it by explictly setting `EndPoint.Port = 80` but it didn't work. 

its description dictates as :

```
// Name of the service port this endpoint refers to. Mutually exclusive with targetPort.
Port string `json:"port,omitempty"`
```

spolti (approved):

sutaakar (approved):
Looks good to me, lets wait for PR check to finish

ricardozanini:
@spolti can be merged

PR #598, opened 2020-10-05
[KOGITO-3516] Services now listen to default HTTP port

ricardozanini:

See: https://issues.redhat.com/browse/KOGITO-3516

Now any service can be called via `http://name.namespace` from the cluster.


sutaakar (approved):
Looks good to me

spolti:
we would need a followup for ssl.

ricardozanini:
> we would need a followup for ssl.

yes indeed, but this is internal services calls only, so will be a different topic...

PR #599, opened 2020-10-05
[KOGITO-3543] - Wrong environment values for Kogito Services after re…

vaibhavjainwiz:
…concile cycles

Jira issue : https://issues.redhat.com/browse/KOGITO-3543

### DIFF
--- a/pkg/infrastructure/services/deployment.go
+++ b/pkg/infrastructure/services/deployment.go
@@ -45,6 +45,10 @@ func createRequiredDeployment(service v1alpha1.KogitoService, resolvedImage stri
 	}
 	labels[framework.LabelAppKey] = service.GetName()
 
+	// clone env var slice so that any changes in deployment env var should not reflect in kogitoInstance env var
+	env := make([]corev1.EnvVar, len(service.GetSpec().GetEnvs()))
+	copy(env, service.GetSpec().GetEnvs())
+
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{Name: service.GetName(), Namespace: service.GetNamespace(), Labels: labels},
 		Spec: appsv1.DeploymentSpec{
@@ -63,7 +67,7 @@ func createRequiredDeployment(service v1alpha1.KogitoService, resolvedImage stri
 									Protocol:      corev1.ProtocolTCP,
 								},
 							},
-							Env:             service.GetSpec().GetEnvs(),
+							Env:             env,
 							Resources:       service.GetSpec().GetResources(),
 							LivenessProbe:   probes.liveness,
 							ReadinessProbe:  probes.readiness,


sutaakar (approved):

pkg/infrastructure/services/deployment.go
```diff
@@ -45,6 +45,10 @@ func createRequiredDeployment(service v1alpha1.KogitoService, resolvedImage stri
 	}
 	labels[framework.LabelAppKey] = service.GetName()
 
+	// clone env var slice so that any changes in deployment env var should not reflect in kogitoInstance env var
```
ricardozanini, line 48: Hi Jain! Thanks for debugging this issue. 

I'm curious about this. Why the changes in the deployment env var are impacting our CR? We should not write the env var from deployment back to the CR. The CR state impacts the deployment, not the other way around.

vaibhavjainwiz: we don't write them back. Actually issue occurs because of special bahaviour of slice in golang. In golang generally argument are passed by value but with slice pass sort of reference. 

https://medium.com/swlh/golang-tips-why-pointers-to-slices-are-useful-and-how-ignoring-them-can-lead-to-tricky-bugs-cac90f72e77b

ricardozanini: I know about the slices. My question is regarding the CR itself. If we are always overwriting the deployment with the env set by the CR, we should not see this behavior.

radtriste (approved):

Sgitario (approved):

ricardozanini:
/jenkins test

@Sgitario have you verified, can we merge it?

ricardozanini (approved):

pkg/infrastructure/services/deployment.go
Kevin-Mok: > If we are always overwriting the deployment with the env set by the CR, we should not see this behavior.

This is a curious issue. How are the env vars from the CR changing such that when you do `GetEnvs()` from it, there are missing/duplicate vars? :thinking:

spolti (approved):

pkg/infrastructure/services/deployment.go
vaibhavjainwiz: @ricardozanini @Kevin-Mok Env var of CR changes even we are not directly setting its value because we Env var of CR and deployment refers to same slice array internally.

```
&appsv1.Deployment{
....
....
Env:             service.GetSpec().GetEnvs()  // assign same slice reference to deployment
....
}

```
and we do changes Env var of deployment to apply KogitoInfra properties. Which causes indirect changes to Env var of CR as well. It happend untill slice didn't expands to it initially assigned size. Once slice reaches its max limit it create new array of double size and after it both Env var of CR and deployment refers to two different slices. 
Because we are providing 9 Env Var in CR definition which causes to initialize slice of 16 element, that the reason why Env var of CR extends to size 16 and Env var of deployment goes upto 17. 
https://github.com/kiegroup/kogito-cloud-operator/blob/666f44856e85c98d8e34cd466d7dfe1f59e11356/pkg/infrastructure/services/resources.go#L80


There are multiple issue in code which we could fix to prevent this type issues in future. like :
1. Operator should never update any value in Spec like this 
https://github.com/kiegroup/kogito-cloud-operator/blob/c6913eeab841551be76bb2bd08bd1cfd273d6663/pkg/infrastructure/services/deployment.go#L108
2. Operator should only update status of CR rather then updating complete object inside updateStatus function.
https://github.com/kiegroup/kogito-cloud-operator/blob/666f44856e85c98d8e34cd466d7dfe1f59e11356/pkg/infrastructure/services/deployer.go#L289

radtriste:
@ricardozanini @sutaakar @vaibhavjainwiz @Sgitario shouldn't we have that one for 0.16 milestone ? (I see 0.17 ...)

ricardozanini:
> @ricardozanini @sutaakar @vaibhavjainwiz @Sgitario shouldn't we have that one for 0.16 milestone ? (I see 0.17 ...)

The PR is on 0.17, that's why I set the milestone there. I'm waiting for QE verification so @vaibhavjainwiz would open a cherry pick. Actually I didn't have the time to look further, but this bug is bugging me.

pkg/infrastructure/services/deployment.go
ricardozanini: Is the KogitoInfra controller who is updating the Deployment or the service controller?

vaibhavjainwiz: Service controller update the deployment.

https://github.com/kiegroup/kogito-cloud-operator/blob/bedafda901c0960a039b6eb5f1570666e70b1e79/pkg/infrastructure/services/resources.go#L79

Sgitario:
I've just wanted to confirm that I did test these changes and now it's working as expected.

ricardozanini:
@vaibhavjainwiz can you please cherry-pick to 0.16.x branch?

pkg/infrastructure/services/deployment.go
ricardozanini: Opened:
https://issues.redhat.com/browse/KOGITO-3556

vaibhavjainwiz:
> @vaibhavjainwiz can you please cherry-pick to 0.16.x branch?

I had raised below PR to cherry-pick 
https://github.com/kiegroup/kogito-cloud-operator/pull/603

PR #600, opened 2020-10-05
[KOGITO-3094] - Enable persistence selection in Data Index controller

spolti:
Signed-off-by: spolti <fspolti@redhat.com>

See: https://issues.redhat.com/browse/KOGITO-3094

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 - [KOGITO-2905](https://issues.redhat.com/browse/KOGITO-2905) - Change in image definition in the CRDs
 - [KOGITO-3039](https://issues.redhat.com/browse/KOGITO-3039) - Agnostic infrastructure layer configuration on Operator
 - [KOGITO-2113](https://issues.redhat.com/browse/KOGITO-2113) - Simplify binary build deployment
+- [KOGITO-3094](https://issues.redhat.com/browse/KOGITO-3094) - Enable persistence selection in Data Index controller
 
 
 ## Bug Fixes
--- a/pkg/controller/kogitodataindex/kogitodataindex_controller.go
+++ b/pkg/controller/kogitodataindex/kogitodataindex_controller.go
@@ -25,9 +25,8 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"path"
-
 	"k8s.io/apimachinery/pkg/runtime"
+	"path"
 
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -175,6 +174,26 @@ func (r *ReconcileKogitoDataIndex) onDeploymentCreate(cli *client.Client, deploy
 		log.Warnf("No container definition for service %s. Skipping applying custom Data Index deployment configuration", kogitoService.GetName())
 	}
 
+	dataIndex := kogitoService.(*appv1alpha1.KogitoDataIndex)
+	// set to the default.
+	persistenceProvider := infrastructure.DefaultDataIndexPersistence
+	if propValue, ok := dataIndex.Spec.Config[infrastructure.DataIndexPersistenceTypeProp]; ok {
+		foundPersistenceType := infrastructure.PersistenceProvider(propValue)
+		if _, ok := infrastructure.PersistenceProviders[foundPersistenceType]; ok {
+			persistenceProvider = infrastructure.PersistenceProviders[foundPersistenceType]
+		} else {
+			log.Warnf("the persistence type %s is invalid, valid are %s, defaulting to %s",
+				foundPersistenceType,
+				infrastructure.PersistenceProviders,
+				persistenceProvider)
+		}
+	} else {
+		log.Infof("the % property was not found, defaulting persistence provider to %s",
+			infrastructure.DataIndexPersistenceTypeProp, persistenceProvider)
+	}
+
+	log.Infof("injecting %s with the persistence provider %s.", infrastructure.DataIndexPersistenceEnv, persistenceProvider)
+	framework.SetEnvVar(infrastructure.DataIndexPersistenceEnv, persistenceProvider, &deployment.Spec.Template.Spec.Containers[0])
 	return nil
 }
 
--- a/pkg/controller/kogitodataindex/kogitodataindex_controller_test.go
+++ b/pkg/controller/kogitodataindex/kogitodataindex_controller_test.go
@@ -80,6 +80,134 @@ func TestReconcileKogitoDataIndex_Reconcile(t *testing.T) {
 	}
 }
 
+func TestReconcileKogitoDataIndex_ReconcileDefaultPersistenceProvider(t *testing.T) {
+	ns := t.Name()
+
+	instance := &v1alpha1.KogitoDataIndex{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-data-index",
+			Namespace: ns,
+		},
+		Spec: v1alpha1.KogitoDataIndexSpec{
+			KogitoServiceSpec: v1alpha1.KogitoServiceSpec{},
+		},
+	}
+
+	cli := test.CreateFakeClient([]runtime.Object{instance}, nil, nil)
+	r := &ReconcileKogitoDataIndex{
+		client: cli,
+		scheme: meta.GetRegisteredSchema(),
+	}
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// basic checks
+	_, err := r.Reconcile(req)
+	if err != nil {
+		t.Fatalf("reconcile: (%v)", err)
+	}
+
+	deployment := &appsv1.Deployment{}
+	exists, _ := kubernetes.ResourceC(cli).FetchWithKey(types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, deployment)
+	assert.True(t, exists)
+	assert.Contains(t, deployment.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+		Name:  infrastructure.DataIndexPersistenceEnv,
+		Value: "infinispan",
+	})
+}
+
+func TestReconcileKogitoDataIndex_ReconcileMongodbPersistenceProvider(t *testing.T) {
+	ns := t.Name()
+
+	instance := &v1alpha1.KogitoDataIndex{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-data-index",
+			Namespace: ns,
+		},
+		Spec: v1alpha1.KogitoDataIndexSpec{
+			KogitoServiceSpec: v1alpha1.KogitoServiceSpec{
+				Config: map[string]string{
+					infrastructure.DataIndexPersistenceTypeProp: infrastructure.PersistenceProviders["mongodb"],
+				},
+			},
+		},
+	}
+
+	cli := test.CreateFakeClient([]runtime.Object{instance}, nil, nil)
+	r := &ReconcileKogitoDataIndex{
+		client: cli,
+		scheme: meta.GetRegisteredSchema(),
+	}
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// basic checks
+	_, err := r.Reconcile(req)
+	if err != nil {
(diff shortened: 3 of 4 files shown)


cmd/kogito/command/install/data_index.go
```diff
@@ -92,6 +97,16 @@ func (i *installDataIndexCommand) Exec(cmd *cobra.Command, args []string) error
 		return err
 	}
 
+	persistenceProvider := v1alpha1.InfinispanPersistenceProvider
+	if i.flags.kogitoDataIndexPersistence == "mongodb" {
```
ricardozanini, line 101: Can you move this value to a constant?

```suggestion
	if i.flags.kogitoDataIndexPersistence == string(v1alpha1.MongoDBPersistenceProvider) {
```

pkg/controller/kogitodataindex/kogitodataindex_controller.go
```diff
@@ -175,6 +175,14 @@ func (r *ReconcileKogitoDataIndex) onDeploymentCreate(cli *client.Client, deploy
 		log.Warnf("No container definition for service %s. Skipping applying custom Data Index deployment configuration", kogitoService.GetName())
 	}
 
+	dataIndex := kogitoService.(*appv1alpha1.KogitoDataIndex)
+	if len(dataIndex.Spec.KogitoDataIndexPersistenceProvider) <= 0 || (dataIndex.Spec.KogitoDataIndexPersistenceProvider != appv1alpha1.InfinispanPersistenceProvider &&
+		dataIndex.Spec.KogitoDataIndexPersistenceProvider != appv1alpha1.MongoDBPersistenceProvider) {
```
ricardozanini, line 180: Having these values into a map would be easier to understand, something like:

```go
if _, ok := providers[dataIndex.Spec.KogitoDataIndexPersistenceProvider]; !ok {
    // invalid
}
```

ricardozanini (commented):
Although I understand the need of such "switch", we are moving away from having such flags/structures in our interface.

The operator is not doing anything here, but set an env var. So in this scenario, I'd say that an user could do:

```yaml
kind: KogitoDataIndex
metadata:
  name: data-index
spec:
  # config attribute to be introduced by https://issues.redhat.com/browse/KOGITO-3273, otherwise JAVA_ARGS as env var will do.
  config:
     kogito.persistence.type: mongodb
     mongodb.url: tcp://11.11.1.11:11
  replicas: 1
```

This way our interface is clear, the state of the object is clearly defined and this `config` attribute will turn into a ConfigMap mounted to the Data Index service.

vaibhavjainwiz:
@ricardozanini Please correct me if I am wrong. 
Actually I have completely different opinion about its implementation approach. In KOGITO-3039 we have completely seperated infra layer from KogitoService. KogitoService could bind to any infra like Infinispan, Redis, Mongo or Kafka just by provides its relevant KogitoInfra references. 
In our case if user want to bind its data-index service with MongoDB then it could create relevant KogitoInfra CR and just provides its references as dependent infra inside data-index CR. To accomplish this we should create new implementation of KogitoInfra as we do for Infinispan/kafka and Keycloak. 
Now point come when we need some identier in Kogito-images to identify which jar operator should use(Infinispan/Mongo). Currently we accomplish this by setting identifier inside Env var of deployment which we fetch in images. With new approach we should keep doing same thing, just this `DATA_INDEX_PERSISTENCE` env var should be set in KogitoInfra CR which will automatically binds to Env var of deployment.

ricardozanini:
@vaibhavjainwiz yes, that's what I said in my comment. I talked to @spolti last night, he will refactor.

About the `KogitoInfra`, it's meant to bind third party CRs with ours, so in this particular scenario here, it will not fit. Not until we have integration with MongoDB operators out there.

For now, the operator will set the env to the image based in the configuration passed through the CR.

pkg/controller/kogitodataindex/kogitodataindex_controller.go
spolti: ok

spolti:
I'll remove the comments after the new approach is ready, please ignore them :)

vaibhavjainwiz:
@ricardozanini Untill that config approach is not ready, could we ask user to directly provides `DATA_INDEX_PERSISTENCE` property in env var directly like 
```
kind: KogitoDataIndex
metadata:
  name: data-index
spec:
  env: 
     - name: DATA_INDEX_PERSISTENCE
       value: mongodb  
  replicas: 1
```
with this change operator don't need to do following :

- first fetch `JAVA_OPTIONS` from envs, 
- extract `-Dkogito.persistence.type` property from `JAVA_OPTIONS` 
- Set `DATA_INDEX_PERSISTENCE` again in envs

@spolti I have doubt abouts required properties. Here we are discussing about two differents `kogito.persistence.type` and `DATA_INDEX_PERSISTENCE`. Do we require both them ? I checked kogito-images implementation to support mongoDB persistence, as I understand it only require `DATA_INDEX_PERSISTENCE` property as identifier to pick between data-index persistence jar.

@ricardozanini @sutaakar With configMap implementation, all properties user provides under `config` are appended to its `application.properties` only not in its envs. But kogito-images expecting to receive `DATA_INDEX_PERSISTENCE` prop in its env. I am not sure if it could work?

sutaakar:
IMHO we don't have to care about `kogito.persistence.type` as it is set in the specific Data index jar. So we need just to choose proper runner on image side. Simplest way could be to use env variable as it is easy to use it in case user runs the Data index image by itself outside of cluster.

spolti:
My understanding on the "kogito.persistence.type" system property is that it will be used for kogito app that has the dataindex-addons, right? 
After the config property is added to the spec then it will be used by a kogito the app and by the data index to define the correct persistence provider.

if we decide to only use the DATA_INDEX_PERSISTENCE env, then when user wants to switch to mongo db, he will have to make 2 configurations, set the system property and this env.

This is my understanding, please correct me if am I wrong.

sutaakar:
After checking the documentation it seems that `kogito.persistence.type` is needed for any Kogito application/service in case it will use other than Infinispan persistence.

vaibhavjainwiz:
> After checking the documentation it seems that `kogito.persistence.type` is needed for any Kogito application/service in case it will use other than Infinispan persistence.

As `kogito.persistence.type` is applicable for all Kogito application/service, we could receive it under Spec of `KogitoServiceSpec`  rather then in EnvVar.
One question, what happen if user provide `kogito.persistence.type=mongodb` in `travels` service but forget to mention it in `visas`?

vaibhavjainwiz:
> I'll remove the comments after the new approach is ready, please ignore them :)

@spolti I had raised PR(https://github.com/kiegroup/kogito-cloud-operator/pull/610) for new approach. Could you please remove commented code now

ricardozanini:
This persistence switch thing should be an unique use case for the Data Index. Since we already have `KogitoInfra` specifying the `Infinispan` implementation, we can use that as a hint to configure the image with the right ENV var. But that's not reliable since I can use whatever server I have at my disposal, like onprem architectures.

Having said that, I'm not a fan of adding a `persistenceType` in the CRD, but would make sense for Data Index ONLY since we need to change the image based on that.

It's not ideal, but I think having two images in this scenario would make WAY more sense:

```
kogito-data-index-mongodb
kogito-data-index-infinispan
```

Setting the image in the Data Index CR `spec.image` and the configuration into `spec.config`.

spolti:
>> One question, what happen if user provide kogito.persistence.type=mongodb in travels service but forget to mention it in visas?


> This persistence switch thing should be an unique use case for the Data Index. Since we already have `KogitoInfra` specifying the `Infinispan` implementation, we can use that as a hint to configure the image with the right ENV var. But that's not reliable since I can use whatever server I have at my disposal, like onprem architectures.
> 
> Having said that, I'm not a fan of adding a `persistenceType` in the CRD, but would make sense for Data Index ONLY since we need to change the image based on that.
> 
> It's not ideal, IMO. I think having two images in this scenario would make WAY more sense:
> 
> ```
> kogito-data-index-mongodb
> kogito-data-index-infinispan
> ```
> 
> Setting the image in the Data Index CR `spec.image` and the configuration into `spec.config`. Done. No conditionals. And the image would have its purpose on itself.
> 
> Otherwise, we just document that the env variable should be set to elect the persistence.

Not sure if I agree with it, it will increase the number of images we'll have to maintain, if data index supports other types of persistence, what you are proposing new images will need to be created.

The initial idea was use _persistenceType_ on spec, which now, we'll be moving to config.spec and maybe keep the support for envs, just in case, remove the env from operator at all.
Using only the spec.config would be a single point of configuration instead have to set the data-index image and the property for the apps.

spolti:
@vaibhavjainwiz @sutaakar @ricardozanini can you please take a look again?
Once we have it merged I'll create a PR to update the docs.

RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@
 - [KOGITO-2905](https://issues.redhat.com/browse/KOGITO-2905) - Change in image definition in the CRDs
 - [KOGITO-3039](https://issues.redhat.com/browse/KOGITO-3039) - Agnostic infrastructure layer configuration on Operator
 - [KOGITO-2113](https://issues.redhat.com/browse/KOGITO-2113) - Simplify binary build deployment
+- [KOGITO-3094](https://issues.redhat.com/browse/KOGITO-3094) - Enable persistence selection in Data Index controller
```
ricardozanini, line 7: You might delete the rest =D

ricardozanini:
> Not sure if I agree with it, it will increase the number of images we'll have to maintain, if data index supports other types of persistence, what you are proposing new images will need to be created.

How complex/hard is to maintain a sibling image? :D I believe it will be way easier than additional code to the operator.  


> The initial idea was use _persistenceType_ on spec, which now, we'll be moving to config.spec and maybe keep the support for retrieving it from JAVA_OPTS env, or remove it and rely only on the spec.config.
> Using only the spec.config would be a single point of configuration instead have to set the data-index image and the property for the apps.

This idea was ok at once, but an user can config a Kogito Service via:

- env variables set in the `spec.env` attribute
- any java properties that can be read by the application via `spec.config` attribute
- a custom ConfigMap attached to the CR via `spec.configMapProperties`

So we will read everything looking for this property? Since it's an application only property (not related to the image), I don't think it makes sense reading from `spec.config`. :thinking: 

Of course, we can always use docs to emphasize that Data Index will ONLY switch the persistence type if:

```yaml
spec:
    config:
    - kogito.persistence.type: mongodb
```

will be set.

But again, these properties suppose to be set only for the application runtime, not infrastructure. :disappointed:

RELEASE_NOTES.md
spolti: do you meant, delete the rest and start a new one?

ricardozanini: Yes! The "rest" is from the older release. We could have a log of everything, but since this file is used by the CI to create the release notes, I believe we would just make harder for them.

@radtriste once we implement this feature in the CI, the PR could also clean up this file on master.

radtriste: tricky part as some updates could be done on master while release is done ... or we should delete the content on master (as a PR) when creating the release branch. wdyt ?

ricardozanini: > or we should delete the content on master (as a PR) when creating the release branch.

+1 to that! :D

radtriste: just thinking it would make the process more complicated ... Gonna think about another possibility ...
added to https://issues.redhat.com/browse/KOGITO-2861 anyway ;)

sutaakar:
I like the separate image approach.
The image size will be smaller. The scripts in the image are simpler. Less configuration in the operator.
Of course the downside is that we need to take care of one (or several) more images.

spolti:
> I like the separate image approach.
> The image size will be smaller. The scripts in the image are simpler. Less configuration in the operator.
> Of course the downside is that we need to take care of one (or several) more images.

That's my point, imagine if in the future we have somehow, 20 kind of persistence types, which would result in 20 more images.
if we do on operator the logic will be the same for all.

and the downside of this would be the image size on future.

Anways, let's vote and decide it ASAP.

ricardozanini:
@radtriste @vaibhavjainwiz @Kaitou786 ^ please chime in

Kaitou786:
IMO having a single image seems better to me. No need to maintain n number of images (for each type of persistence we might support in the future).

radtriste:
The different images won't be so much different in fact. just our testing might be longer.
I would go for multiple images as it will be easier to maintain in the future. and it would ease the complexity on operator's side.

vaibhavjainwiz:
I also think maintaining multiple image would be better approach. Every time new persistence come, developer don't need to change existing working code. He could easily create new image by assembly existing CEkit modules. With this, we also sure that new functionality won't impact existing functionality and tester also don't need to test all functionality of like `data-index`. If changes are done in `data-index-mongo` image then testers only need to test mongo related functionality.

spolti:
closing, replaced by https://github.com/kiegroup/kogito-images/pull/265

PR #611, opened 2020-10-14
[KOGITO-3586] BDD tests should have unique namespace

radtriste:
https://issues.redhat.com/browse/KOGITO-3586

### DIFF
--- a/test/steps/data.go
+++ b/test/steps/data.go
@@ -26,6 +26,10 @@ import (
 	imgv1 "github.com/openshift/api/image/v1"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


sutaakar:
@radtriste It may be simpler to use map, that way you can avoid looping through the slice to check whether the namespace exists. Also, the race condition could happen there.
What about using `sync.Map` as is done in logging?

radtriste:
@sutaakar updated

sutaakar (approved):
Looks good to me, thanks

PR #602, opened 2020-10-06
[KOGITO-2562] SW: Auto-creation of Knative Eventing resources 

ricardozanini:
Many thanks for submiting your Pull Request :heart:! 

See: https://issues.redhat.com/browse/KOGITO-2562
Depends: https://github.com/kiegroup/kogito-runtimes/pull/806

What we did:

1. Small refactoring on `KogitoInfra` controllers to have less packages and have the Conditions structure similar to the other Kubernetes objects. `KogitoServices` are using this new structure to be aware of the right time to deploy the application. For example: if a given Kafka resource is not ready, the application won't be deployed. 

2. `KogitoInfra` has a new view from cli:
```
NAME                   RESOURCE NAME   KIND     API VERSION               STATUS   REASON
kogito-knative-infra   default         Broker   eventing.knative.dev/v1   True     
```
Users can now quickly debug any possible problem around infra. I will replicate this behavior as part of [KOGITO-3556](https://issues.redhat.com/browse/KOGITO-3556) for the other objects (Build and Services).

3. The `messaging/topics` endpoint from the Kogito service is now fetched. If there's a `KogitoInfra` bound to a Kafka or Knative resource, it will create the required objects (`KafkaTopics` or `SinkBinding`/`Triggers).

For testing purposes we can use:
- Knative: https://github.com/kiegroup/kogito-examples/pull/401
- Kafka: any * kafka * example. The topics should be created on demand if a Kafka instance is bound to a `KogitoInfra`

The topics creation should change soon as part of [KOGITO-3610](https://issues.redhat.com/browse/KOGITO-3610) cc @evacchi . The operator should be aware of any topics been created, but from KOGITO-3610 we will likely to have only one external topic to consume/publish. The messages should be filtered by CE type or source.

Having a `KogitoInfra` bound to your service:
```
Spec:
  Resource:
    API Version:  eventing.knative.dev/v1
    Kind:         Broker
    Name:         default
Status:
  Condition:
    Last Transition Time:  2020-10-15T18:04:41Z
    Status:                True
    Type:                  Success
```

You will have triggers been created:
```
Spec:
  Broker:  default
  Filter:
    Attributes:
      Type:  travellers
  Subscriber:
    Ref:
      API Version:  v1
      Kind:         Service
      Name:         process-knative-quickstart-quarkus
      Namespace:    kogito-2562
```

SinkBinding:
```
Spec:
  Sink:
    Ref:
      API Version:  eventing.knative.dev/v1
      Kind:         Broker
[...]

PR #617, opened 2020-10-19
KOGITO-3622 - Add grafana to kogito-infra

r00ta:
Jira: https://issues.redhat.com/browse/KOGITO-3622

Scenario: Trusty Monitoring
As you can see in the following schema 
overall (1)1
there are two external resources: prometheus and grafana. The prometheus instance is already managed by the operator, but the grafana resources not. 

The workflow is the following: 
1) A kogito runtime application with the monitoring addon is compiled and a docker image is built. That application will expose the name of the dashboards under the path `/monitoring/dashboards/list.json`. `list.json` contains the name of the dashboards that can be downloaded at `/monitoring/dashboards/xxx.json`. 
2) In case the user uses the grafana `kogito-infra` resource, he has to create it. Otherwise he has to install and deploy a grafana operator resource. 
3) When the kogito runtime application is deployed with the operator (a special flag `enableGrafanaDashboards` must be set) ,then the operator will fetch the dashboards and will create grafana dashboards out of them -> the related jira ticket is the following:   https://issues.redhat.com/browse/KOGITO-3624

Please, note that the final feature is composed by the following PR/ticket and the ticket https://issues.redhat.com/browse/KOGITO-3624 that still has to be implemented by me. 


pkg/controller/kogitoinfra/grafana/resource.go
```diff
@@ -0,0 +1,63 @@
...
+	} else if !exits {
+		log.Debug("Kogito grafana instance is not exists")
+		return nil, nil
+	} else {
+		log.Debug("Kogito grafana instance found")
+		return grafanaInstance, nil
+	}
+}
+
+func createNewGrafanaInstance(cli *client.Client, name string, namespace string, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) (*grafana.Grafana, error) {
```
spolti, line 51: will we going continue to provision infra for user?

pkg/controller/kogitoinfra/kogitoinfra_service.go
```diff
@@ -55,3 +59,8 @@ func IsKafkaResource(instance *v1alpha1.KogitoInfra) bool {
 func isKeycloakResource(instance *v1alpha1.KogitoInfra) bool {
 	return instance.Spec.Resource.APIVersion == infrastructure.KeycloakAPIVersion && instance.Spec.Resource.Kind == infrastructure.KeycloakKind
 }
+
+// isGrafanaResource check if provided KogitoInfra instance is for Keycloak resource
```
spolti, line 63: grafana?

spolti (changes requested):
just a little thing.

@ricardozanini haven't we decided to no create infra anymore or will continue doing it and later let this only for dev mode?

ricardozanini (changes requested):
Hi @r00ta! 

Thanks for your PR, but this is not the strategy we are aiming for 1.0

`KogitoInfra` is meant to **bind** the Kogito services with third party infra, not creating the infra itself. We just kept Kafka/Infinispan/Keycloak for now for backward compatibility.

For Kafka: We create KafkaTopics on already created Kafka instance.
For Infinispan: We fetch Infinispan server details and inject it into the service, so it could connect to it
For Knative Eventing: We bind a given broker with a Kogito service via SinkBindings and Triggers.
For Prometheus: We fetch for the `metrics` endpoint and create `ServiceMonitors` to bind then together, no `KogitoInfra` needed.

For Graphana: Users should create their own Graphana instance and the operator would fetch for the `/dashboards` endpoint, if it's in there, we create [GraphanaDashboards](https://github.com/integr8ly/grafana-operator/blob/master/documentation/dashboards.md) to be discovered by the Graphana Operator. We **won't create** Graphana instances whatsoever.

I'd say that we won't merge this PR since it's not the strategy we are aiming.

r00ta:
Hi @ricardozanini , thanks for the clarification. np, I'll implement what you just suggested. I'll close this

ricardozanini:
> Hi @ricardozanini , thanks for the clarification. np, I'll implement what you just suggested. I'll close this

Take a look at #602 it's pretty similar: we fetch for topics and create Knative resources.

PR #623, opened 2020-10-21
KOGITO-3624 - Fetch and deploy grafana dashboards

r00ta:
Jira: https://issues.redhat.com/browse/KOGITO-3624

This PR aims to fetch the static dashboards exposed by a kogito runtime application **with the monitoring addon** under the endpoint `/monitoring/dashboards/` and create a `GrafanaDashboard` resource for each of them (see https://operatorhub.io/operator/grafana-operator). If the monitoring addon is not enabled (i.e. the endpoint `/monitoring/dashboards/list.json` returns `404`, there are no effects).
The file `/monitoring/dashboards/list.json` is a list of strings, containing the dashboard names (let's assume `["dashboard1.json", "dashboard2.json"]).
Then the raw json dashboards can be downloaded at `monitoring/dashboards/dashboard1.json` and `monitoring/dashboards/dashboard2.json`.  

The grafana operator, the `Grafana` and `GrafanaDataSource` resources **must** be managed by the user at all, as we pointed out in this discussion https://github.com/kiegroup/kogito-cloud-operator/pull/617#pullrequestreview-511873377 . The only responsability of the operator will be to create a `GrafanaDashboard` for each dashboard, that's it. 


pkg/infrastructure/services/deployer.go
```diff
@@ -222,10 +227,45 @@ func (s *serviceDeployer) Deploy() (reconcileAfter time.Duration, err error) {
...
+	if err != nil || reconcileAfter > 0 {
+		return
+	}
+
+	dashboards, err := FetchGrafanaDashboards(s.client, s.instance)
+	if err != nil {
+		return reconciliationPeriodAfterDashboardsError, err
+	}
+
+	reconcileAfter, err = deployGrafanaDashboards(dashboards, s.client, s.getNamespace())
```
ricardozanini, line 239: Please move this to the `s.configureMonitoring()` function, there refactor to call a `Prometheus` and `Grafana` resources.

pkg/infrastructure/services/deployer.go
```diff
@@ -222,10 +227,45 @@ func (s *serviceDeployer) Deploy() (reconcileAfter time.Duration, err error) {
...
 
 	return
 }
 
+func deployGrafanaDashboards(dashboards []GrafanaDashboard, cli *client.Client, namespace string) (time.Duration, error) {
```
ricardozanini, line 244: Can you please move this to the same `configureMonitoring` function file?

pkg/infrastructure/services/messaging_test.go
```diff
@@ -80,12 +100,16 @@ func createServiceInstance(t *testing.T) v1alpha1.KogitoService {
 	}
 }
 
-func mockKogitoSvcReplies(t *testing.T, jsonResponse string) *httptest.Server {
-	handler := http.NewServeMux()
-	handler.HandleFunc(topicInfoPath, func(writer http.ResponseWriter, request *http.Request) {
-		_, err := writer.Write([]byte(jsonResponse))
-		assert.NoError(t, err)
-	})
+func mockKogitoSvcReplies(t *testing.T, handlers []serverHandler) *httptest.Server {
```
ricardozanini, line 103: I think it's time to move this to another go file, maybe `mock_svc_test.go` since it's serving different use cases.

pkg/infrastructure/services/prometheus.go
```diff
@@ -25,7 +27,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"net/http"
```
ricardozanini, line 28: I believe we are missing a `go fmt` alignment here. I'll take a look later. You don't need to worry about it now.

ricardozanini (changes requested):
Nice work @r00ta!! Looks really neat, I just left some minor comments to move some stuff around. Looks good.

Can you handle the docs as well and update the `RELEASE_NOTES.md` file? Many thanks!

pkg/infrastructure/services/deployer.go
r00ta: Hi @ricardozanini , what about 
```

func (s *serviceDeployer) configureMonitoring() (time.Duration, error) {
	if failedVerifyAddon, err := configurePrometheus(s.client, s.instance, s.scheme); err != nil {
		return 0, err
	} else if failedVerifyAddon {
		return reconciliationPeriodMonitoringEndpointNotAvailable, nil
	}

	reconcileAfter, err := configureGrafanaDashboards(s.client, s.instance, s.getNamespace())
	if err != nil {
		return reconciliationPeriodAfterDashboardsError, err
	}

	return reconcileAfter, err
}
```
and then create the function `configureGrafanaDashboards` in the `dashboards.go` file, moving there also the existing function `deployGrafanaDashboards`?

r00ta: I pushed a proposal, let me know what do you think. I can revert it just in case :)

pkg/infrastructure/services/messaging_test.go
r00ta: Done

pkg/infrastructure/services/messaging_test.go
```diff
@@ -44,7 +50,14 @@ func Test_fetchRequiredTopicsWithEmptyReply(t *testing.T) {
...
 
-	server := mockKogitoSvcReplies(t, emptyResponse)
+	handlers := []serverHandler{
+		{
+			Path:         topicInfoPath,
+			JSONResponse: emptyResponse,
+		},
+	}
+
+	server := mockKogitoSvcReplies(t, handlers)
```
ricardozanini, line 60: Don't need a temp with a declared slice :)

```suggestion
	server := mockKogitoSvcReplies(t, serverHandler{{Path: topicInfoPath, JSONResponse: emptyResponse}})
```

pkg/infrastructure/services/messaging_test.go
```diff
@@ -15,23 +15,29 @@
...
 
-	server := mockKogitoSvcReplies(t, responseWithTopics)
+	handlers := []serverHandler{
+		{
+			Path:         topicInfoPath,
+			JSONResponse: responseWithTopics,
+		},
+	}
+
+	server := mockKogitoSvcReplies(t, handlers)
```
ricardozanini, line 40: See my suggestion bellow

pkg/infrastructure/services/mock_svc_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+type serverHandler struct {
+	Path         string
+	JSONResponse string
+}
+
+func mockKogitoSvcReplies(t *testing.T, handlers []serverHandler) *httptest.Server {
```
ricardozanini, line 30:
```suggestion
func mockKogitoSvcReplies(t *testing.T, handlers ...serverHandler) *httptest.Server {
```

ricardozanini (changes requested):
@r00ta we can move forward! I just left one more suggestion that I missed somewhat in the previous review, sorry. :(

r00ta:
Hi @ricardozanini , I've completed the refactoring. Let me know if this looks good to you

ricardozanini (approved):

r00ta:
Hi @vaibhavjainwiz , thanks for the tips! I've refactored the files, let me know if this looks good to you

pkg/infrastructure/services/dashboards.go
```diff
@@ -0,0 +1,150 @@
...
+const (
+	// dashboardPath which the dashboards are fetched
+	dashboardsPath = "/monitoring/dashboards/"
+
+	// GrafanaDashboardAppName label app name to be used when a GrafanaDashboard is created
+	GrafanaDashboardAppName = "grafana"
+)
+
+// FetchGrafanaDashboards fetches the grafana dashboards from the KogitoService
+func FetchGrafanaDashboards(cli *client.Client, instance v1alpha1.KogitoService) ([]GrafanaDashboard, error) {
```
ricardozanini, line 50: Quick question: Why the `Fetch*` services are exported? Can't we keep then private?

```suggestion
func fetchGrafanaDashboards(cli *client.Client, instance v1alpha1.KogitoService) ([]GrafanaDashboard, error) {
```

r00ta: Hi @ricardozanini , sure. I've changed accordingly

pkg/infrastructure/services/dashboards.go
```diff
@@ -0,0 +1,147 @@
...
+
+func deployGrafanaDashboards(dashboards []GrafanaDashboard, cli *client.Client, kogitoService v1alpha1.KogitoService, scheme *runtime.Scheme, namespace string) (time.Duration, error) {
+	for _, dashboard := range dashboards {
+		resourceName := strings.ReplaceAll(strings.ToLower(dashboard.Name), ".json", "")
+		dashboardDefinition := &grafanav1.GrafanaDashboard{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      resourceName,
+				Namespace: namespace,
+				Labels: map[string]string{
+					framework.LabelAppKey: GrafanaDashboardAppName,
```
vaibhavjainwiz, line 132: Could we provide actual kogitoService name in place of hardcoded `grafana` keyword?

vaibhavjainwiz (approved):
Apart from label LGTM

pkg/infrastructure/services/dashboards.go
r00ta: Hi @vaibhavjainwiz , if I use the kogito service name (or any other string), I get the following error in the grafana operator `":"dashboard trusty-demo/operational-dashboard-loaneligibility found but selectors do not match"}` . 
Only with `grafana` it works. Any thoughts?

ricardozanini: I believe it's a requirement for the Grafana operator. We might need to take a look at their docs. Plus, our label in here won't hurt like `kogito.kie.org/service: serviceName`.

ricardozanini:
@spolti  @sutaakar can you guys please take a look?

pkg/infrastructure/kogito_service.go
```diff
@@ -107,7 +107,7 @@ func getServiceEndpoints(client *client.Client, namespace string, serviceHTTPRou
 // CreateKogitoServiceURI provides kogito service URI for given instance name
 func CreateKogitoServiceURI(service v1alpha1.KogitoService) string {
```
spolti, line 108: not related, but, shouldn't this function be changed?
it returns the service url, not uri and does not create it. wdyt?

pkg/infrastructure/services/dashboards.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func fetchGrafanaDashboardNamesForURL(serverURL string) ([]string, error) {
+	dashboardsURL := fmt.Sprintf("%s%s%s", serverURL, dashboardsPath, "list.json")
+	resp, err := http.Get(dashboardsURL)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
```
spolti, line 75: print some debug message?

sutaakar:
Checked manually, Grafana is properly connected to Prometheus, showing expected results.
Going to review the code...

pkg/infrastructure/services/dashboards.go
```diff
@@ -0,0 +1,147 @@
...
+	var dashboards []GrafanaDashboard
+	for _, name := range dashboardNames {
+		dashboardURL := fmt.Sprintf("%s%s%s", serverURL, dashboardsPath, name)
+		resp, err := http.Get(dashboardURL)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode == http.StatusNotFound {
+			return nil, nil
```
spolti, line 99: debug?

spolti (approved):
just a few things, not blockers though.
nice work.

pkg/infrastructure/services/dashboards.go
```diff
@@ -0,0 +1,147 @@
...
+					framework.LabelAppKey: GrafanaDashboardAppName,
+				},
+			},
+			Spec: grafanav1.GrafanaDashboardSpec{
+				Json: dashboard.RawJSONDashboard,
+				Name: dashboard.Name,
+			},
+		}
+		if _, err := kubernetes.ResourceC(cli).CreateIfNotExistsForOwner(dashboardDefinition, kogitoService, scheme); err != nil {
+			log.Warnf("Error occurs while creating dashboard %s, not going to reconcile the resource.", dashboard.Name, err)
```
sutaakar, line 141:
```suggestion
			log.Warnf("Error occurs while creating dashboard %s, not going to reconcile the resource: %v", dashboard.Name, err)
```

pkg/infrastructure/services/dashboards.go
```diff
@@ -0,0 +1,147 @@
...
+				Labels: map[string]string{
+					framework.LabelAppKey: GrafanaDashboardAppName,
+				},
+			},
+			Spec: grafanav1.GrafanaDashboardSpec{
+				Json: dashboard.RawJSONDashboard,
+				Name: dashboard.Name,
+			},
+		}
+		if _, err := kubernetes.ResourceC(cli).CreateIfNotExistsForOwner(dashboardDefinition, kogitoService, scheme); err != nil {
```
sutaakar, line 140: Minor - this call returns boolean parameter indicating whether dashboard was created. It would be nice to show `Successfully created grafana dashboard` below just in case object was created (to keep operator log clean)

pkg/infrastructure/services/deployer.go
```diff
@@ -344,5 +346,11 @@ func (s *serviceDeployer) configureMonitoring() (time.Duration, error) {
 	} else if failedVerifyAddon {
 		return reconciliationPeriodMonitoringEndpointNotAvailable, nil
 	}
-	return 0, nil
+
+	reconcileAfter, err := configureGrafanaDashboards(s.client, s.instance, s.scheme, s.getNamespace())
+	if err != nil {
+		return reconciliationPeriodAfterDashboardsError, err
```
sutaakar, line 352: In case reconciliation returns error then reconciliation period is ignored and default error behaviour is used (reconciling every second for first few reconciliation loops, then exponentially raising interval).

pkg/infrastructure/services/dashboards.go
sutaakar: I suppose it is because the default Grafana CR from OCP contains:
```
  dashboardLabelSelector:
    - matchExpressions:
        - key: app
          operator: In
          values:
            - grafana
```

sutaakar: IMHO we can put there the usual application name, possibly adding a note in docs for user to set proper selector in Grafana CR

pkg/infrastructure/services/dashboards_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+func Test_fetchDashboardNames(t *testing.T) {
+	dashboardNames := `["dashboard1.json", "dashboard2.json"]`
+
+	server := mockKogitoSvcReplies(t, serverHandler{Path: dashboardsPath + "list.json", JSONResponse: dashboardNames})
+	defer server.Close()
+
+	dashboards, err := fetchGrafanaDashboardNamesForURL(server.URL)
+	assert.NoError(t, err)
+	assert.NotEmpty(t, dashboards)
```
sutaakar, line 38: Nitpick, can you please check that proper dashboard names are returned?

sutaakar:
BDD test coverage will be implemented in https://issues.redhat.com/browse/KOGITO-3720

pkg/infrastructure/services/deployer.go
r00ta: Hi @sutaakar , thanks for the catch! I've modified the function so to return 
```
reconciliationPeriodAfterDashboardsError, nil
```
in case of error.

pkg/infrastructure/kogito_service.go
r00ta: Yep, I agree. What about renaming this to `GetKogitoServiceURL`?

r00ta:
Hi @spolti @sutaakar , thank you for the suggestions! I've resolved the comments, let me know if the solutions look good to you!

sutaakar:
@r00ta Please take a look on the comment thread related to the grafana label above. The grafana label selector is configurable in grafana CR.

pkg/infrastructure/services/deployer.go
ricardozanini: But please, do not ignore the error. Log it as a WARN at least.

pkg/infrastructure/kogito_service.go
ricardozanini: I agree with the new naming.

pkg/infrastructure/services/dashboards.go
ricardozanini: Thanks for the info, @sutaakar 
That's the proper way of handling this, imo.

r00ta: Hi @sutaakar @ricardozanini , I've changed the label name to `kogitoService.GetName()`, thanks for the tip!

pkg/infrastructure/kogito_service.go
r00ta: Done

pkg/infrastructure/services/deployer.go
r00ta: Done :+1:

r00ta:
Hi @sutaakar , thank you very much for the tip on the grafana label! Let me know if this looks good now

r00ta:
I don't get why the pipelines are failing on `CreateKogitoServiceURI` not defined.. it has been replaced and `CreateKogitoServiceURI` is not there at all

ricardozanini:
@r00ta please always run `make vet && make test` locally and check for errors before pushing.

r00ta:
Hi @ricardozanini , sorry there must be something wrong here. Locally `make test` looks good but here in the pipeline is failing. Do you have any tip? Maybe some packages are cached?

Kaitou786:
Hi @r00ta, the function you renamed from `CreateKogitoServiceURI` to `GetKogitoServiceURL` is not changed [here](https://github.com/r00ta/kogito-cloud-operator/blob/KOGITO-3624-grafana-dashboards/pkg/infrastructure/kogito_service_test.go#L39), after that linting and unit test should suceed :)

r00ta:
> Hi @r00ta, the function you renamed from `CreateKogitoServiceURI` to `GetKogitoServiceURL` is not changed [here](https://github.com/r00ta/kogito-cloud-operator/blob/KOGITO-3624-grafana-dashboards/pkg/infrastructure/kogito_service_test.go#L39), after that linting and unit test should suceed :)

Thank you for checking this, my local repo was completely screwed up..

ricardozanini (approved):

ricardozanini:
@sutaakar do you want to take a final look before the merge?

sutaakar:
Sure, give me few minutes....

sutaakar (approved):
Looks good to me

PR #627, opened 2020-10-26
[KOGITO-3689][KOGITO-2759] MongoDB support

radtriste:
https://issues.redhat.com/browse/KOGITO-3689
https://issues.redhat.com/browse/KOGITO-2759

Depends on https://github.com/kiegroup/kogito-examples/pull/432

### DIFF
--- a/examples/data-index-mongodb.yaml
+++ b/examples/data-index-mongodb.yaml
@@ -8,24 +8,37 @@ spec:
     apiVersion: kafka.strimzi.io/v1beta1
     kind: Kafka
 ---
+# Mongodb kubernetes operator and MongoDB instance should be pre-installed in namespace
+# See https://github.com/mongodb/mongodb-kubernetes-operator
+# See also at the end of the file for a MongoDB instance definition
+apiVersion: app.kiegroup.org/v1beta1
+kind: KogitoInfra
+metadata:
+  name: kogito-mongodb
+spec:
+  resource:
+    apiVersion: mongodb.com/v1
+    kind: MongoDB
+    name: external-mongodb # to change if you don't use the example MongoDB below
+  infraProperties:
+    username: kogitouser
+    database: kogito_dataindex
+    # password and connection string will be guessed from MongoDB instance
+---
 # requires a existing mongodb instance running on the target namespace
 apiVersion: app.kiegroup.org/v1beta1
-kind: KogitoDataIndex
+kind: KogitoSupportingService
 metadata:
   name: data-index
 spec:
-  image: quay.io/kiegroup/kogito-data-index-mongodb:latest
-  config:
-    quarkus.mongodb.connection-string: 'mongodb://mongodb-service7:27017'
-    quarkus.mongodb.credentials.password: kogito_user
-    quarkus.mongodb.credentials.username: kogito_pwd
-    quarkus.mongodb.database: kogito_dataindex
+  serviceType: DataIndex
   # environment variables to set in the runtime container. Example: JAVA_OPTIONS: "-Dquarkus.log.level=DEBUG"
   #env:
     # - name: JAVA_OPTIONS
     #    value: "-Dquarkus.log.level=DEBUG"
   # number of pods to be deployed
   replicas: 1
+  image: quay.io/kiegroup/kogito-data-index-mongodb:latest
   # Limits and requests for the Data Index pod
   #memoryLimit: ""
   #memoryRequest: ""
@@ -34,4 +47,54 @@ spec:
   # details about the kogito infra
   infra:
     - kogito-kafka-infra
+    - kogito-mongodb
 
+####### Setup simple MongoDB
+# This does require https://github.com/mongodb/mongodb-kubernetes-operator to be installed in the namespace
+# Uncomment below to create a MongoDB instance
+# ---
+# apiVersion: v1
+# kind: Secret
+# metadata:
+#   name: external-mongodb-secret
+# type: Opaque
+# data:
+#   password: bXlwYXNz #mypass
+# ---
+# apiVersion: mongodb.com/v1
+# kind: MongoDB
+# metadata:
+#   name: external-mongodb
+# spec:
+#   members: 1
+#   security:
+#     authentication:
+#       modes:
+#       - SCRAM
+#   statefulSet: # statefulSet for openshift only
+#     spec:
+#       template:
+#         spec:
+#           containers:
+#           - env:
+#             - name: MANAGED_SECURITY_CONTEXT
+#               value: "true"
+#             name: mongodb-agent
+#           - env:
+#             - name: MANAGED_SECURITY_CONTEXT
+#               value: "true"
+#             name: mongod
+#   type: ReplicaSet
+#   users:
+#   - db: admin
+#     name: kogitouser
+#     passwordSecretRef:
+#       name: external-mongodb-secret
+#     roles:
+#     - db: kogito_dataindex
+#       name: dbOwner
+#     - db: kogito_dataindex
+#       name: clusterAdmin
+#     - db: kogito_dataindex
+#       name: userAdminAnyDatabase
+#   version: 4.4.1
--- a/pkg/apis/app/v1beta1/kogitoinfra_types.go
+++ b/pkg/apis/app/v1beta1/kogitoinfra_types.go
@@ -51,6 +51,13 @@ type KogitoInfraSpec struct {
 	// Resource for the service. Example: Infinispan/Kafka/Keycloak.
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	Resource Resource `json:"resource,omitempty"`
+
+	// +optional
+	// +mapType=atomic
+	// Optional properties which would be needed to setup correct runtime/service configuration, based on the resource type.
+	// For example, MongoDB will require `username` and `database` as properties for a correct setup, else it will fail
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	InfraProperties map[string]string `json:"infraProperties,omitempty"`
 }
 
 // KogitoInfraStatus defines the observed state of KogitoInfra.
@@ -91,6 +98,10 @@ const (
 	UnsupportedAPIKind KogitoInfraConditionReason = "UnsupportedAPIKind"
 	// ResourceNotReady related resource is not ready
 	ResourceNotReady KogitoInfraConditionReason = "ResourceNotReady"
+	// ResourceConfigError related resource is not configured properly
+	ResourceConfigError KogitoInfraConditionReason = "ResourceConfigError"
+	// ResourceMissingResourceConfig related resource is missing a config information to continue
+	ResourceMissingResourceConfig KogitoInfraConditionReason = "ResourceMissingConfig"
 )
 
 // KogitoInfraCondition ...
--- a/pkg/controller/kogitoinfra/infinispan.go
+++ b/pkg/controller/kogitoinfra/infinispan.go
@@ -17,8 +17,9 @@ package kogitoinfra
 import (
 	"context"
 	"fmt"
-	"software.sslmate.com/src/go-pkcs12"
 	"sort"
+
+	"software.sslmate.com/src/go-pkcs12"
 
 	infinispan "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
 	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1beta1"
@@ -55,16 +56,16 @@ const (
 	envVarInfinispanUser
 	// envVarInfinispanPassword environment variable for setting infinispan password
 	envVarInfinispanPassword
-	enablePersistenceEnvKey = "ENABLE_PERSISTENCE"
 	// saslPlain is the PLAIN type.
-	saslPlain                  = "PLAIN"
-	pkcs12CertType             = "PKCS12"
-	certMountPath              = infrastructure.KogitoHomeDir + "/certs/infinispan"
-	truststoreSecretKey        = "truststore.p12"
-	truststoreMountPath        = certMountPath + "/" + truststoreSecretKey
-	infinispanTLSSecretKey     = "tls.crt"
-	infinispanCertMountName    = "infinispan-cert"
-	infinispanEnvKeyCredSecret = "INFINISPAN_CREDENTIAL_SECRET"
+	saslPlain                         = "PLAIN"
+	pkcs12CertType                    = "PKCS12"
+	certMountPath                     = infrastructure.KogitoHomeDir + "/certs/infinispan"
+	truststoreSecretKey               = "truststore.p12"
+	truststoreMountPath               = certMountPath + "/" + truststoreSecretKey
+	infinispanTLSSecretKey            = "tls.crt"
+	infinispanCertMountName           = "infinispan-cert"
+	infinispanEnvKeyCredSecret        = "INFINISPAN_CREDENTIAL_SECRET"
+	infinispanEnablePersistenceEnvKey = "ENABLE_PERSISTENCE"
 )
 
 var (
@@ -115,7 +116,7 @@ func (i *infinispanInfraReconciler) getInfinispanSecretEnvVars(infinispanInstanc
 		}
 	}
 
-	envProps = append(envProps, framework.CreateEnvVar(enablePersistenceEnvKey, "true"))
+	envProps = append(envProps, framework.CreateEnvVar(infinispanEnablePersistenceEnvKey, "true"))
 	secretName := customInfinispanSecret.Name
 	envProps = append(envProps, framework.CreateEnvVar(infinispanEnvKeyCredSecret, secretName))
 	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesInfinispanSpring[envVarInfinispanUser], secretName, infrastructure.InfinispanSecretUsernameKey))
--- a/pkg/controller/kogitoinfra/mongodb.go
+++ b/pkg/controller/kogitoinfra/mongodb.go
@@ -1,0 +1,330 @@
+// Copyright 2020 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kogitoinfra
+
+import (
+	"fmt"
+	"net/url"
+
+	"github.com/kiegroup/kogito-cloud-operator/pkg/apis/app/v1beta1"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/client/kubernetes"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/framework"
+	"github.com/kiegroup/kogito-cloud-operator/pkg/infrastructure"
+	mongodb "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	mongoDBSecretName = "kogito-mongodb-credential"
+
+	// Using URI for Quarkus as this is what we get from MongoDB instance
+	// and host/port for Spring Boot because URI cannot be used with credentials (Spring Boot Starter restriction) ...
+	// https://github.com/spring-projects/spring-boot/blob/b7fdf8fe87da1c01ff6aca041170a02f11280a1a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java#L61-L64
+	appPropMongoDBURI  = iota // for Quarkus
+	appPropMongoDBHost        // for Spring boot
+	appPropMongoDBPort        // for Spring boot
+
+	envVarMongoDBAuthDatabase
+	envVarMongoDBUser
+	envVarMongoDBPassword
+	envVarMongoDBDatabase
+
+	mongoDBEnvKeyCredSecret        = "MONGODB_CREDENTIAL_SECRET"
+	mongoDBEnablePersistenceEnvKey = "ENABLE_PERSISTENCE"
+
+	infraPropertiesUserKey         = "username"
+	infraPropertiesDatabaseKey     = "database"
+	infraPropertiesAuthDatabaseKey = "auth-database"
+)
+
+var (
+	// MongoDB variables for the KogitoInfra deployed infrastructure.
+	//For Quarkus: https://quarkus.io/guides/mongoDB-client#quarkus-mongoDB-client_configuration
+	//For Spring: https://github.com/mongoDB/mongoDB-spring-boot/blob/master/mongoDB-spring-boot-starter-remote/src/test/resources/test-application.properties
+
+	// propertiesMongoDBQuarkus mongoDB properties for quarkus runtime
+	propertiesMongoDBQuarkus = map[int]string{
+		appPropMongoDBURI: "quarkus.mongodb.connection-string",
+
+		envVarMongoDBAuthDatabase: "QUARKUS_MONGODB_CREDENTIALS_AUTH_SOURCE",
+		envVarMongoDBUser:         "QUARKUS_MONGODB_CREDENTIALS_USERNAME",
+		envVarMongoDBPassword:     "QUARKUS_MONGODB_CREDENTIALS_PASSWORD",
+		envVarMongoDBDatabase:     "QUARKUS_MONGODB_DATABASE",
+	}
+	// propertiesMongoDBSpring mongoDB properties for spring boot runtime
+	propertiesMongoDBSpring = map[int]string{
+		appPropMongoDBHost: "spring.data.mongodb.host",
+		appPropMongoDBPort: "spring.data.mongodb.port",
+
+		envVarMongoDBAuthDatabase: "SPRING_DATA_MONGODB_AUTHENTICATION_DATABASE",
+		envVarMongoDBUser:         "SPRING_DATA_MONGODB_USERNAME",
+		envVarMongoDBPassword:     "SPRING_DATA_MONGODB_PASSWORD",
+		envVarMongoDBDatabase:     "SPRING_DATA_MONGODB_DATABASE",
+	}
+)
+
+func (i *mongoDBInfraReconciler) getMongoDBSecretEnvVars(mongoDBInstance *mongodb.MongoDB) ([]corev1.EnvVar, error) {
+	var envProps []corev1.EnvVar
+
+	customMongoDBSecret, resultErr := i.loadCustomKogitoMongoDBSecret(i.instance.Namespace)
+	if resultErr != nil {
+		return nil, resultErr
+	}
+
+	if customMongoDBSecret == nil {
+		customMongoDBSecret, resultErr = i.createCustomKogitoMongoDBSecret(i.instance.Namespace, mongoDBInstance)
+		if resultErr != nil {
+			return nil, resultErr
+		}
+	}
+
+	envProps = append(envProps, framework.CreateEnvVar(mongoDBEnablePersistenceEnvKey, "true"))
+	mongoDBSecretName := customMongoDBSecret.Name
+	envProps = append(envProps, framework.CreateEnvVar(mongoDBEnvKeyCredSecret, mongoDBSecretName))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBAuthDatabase], mongoDBSecretName, infrastructure.MongoDBAppSecretAuthDatabaseKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBAuthDatabase], mongoDBSecretName, infrastructure.MongoDBAppSecretAuthDatabaseKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBUser], mongoDBSecretName, infrastructure.MongoDBAppSecretUsernameKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBUser], mongoDBSecretName, infrastructure.MongoDBAppSecretUsernameKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBPassword], mongoDBSecretName, infrastructure.MongoDBAppSecretPasswordKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBPassword], mongoDBSecretName, infrastructure.MongoDBAppSecretPasswordKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBDatabase], mongoDBSecretName, infrastructure.MongoDBAppSecretDatabaseKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBDatabase], mongoDBSecretName, infrastructure.MongoDBAppSecretDatabaseKey))
+	return envProps, nil
+}
+
+func (i *mongoDBInfraReconciler) getMongoDBAppProps(mongoDBInstance *mongodb.MongoDB) (map[string]string, error) {
+	appProps := map[string]string{}
+
+	mongoDBURI := mongoDBInstance.Status.MongoURI
+	if len(mongoDBURI) > 0 {
+		mongoDBParsedURL, err := url.ParseRequestURI(mongoDBURI)
+		if err != nil {
+			return nil, err
+		}
+		appProps[propertiesMongoDBSpring[appPropMongoDBHost]] = mongoDBParsedURL.Hostname()
+		appProps[propertiesMongoDBSpring[appPropMongoDBPort]] = mongoDBParsedURL.Port()
+		appProps[propertiesMongoDBQuarkus[appPropMongoDBURI]] = mongoDBURI
+	}
+
+	return appProps, nil
+}
+
+func (i *mongoDBInfraReconciler) updateMongoDBAppPropsInStatus(mongoDBInstance *mongodb.MongoDB) error {
+	log.Debug("going to Update MongoDB app properties in kogito infra instance status")
+	appProps, err := i.getMongoDBAppProps(mongoDBInstance)
+	if err != nil {
+		return err
+	}
+	i.instance.Status.AppProps = appProps
+	log.Debugf("Following app properties are set infra status : %s", appProps)
+	return nil
+}
+
+func (i *mongoDBInfraReconciler) updateMongoDBEnvVarsInStatus(mongoDBInstance *mongodb.MongoDB) error {
+	log.Debug("going to Update MongoDB env properties in kogito infra instance status")
+	envVars, err := i.getMongoDBSecretEnvVars(mongoDBInstance)
+	if err != nil {
+		return err
+	}
+	i.instance.Status.Env = envVars
+	log.Debugf("Following env properties are set infra status : %s", envVars)
+	return nil
+}
+
+func (i *mongoDBInfraReconciler) loadDeployedMongoDBInstance(instanceName string, namespace string) (*mongodb.MongoDB, error) {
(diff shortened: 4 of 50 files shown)


radtriste:
- data-index ok
- runtimes quarkus non native ok
- runtimes spring boot is having some problems. deactivated for now
- runtimes quarkus native test in progress

radtriste:
Setting back to draft as I got some weird errors when scenario are launched in parallel

radtriste:
For MongoDB / KAfka examples, I created https://issues.redhat.com/browse/KOGITO-3756 as I think it should be done in another PR.
Investigating Spring Boot issues ...

radtriste:
Spring Boot issues will be solved with https://github.com/kiegroup/kogito-examples/pull/432 :)

radtriste:
once https://github.com/kiegroup/kogito-cloud-operator/pull/646 is rebased, will merge that one and update the controller

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,308 @@
...
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBUser], mongoDBSecretName, infrastructure.MongoDBAppSecretUsernameKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBUser], mongoDBSecretName, infrastructure.MongoDBAppSecretUsernameKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBPassword], mongoDBSecretName, infrastructure.MongoDBAppSecretPasswordKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBPassword], mongoDBSecretName, infrastructure.MongoDBAppSecretPasswordKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBSpring[envVarMongoDBDatabase], mongoDBSecretName, infrastructure.MongoDBAppSecretDatabaseKey))
+	envProps = append(envProps, framework.CreateSecretEnvVar(propertiesMongoDBQuarkus[envVarMongoDBDatabase], mongoDBSecretName, infrastructure.MongoDBAppSecretDatabaseKey))
```
ricardozanini, line 104: For future reference: I believe we need to segregate the properties in the structure like:

```go
type RuntimeProperties struct {
    Properties map[string]string
    Runtime    Runtime
}
```

This way the service would know which property to fetch and apply. Right now we have both properties in the configuration files..

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,308 @@
...
+	return envProps, nil
+}
+
+func getMongoDBAppProps(cli *client.Client, mongoDBInstance *mongodb.MongoDB) (map[string]string, error) {
+	appProps := map[string]string{}
+
+	mongoDBURI := mongoDBInstance.Status.MongoURI
```
ricardozanini, line 111: Why not using the attribute directly and drop this temp?

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,308 @@
...
+	} else if !exits {
+		log.Debug("Kogito mongoDB instance is not exists")
+		return nil, nil
+	} else {
+		log.Debug("Kogito mongoDB instance found")
+		return mongoDBInstance, nil
+	}
+}
+
+func createNewMongoDBInstance(cli *client.Client, name string, namespace string, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) (*mongodb.MongoDB, error) {
```
ricardozanini, line 162: I think we need to stay away from this approach. The others are just kept there for backward compatibility, should go soon. Users must create their references, we just bound to them.

ricardozanini: In Knative use case, for instance, we do not create the `Broker`. It's a resource we expect from the user since it has its subtleties that are tied to a particular scenario.

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,308 @@
...
+	} else if !exits {
+		log.Errorf("%s not found", mongoDBSecretName)
+		return nil, nil
+	} else {
+		log.Debugf("%s successfully fetched", mongoDBSecretName)
+		return secret, nil
+	}
+}
+
+func createCustomKogitoMongoDBSecret(cli *client.Client, namespace string, mongoDBInstance *mongodb.MongoDB, instance *v1alpha1.KogitoInfra, scheme *runtime.Scheme) (*corev1.Secret, error) {
```
ricardozanini, line 200: Could you please add a reference to the MongoDB operator documentation that mentions this use case?

pkg/infrastructure/mongodb.go
```diff
@@ -0,0 +1,279 @@
...
+	mongodb "github.com/mongodb/mongodb-kubernetes-operator/pkg/apis/mongodb/v1"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	membersSize    = 1
+	mongoDBVersion = "4.4.1"
```
ricardozanini, line 31: oh! this is a super risk. Wondering if we could read this property from somewhere else? Maybe from their release page? At least we could take the latest minor. :thinking:

pkg/infrastructure/mongodb.go
```diff
@@ -0,0 +1,279 @@
...
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			defaultPasswordSecretRef: password,
+		},
+	}
+}
+
+// GetMongoDBStub returns the preconfigured MongoDB stub with set namespace, name and secretName
+func GetMongoDBStub(openshift bool, namespace, name string, users []MongoDBUserCred) *mongodb.MongoDB {
```
ricardozanini, line 180: Dropping the requirement of creating the instance ourselves, I believe we will need to move this code to the BDD tests instead, since there I believe would make much more sense to have MongoDB installed automatically.

ricardozanini (commented):
This is super cool, Tristan! Would love to add this feature for 1.0. I just left some comments. I believe we can drop the MongoDB instance creation. It's not our responsibility. Infinispan, Kafka and Keycloak will follow soon. For 1.0 I'll add a note warning the deprecation.

pkg/controller/kogitoinfra/mongodb.go
radtriste: I guess this will be done after 1.0 ;)

pkg/controller/kogitoinfra/mongodb.go
radtriste: I just don't want to repeat it 3 times into the code

pkg/controller/kogitoinfra/mongodb.go
radtriste: https://issues.redhat.com/browse/KOGITO-3777

pkg/infrastructure/mongodb.go
radtriste: This is for the creation of the MongoDB instance, gonna move this to BDD

pkg/controller/kogitoinfra/mongodb.go
radtriste: removing

radtriste:
@ricardozanini updated. please have a look. I am retesting it in parallel.

I was also thinking, in KogitoInfra, shouldn't we have a "configuration" map (like a map[string]string) that would allow for example to select the db. This would be quite useful as there could be many databases and users to those db, and we need to retrieve correct user to it :S

We could have something like
```
kind: KogitoInfra
metadata:
  name: kogito-mongodb
spec:
  resource:
    apiVersion: mongodb.com/v1
    kind: MongoDB
  config:
    database: <dbName>
```

wdyt ?

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -187,14 +202,12 @@ func (i *mongoDBInfraReconciler) createCustomKogitoMongoDBSecret(namespace strin
...
-		secret.StringData = map[string]string{
+		StringData: map[string]string{
 			infrastructure.MongoDBAppSecretAuthDatabaseKey: credentials.AuthDatabase,
 			infrastructure.MongoDBAppSecretUsernameKey:     credentials.Username,
 			infrastructure.MongoDBAppSecretPasswordKey:     credentials.Password,
 			infrastructure.MongoDBAppSecretDatabaseKey:     credentials.Database,
-		}
+		},
 	}
 	if err := controllerutil.SetOwnerReference(i.instance, secret, i.scheme); err != nil {
```
ricardozanini, line 212: Can't we use `framework.SetOwner` instead?

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,290 @@
...
+	} else {
+		log.Debug("Kogito mongoDB instance found")
+		return mongoDBInstance, nil
+	}
+}
+
+func (i *mongoDBInfraReconciler) loadCustomKogitoMongoDBSecret(namespace string) (*corev1.Secret, error) {
+	log.Debugf("Fetching %s ", mongoDBSecretName)
+	secret := &corev1.Secret{}
+	if exits, err := kubernetes.ResourceC(i.client).FetchWithKey(types.NamespacedName{Name: mongoDBSecretName, Namespace: namespace}, secret); err != nil {
```
ricardozanini, line 167:
```suggestion
	if exists, err := kubernetes.ResourceC(i.client).FetchWithKey(types.NamespacedName{Name: mongoDBSecretName, Namespace: namespace}, secret); err != nil {
```

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,290 @@
...
+		log.Errorf("Error occurs while fetching %s", mongoDBSecretName)
+		return nil, err
+	} else if !exits {
+		log.Errorf("%s not found", mongoDBSecretName)
```
ricardozanini, line 171: We might need to return a specialized error here to reconcile after T? Take a look at the `errors.go` file.

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,290 @@
...
+			namespace = i.instance.Namespace
+			log.Debugf("Namespace is not provided for custom resource, taking instance namespace(%s) as default", namespace)
+		}
+		mongoDBInstance, resultErr = i.loadDeployedMongoDBInstance(i.instance.Spec.Resource.Name, namespace)
+		if resultErr != nil {
+			return false, resultErr
+		}
+		if mongoDBInstance == nil {
+			return false, errorForResourceNotFound("MongoDB", i.instance.Spec.Resource.Name, namespace)
+		}
```
ricardozanini, line 264: This code is essentially the same thing as the `else`  branch. We can move it to outside of the condition.

pkg/infrastructure/mongodb.go
```diff
@@ -0,0 +1,163 @@
...
+
+	secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: user.PasswordSecretRef.Name, Namespace: mongoDBInstance.Namespace}}
+	if exists, err := kubernetes.ResourceC(cli).Fetch(secret); err != nil {
+		return err
+	} else if exists {
+		passwordKey := DefaultPasswordSecretRef
+		if user.PasswordSecretRef.Key != "" {
+			passwordKey = user.PasswordSecretRef.Key
+		}
+		creds.Password = string(secret.Data[passwordKey])
```
ricardozanini, line 131: Is this safe? Where's this data will be saved? In the MongoDB CR itself?

ricardozanini (commented):
Sorry a couple more comments :disappointed:

pkg/controller/kogitoinfra/mongodb.go
radtriste: did not get that one ...

pkg/infrastructure/mongodb.go
radtriste: this should be stored into a secret for the runtime/data-index to have it via env var secret

pkg/controller/kogitoinfra/mongodb.go
radtriste: I think this needs more redo on all the infra controllers.
I have just did the think similar to https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/controller/kogitoinfra/infinispan.go#L288

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,289 @@
...
+		appProps[propertiesMongoDBSpring[appPropMongoDBHost]] = mongoDBParsedURL.Hostname()
+		appProps[propertiesMongoDBSpring[appPropMongoDBPort]] = mongoDBParsedURL.Port()
+		appProps[propertiesMongoDBQuarkus[appPropMongoDBURI]] = mongoDBURI
+	}
+
+	return appProps, nil
+}
+
+func (i *mongoDBInfraReconciler) updateMongoDBAppPropsInStatus(mongoDBInstance *mongodb.MongoDB) error {
+	log.Debugf("going to Update MongoDB app properties in kogito infra instance status")
```
sutaakar, line 127:
```suggestion
	log.Debug("going to Update MongoDB app properties in kogito infra instance status")
```

radtriste:
seems latest changes brought some problems with the test ... gonna investigate it

pkg/controller/kogitoinfra/mongodb.go
ricardozanini: Sure! But can't we start here and open a JIRA to follow up others? Optional, do if you have the time. Otherwise a follow up JIRA would do.

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,289 @@
...
+		if mongoDBAvailable, err := infrastructure.IsMongoDBOperatorAvailable(i.client, i.instance.Namespace); err != nil {
+			return false, err
+		} else if !mongoDBAvailable {
+			return false, errorForResourceAPINotFound(&i.instance.Spec.Resource)
+		}
+		mongoDBInstance, resultErr = i.loadDeployedMongoDBInstance(infrastructure.MongoDBInstanceName, i.instance.Namespace)
+		if resultErr != nil {
+			return false, resultErr
+		}
+	}
```
ricardozanini, line 277: @radtriste that's what I meant:

```suggestion
    mongoDBName := i.instance.Spec.Resource.Name
    mongoDBNamespace := i.instance.Spec.Resource.Namespace
	if len(mongoDBName) > 0 {
		log.Debugf("Custom mongodb instance reference is provided")

		if len(mongoDBNamespace) == 0 {
			mongoDBNamespace = i.instance.Namespace
			log.Debugf("Namespace is not provided for custom resource, taking instance namespace(%s) as default", namespace)
		}
	} else {
	        mongoDBName = infrastructure.MongoDBInstanceName
	        mongoDBNamespace = i.instance.Namespace
		// create/refer kogito-mongodb instance
		log.Debugf("Custom mongodb instance reference is not provided")
		// Verify MongoDB Operator (it's installation is required in the same namespace, that's why we do this check as well)
		if mongoDBAvailable, err := infrastructure.IsMongoDBOperatorAvailable(i.client, i.instance.Namespace); err != nil {
			return false, err
		} else if !mongoDBAvailable {
			return false, errorForResourceAPINotFound(&i.instance.Spec.Resource)
		}
	}
	
	if mongoDBInstance, resultErr := i.loadDeployedMongoDBInstance(mongoDBName, mongoDBNamespace); err != nil {
		return false, resultErr
	} else if mongoDBInstance == nil {
	       return false, errorForResourceNotFound("MongoDB", i.instance.Spec.Resource.Name, namespace)
	}
```

test/features/deploy_kogito_runtime.feature
```diff
@@ -300,6 +302,71 @@ Feature: Deploy Kogito Runtime
...
+  @externalcomponent
+  @mongodb
+  Scenario: Deploy <example-service> service with Maven profile <profile> using external MongoDB
+    Given Kogito Operator is deployed with MongoDB operator
+    And MongoDB instance "external-mongodb" is deployed with configuration:
+      | username | developer            |
+      | password | mypass               |
+      | database | kogito_dataindex     |
+    And Install MongoDB Kogito Infra "external-mongodb" within 5 minutes with configuration:
+      | resource | name     | external-infinispan  |
```
sutaakar, line 317:
```suggestion
      | resource | name     | external-mongodb  |
```

test/framework/mongodb.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+var (
+	mongoDBOperatorDeployFilesURI = "https://raw.githubusercontent.com/mongodb/mongodb-kubernetes-operator/" + mongoDBOperatorVersion + "/deploy/"
+
+	// DefaultMongoDBSecretName is the default name for the user's secret
+	DefaultMongoDBSecretName = infrastructure.DefaultMongoDBUsername + "-mongodb-secret"
+)
+
+// DeployMongoDBInstance deploys an instance of Infinispan
```
sutaakar, line 43:
```suggestion
// DeployMongoDBInstance deploys an instance of Mongo DB
```

test/framework/operator.go
```diff
@@ -298,3 +304,88 @@ func CreateNamespacedSubscriptionIfNotExist(namespace string, subscriptionName s
...
+
+				object.(*apiextensionsv1beta1.CustomResourceDefinition).Spec.Names.Plural = crdName
+				object.(*apiextensionsv1beta1.CustomResourceDefinition).ObjectMeta.Name = metadataName
+			}
+		}); err != nil {
+			return err
+		}
+	}
+
+	// TODO: error handling, go lint is screaming about this
```
sutaakar, line 332: I guess this can be deleted now.

test/framework/operator.go
```diff
@@ -298,3 +304,88 @@ func CreateNamespacedSubscriptionIfNotExist(namespace string, subscriptionName s
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// Set correct file to be deployed
+	if IsOpenshift() {
+		// Used to give correct access to pvc/secret
+		// https://github.com/mongodb/mongodb-kubernetes-operator/issues/212#issuecomment-704744307
+		output, err := CreateCommand("oc", "adm", "policy", "add-scc-to-user", "anyuid", "system:serviceaccount:"+namespace+":mongodb-kubernetes-operator").WithLoggerContext(namespace).Sync("add-scc-to-user").Execute()
```
sutaakar, line 362: Why this command needs to be synchronized?

pkg/controller/kogitoinfra/mongodb.go
radtriste: Coming back to that, why would we need to reconcile ?
this is something that the operator is creating for the application to have auth information...
we could when creating it though

radtriste: thinking of something. testing it and then get back ;)

radtriste: added https://github.com/kiegroup/kogito-cloud-operator/pull/627/files#diff-65b99f6143ba7d89e73a9c85839d59c54bcf857ef2a77919de76c68165d3e9dcR63-R69

test/framework/operator.go
radtriste: when executed in parallel and almost simultaneously, this creates conflict and only one command will succeed. This allows to wait for execution of the command and avoid failing command execution when parallel

radtriste:
@Kaitou786 Could you check the OLM test ? this does not seem to be related to the PR or ?

ricardozanini:
> @Kaitou786 Could you check the OLM test ? this does not seem to be related to the PR or ?

I think we can drop the infinispan, keycloak and kafka manifests, might be related.

```
time="2020-12-04T13:12:10Z" level=fatal msg="permissive mode disabled" error="error loading manifests from directory: error loading package into db: FOREIGN KEY constraint failed"
```

Kaitou786:
> > @Kaitou786 Could you check the OLM test ? this does not seem to be related to the PR or ?
> 
> I think we can drop the infinispan, keycloak and kafka manifests, might be related.
> 
> ```
> time="2020-12-04T13:12:10Z" level=fatal msg="permissive mode disabled" error="error loading manifests from directory: error loading package into db: FOREIGN KEY constraint failed"
> ```

I tested this with removing those other operators [here](https://github.com/kiegroup/kogito-cloud-operator/pull/689) but didn't worked, probably something changed on the OLM CI I'll look more into this on monday :)

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,366 @@
...
+// MongoDBCredential holds the credentials information of a user into an mongoDB server
+type MongoDBCredential struct {
+	AuthDatabase string `yaml:"auth-database"`
+	Username     string `yaml:"username"`
+	Password     string `yaml:"password"`
+	Database     string `yaml:"database"`
+}
+
+// fillMongoDBCredential fills the credential of the MongoDB server deployed with the Kogito Operator.
+// Prefilled values will be taken into account and corrected if needed.
```
sutaakar, line 300: Is it a good approach to correct provided values? Shouldn't we rather throw an error complaining about wrong values provided?

test/framework/operator.go
sutaakar: ok

sutaakar:
one question to clarify above, otherwise looks good.

pkg/controller/kogitoinfra/mongodb.go
radtriste: This is something I put in place when i was also creating an instance when none was provided.
Seems good not to have default values and raise an error indeed.
So we should consider that the MongoDB will always need Name, AuthDB and Database in the KogitoInfra Config part.
@ricardozanini wdyt ?

ricardozanini: We need to implement Admission controllers, actually. So we could validate the CR before it's created. 

I think raising an error is a good approach, but best if we can have a named error, so users would clearly see what's wrong via `status` attribute of the `KogitoInfra` CR. And also the controller won't reconcile anymore.

Adding defaults, in this case, **could** lead to hard to catch errors during runtime if the connection information was incorrectly set by the operator just because we defined default values.

ricardozanini (approved):

pkg/controller/kogitoinfra/mongodb.go
radtriste: oki. gonna update that

radtriste:
@sutaakar @ricardozanini removed default params and added some errors. Let me know.
I am testing on CRC for now (Openshift having some problems due to Openstack migration)

spolti (approved):

radtriste:
After removing default params I just came accross the problem that CLI was not implemented to add information to KogitoInfra (because of default params being the same as in BDD test before the change, thx @sutaakar @ricardozanini )

So a bit of change here:

- `KogitoInfra.Config` becomes `KogitoInfra.InfraProperties`
- CLI defines a generic `PropertiesFlag` with option `--property`. This flag could be reused easily for other components.
- Naming errors `errorForMissingResourceConfig` and `errorForResourceConfigError`. No reconcialiation for them.

pkg/apis/app/v1beta1/kogitoinfra_types.go
```diff
@@ -51,6 +51,13 @@ type KogitoInfraSpec struct {
 	// Resource for the service. Example: Infinispan/Kafka/Keycloak.
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	Resource Resource `json:"resource,omitempty"`
+
+	// +optional
+	// +mapType=atomic
+	// Optional properties which would be needed to setup correct runtime/service configuration, based on the resource type.
+	// For example, MongoDB will require `username` and `database` as properties for a correct setup, else it will fail
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	InfraProperties map[string]string `json:"config,omitempty"`
```
ricardozanini, line 60: Question: why we have `InfraProperties` in the attribute but `config` in the CRD? If you're going to rename the type, you could also rename the CRD interface as well.

radtriste: woops. good catch. gonna update this.

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,330 @@
...
+
+// MongoDBCredential holds the credentials information of a user into an mongoDB server
+type MongoDBCredential struct {
+	AuthDatabase string `yaml:"auth-database"`
+	Username     string `yaml:"username"`
+	Password     string `yaml:"password"`
+	Database     string `yaml:"database"`
+}
+
+// fillMongoDBCredential fills the credential of the MongoDB server deployed with the Kogito Operator.
```
ricardozanini, line 279:
```suggestion
// retrieveMongoDBCredentials 
```
We might need to rethink this description.

pkg/infrastructure/mongodb.go
```diff
@@ -0,0 +1,81 @@
...
+	// MongoDBAppSecretDatabaseKey is the secret database key set in the linked secret for an application
+	MongoDBAppSecretDatabaseKey = "database"
+	// MongoDBAppSecretUsernameKey is the secret username key set in the linked secret for an application
+	MongoDBAppSecretUsernameKey = "username"
+	// MongoDBAppSecretPasswordKey is the secret password key set in the linked secret for an application
+	MongoDBAppSecretPasswordKey = "password"
+)
+
+var (
+	// MongoDBAPIVersion refers to kafka APIVersion
```
ricardozanini, line 48:
```suggestion
	// MongoDBAPIVersion refers to MongoDB APIVersion
```

ricardozanini (commented):
I left some minor comments. Great job, man!

radtriste:
@ricardozanini thx, updated

test/steps/kogitoinfra.go
```diff
@@ -45,6 +45,8 @@ func (data *Data) installKogitoInfraWithinMinutesWithConfiguration(targetResourc
 		return err
 	}
 
+	framework.GetLogger(data.Namespace).Infof("Got kogitoInfra config %v", infraResource.Spec.InfraProperties)
```
sutaakar, line 48: Seems like debugging log, is it needed?

test/steps/mappers/create_kogito_infra.go
```diff
@@ -55,6 +56,7 @@ func mapKogitoInfraTableRow(row *TableRow, kogitoInfra *v1beta1.KogitoInfra) (ma
 
 	switch firstColumn {
 	case kogitoInfraConfigKey:
+		framework.GetLogger(kogitoInfra.Namespace).Infof("Got config %s", getSecondColumn(row))
```
sutaakar, line 59: Same here

test/steps/kogitoinfra.go
radtriste: it is...updated

radtriste:
tests ok on CRC ... current BDD test error is unrelated

/jenkins test

sutaakar:
The BDD test error is most likely related to https://github.com/kiegroup/kogito-runtimes/pull/924

sutaakar:
@radtriste When I create a KogitoInfra without configuration then I would expect some descriptive error in the operator log or in KogitoInfra status. However there is nothing relevant.
Used CR:
```
kind: KogitoInfra
metadata:
  name: external-mongodb
spec:
  resource:
    apiVersion: mongodb.com/v1
    kind: MongoDB
    name: external-mongodb
```
Operator log:
```
{"level":"info","T":"2020-12-08T18:37:53.759Z","logger":"kogitoinfra_controller","msg":"MongoDB instance is running"}
{"level":"error","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"kogito-mongodb-credential not found"}
{"level":"info","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"Updating Kogito Infra status"}
{"level":"info","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"Kogito Infra successfully reconciled"}
{"level":"info","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"Updating kogitoInfra value with new properties : external-mongodb"}
{"level":"info","T":"2020-12-08T18:37:53.777Z","logger":"kogitoinfra_controller","msg":"Successfully Update Kogito Infra status"}
```
Infra status shows success. 

Can you also please update example - https://github.com/kiegroup/kogito-cloud-operator/blob/master/examples/data-index-mongodb.yaml ?

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (i *mongoDBInfraReconciler) loadCustomKogitoMongoDBSecret(namespace string) (*corev1.Secret, error) {
+	log.Debugf("Fetching %s ", mongoDBSecretName)
+	secret := &corev1.Secret{}
+	if exits, err := kubernetes.ResourceC(i.client).FetchWithKey(types.NamespacedName{Name: mongoDBSecretName, Namespace: namespace}, secret); err != nil {
+		log.Errorf("Error occurs while fetching %s", mongoDBSecretName)
+		return nil, err
+	} else if !exits {
+		log.Errorf("%s not found", mongoDBSecretName)
```
sutaakar, line 173: Does this need to be logged as an error (or logged at all)?
Missing custom Kogito secret can mean that the operator just didn't create a new one yet. We don't have a broader context in this function so IMHO it should be ok to return nil and not log anything. Possible logging can be handled by other functions using this function.

sutaakar: Or possible reducing the log level to `Debugf`

sutaakar:
Please update also RELEASE_NOTES.md

radtriste:
> @radtriste When I create a KogitoInfra without configuration then I would expect some descriptive error in the operator log or in KogitoInfra status. However there is nothing relevant.
> Used CR:
> 
> ```
> apiVersion: app.kiegroup.org/v1beta1
> kind: KogitoInfra
> metadata:
>   name: external-mongodb
> spec:
>   resource:
>     apiVersion: mongodb.com/v1
>     kind: MongoDB
>     name: external-mongodb
> ```
> 
> Operator log:
> 
> ```
> {"level":"info","T":"2020-12-08T18:37:53.759Z","logger":"kogitoinfra_controller","msg":"MongoDB instance is running"}
> {"level":"error","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"kogito-mongodb-credential not found"}
> {"level":"info","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"Updating Kogito Infra status"}
> {"level":"info","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"Kogito Infra successfully reconciled"}
> {"level":"info","T":"2020-12-08T18:37:53.768Z","logger":"kogitoinfra_controller","msg":"Updating kogitoInfra value with new properties : external-mongodb"}
> {"level":"info","T":"2020-12-08T18:37:53.777Z","logger":"kogitoinfra_controller","msg":"Successfully Update Kogito Infra status"}
> ```
> 
> Infra status shows success.
> 
> Can you also please update example - https://github.com/kiegroup/kogito-cloud-operator/blob/master/examples/data-index-mongodb.yaml ?

@sutaakar I think the problem is coming from copying from infinispan controller which has the same problem I think:
https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/controller/kogitoinfra/infinispan.go#L390-L398
The error is just ignored ... Gonna correct that

radtriste:
Look a lot better :)
```
{"level":"error","ts":1607500967.2168424,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"kogitoinfra-controller","name":"external-mongodb","namespace":"tradisso-local-cucumber-3322","error":"Resource external-mongodb, configuration information (username) is missing","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/tradisso/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/tradisso/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.1/pkg/internal/controller/controller.go:235\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/tradisso/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.1/pkg/internal/controller/controller.go:209\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/home/tradisso/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.1/pkg/internal/controller/controller.go:188\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n\t/home/tradisso/go/pkg/mod/k8s.io/apimachinery@v0.18.3/pkg/util/wait/wait.go:155\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil\n\t/home/tradisso/go/pkg/mod/k8s.io/apimachinery@v0.18.3/pkg/util/wait/wait.go:156\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/tradisso/go/pkg/mod/k8s.io/apimachinery@v0.18.3/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/tradisso/go/pkg/mod/k8s.io/apimachinery@v0.18.3/pkg/util/wait/wait.go:90"}

[tradisso@tradisso ~]$ oc get kogitoinfra
NAME               RESOURCE NAME      KIND      API VERSION      STATUS   REASON
external-mongodb   external-mongodb   MongoDB   mongodb.com/v1   False    ResourceMissingConfig
```

sutaakar:
@radtriste There is a disadvantage with latest approach. It triggers error reconciliations in operator. Though I don't know about any approach which would allow setting Failure status for infra while not triggering error reconciliation (without significant refactoring).

ricardozanini:
> @radtriste There is a disadvantage with latest approach. It triggers error reconciliations in operator. Though I don't know about any approach which would allow setting Failure status for infra while not triggering error reconciliation (without significant refactoring).

The refactoring that I did do just that. If it's a "generic" error, it retriggers a reconciliation loop. Otherwise, it will just update the status, reason, and will stop the reconciliation.

> I think the problem is coming from copying from infinispan controller which has the same problem I think:

Do you mind fixing that as well in this PR (I haven't seen the changes yet)?

pkg/controller/kogitoinfra/infinispan.go
```diff
@@ -388,13 +389,13 @@ func (i *infinispanInfraReconciler) Reconcile() (requeue bool, resultErr error)
 		return false, errorForResourceNotReadyError(fmt.Errorf("infinispan instance %s not ready. Waiting for Condition.Type == True", infinispanInstance.Name))
 	}
 	if resultErr := i.updateInfinispanVolumesInStatus(infinispanInstance); resultErr != nil {
-		return false, nil
+		return true, resultErr
```
ricardozanini, line 392: Nice :+1:

pkg/controller/kogitoinfra/mongodb.go
```diff
@@ -0,0 +1,330 @@
...
+	log.Debugf("Got MongoDB instance %v", mongoDBInstance)
+	if mongoDBInstance.Status.Phase != mongodb.Running {
+		return false, errorForResourceNotReadyError(fmt.Errorf("mongoDB instance %s not ready. Waiting for Status.Phase == Running", mongoDBInstance.Name))
+	}
+	log.Info("MongoDB instance is running")
+	if resultErr = i.updateMongoDBAppPropsInStatus(mongoDBInstance); resultErr != nil {
+		return true, resultErr
+	}
+	if resultErr = i.updateMongoDBEnvVarsInStatus(mongoDBInstance); resultErr != nil {
+		return true, resultErr
```
ricardozanini, line 266: This error will immediately reconcile?

radtriste: why would it ? 
there is that https://github.com/kiegroup/kogito-cloud-operator/blob/f7ff09496d376272a5fb4a1500b4a783f2e47b20/pkg/controller/kogitoinfra/kogitoinfra_controller.go#L123

ricardozanini: Yup, we need to refactor this interface: https://issues.redhat.com/browse/KOGITO-3987

radtriste:
@sutaakar should be done now.

Tested on CRC ok. Launching smoke tests.

/jenkins test

examples/data-index-mongodb.yaml
```diff
@@ -8,24 +8,37 @@ spec:
...
+  name: kogito-mongodb
+spec:
+  resource:
+    apiVersion: mongodb.com/v1
+    kind: MongoDB
+    name: external-mongodb # to change if you don't use the example MongoDB below
+  infraProperties:
+    username: kogitouser
+    database: kogito_dataindex
+    # password and connection string will be guessed from MongoDB instance
```
sutaakar, line 26:
```suggestion
    # password and connection string will be retrieved from MongoDB instance
```

sutaakar (approved):
Looks good to me, great job.

ricardozanini (approved):

ricardozanini:
@spolti do you mind taking a look at it?

ricardozanini:
@radtriste can you please rebase?

ricardozanini:
@sutaakar have you verified, right? I'll merge after the CI. Had plenty of time for others to review it. We are on the verge to merge the SDK upgrade, I don't want to make @radtriste work on rebasing everything after that.

PR #643, opened 2020-10-30
[KOGITO-3736] - Reconciler error in Data index controller when Kogito…

vaibhavjainwiz:
…Runtime is deployed

Jira issue : https://issues.redhat.com/browse/KOGITO-3736

**Analysis**

As analysed provided logs, I have found that Data-index controller called to reconcile with name and namespace of KogitoRuntimeService. I happed because data-index controller is watching for config map whose owner is KogitoRuntime and handler is provided as of type `

EnqueueRequestForOwner`. Any controller should only watch object which it owns not others. But In this data-index controller is watching object which onwed by KogitoRuntime.

**Solution**:

KogitoRuntime should monitor configMap which it owns and update deployment object of data-index directly if protoBuf config map update. With this change there is no requirement to maintain seperate controller for data-index.


ricardozanini (approved):
+1 to this refactoring!

@Kaitou786 @vaibhavjainwiz  I remember you saying that we had to create another controller because of the unnecessary reconciliations in a scenario where we had lots of supporting services. I'm assuming that after this PR won't be a problem any more?

vaibhavjainwiz:
> +1 to this refactoring!
> 
> @Kaitou786 @vaibhavjainwiz I remember you saying that we had to create another controller because of the unnecessary reconciliations in a scenario where we had lots of supporting services. I'm assuming that after this PR won't be a problem any more?

I had tested it for couple of scenerios. Thing should work as expected. :crossed_fingers:

Kaitou786 (approved):

pkg/controller/kogitoruntime/kogitoruntime_controller.go
```diff
@@ -112,6 +112,10 @@ func (r *ReconcileKogitoRuntime) Reconcile(request reconcile.Request) (result re
 		return
 	}
 
+	if err = infrastructure.MountProtoBufConfigMapOnDataIndex(r.client, instance); err != nil {
+		return
```
spolti, line 116: shouldn't it print some warning message or debug if config map fails to mount?

spolti (changes requested):

ricardozanini:
@vaibhavjainwiz could you please rebase your PR and review the comments? Many thanks!

pkg/controller/kogitoruntime/kogitoruntime_controller.go
vaibhavjainwiz: I had added Error msg here

pkg/infrastructure/data_index.go
```diff
@@ -78,3 +89,97 @@ func InjectDataIndexURLIntoSupportingService(client *client.Client, namespace st
...
+	}
+	return nil
+}
+
+func appendProtoBufVolumeIntoDeployment(deployment *appsv1.Deployment, cm corev1.ConfigMap) {
+	configMapDefaultMode := protoBufConfigMapVolumeDefaultMode
+	volumeExists := false
+	for _, volume := range deployment.Spec.Template.Spec.Volumes {
+		if volume.Name == cm.Name {
+			volumeExists = true
```
sutaakar, line 134: You can directly call return here instead of operating with flag.

pkg/infrastructure/data_index.go
```diff
@@ -78,3 +89,97 @@ func InjectDataIndexURLIntoSupportingService(client *client.Client, namespace st
...
+}
+
+func appendProtoBufVolumeMountIntoDeployment(deployment *appsv1.Deployment, cm corev1.ConfigMap) {
+	for fileName := range cm.Data {
+
+		volumeMountExists := false
+		mountPath := path.Join(defaultProtobufMountPath, cm.Labels["app"], fileName)
+		for _, volumeMount := range deployment.Spec.Template.Spec.Containers[0].VolumeMounts {
+			if volumeMount.MountPath == mountPath {
+				volumeMountExists = true
```
sutaakar, line 161: Same here

pkg/infrastructure/data_index.go
vaibhavjainwiz: done

pkg/infrastructure/data_index.go
vaibhavjainwiz: done

pkg/controller/kogitosupportingservice/kogitosupportingservice_controller.go
```diff
@@ -224,10 +206,10 @@ type SupportingServiceResource interface {
 // map of all the kogitoSupportingService
 // Note: Data Index is not part of this map because it has it's own controller
 var kogitoSupportingServices = map[v1alpha1.ServiceType]SupportingServiceResource{
-	v1alpha1.Explainability: &ExplainabilitySupportingServiceResource{},
+	v1alpha1.DataIndex:      &DataIndexSupportingServiceResource{},
 	v1alpha1.JobsService:    &JobsServiceSupportingServiceResource{},
 	v1alpha1.MgmtConsole:    &MgmtConsoleSupportingServiceResource{},
-	v1alpha1.TaskConsole:    &TaskConsoleSupportingServiceResource{},
```
sutaakar, line 230: Why is TaskConsole removed here?

pkg/controller/kogitosupportingservice/kogitosupportingservice_controller.go
```diff
@@ -224,10 +206,10 @@ type SupportingServiceResource interface {
...
+	v1alpha1.Explainability: &ExplainabilitySupportingServiceResource{},
```
sutaakar, line 212: Can you please keep the alphabetical order here?

pkg/infrastructure/data_index.go
```diff
@@ -30,6 +33,14 @@ const (
...
 	// Data index WS URL env
 	dataIndexWSRouteEnv = "KOGITO_DATAINDEX_WS_URL"
+	// Default Proto Buf file path
+	defaultProtobufMountPath = KogitoHomeDir + "/data/protobufs"
+	// Proto Buf folder env
+	protoBufKeyFolder string = "KOGITO_PROTOBUF_FOLDER"
+	// Proto Buf watch env
+	protoBufKeyWatch string = "KOGITO_PROTOBUF_WATCH"
+	// Proto Buf volume mode
+	protoBufConfigMapVolumeDefaultMode int32 = 420
```
sutaakar, line 43: Can you move this constant to https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/framework/file_permissions.go ?

sutaakar:
Otherwise the code looks ok, will try it on OCP instance.

pkg/controller/kogitosupportingservice/kogitosupportingservice_controller.go
vaibhavjainwiz: Sorry.. It got removed in merging to master. I added it back

pkg/controller/kogitosupportingservice/kogitosupportingservice_controller.go
vaibhavjainwiz: done

pkg/infrastructure/data_index.go
vaibhavjainwiz: done

ricardozanini:
/jenkins test

@spolti @sutaakar can you guys please take a look? :)

pkg/controller/kogitosupportingservice/kogitodataindex_reconciler.go
```diff
@@ -0,0 +1,70 @@
...
+	"time"
+)
+
+// DataIndexSupportingServiceResource implementation of SupportingServiceResource
+type DataIndexSupportingServiceResource struct {
+}
+
+// Reconcile reconcile Data Index
+func (*DataIndexSupportingServiceResource) Reconcile(client *client.Client, instance *appv1alpha1.KogitoSupportingService, scheme *runtime.Scheme) (reconcileAfter time.Duration, err error) {
+	log.Infof("Reconciling KogitoDataIndex for %s in %s", instance.Name, instance.Namespace)
```
sutaakar, line 35: Seems to be a duplicit log entry, in the log I see:
```
{"level":"info","T":"2020-11-05T17:20:54.598Z","logger":"kogitosupportingservice_controller","msg":"Reconciling KogitoSupportingService for data-index in ksuta-test"}
{"level":"info","T":"2020-11-05T17:20:54.608Z","logger":"kogitosupportingservice_controller","msg":"Reconciling KogitoDataIndex for data-index in ksuta-test"}
```
I think this log can be removed.

sutaakar:
One general comment, I have noticed that *-protobuf-files config map is created even when there are no protobuf files in Kogito application. Would it have sense to skip creation of *-protobuf-files config map in that case?

ricardozanini:
> One general comment, I have noticed that *-protobuf-files config map is created even when there are no protobuf files in Kogito application. Would it have sense to skip creation of *-protobuf-files config map in that case?

In the old feature, we always had to create it because the operator wouldn't know if there were protobuf files or not in the service. Now that we fetch for this information, I would say that we could skip this creation. Can you file a JIRA to follow this up? :D

sutaakar:
Reported as https://issues.redhat.com/browse/KOGITO-3779

pkg/infrastructure/data_index.go
```diff
@@ -78,3 +87,87 @@ func InjectDataIndexURLIntoSupportingService(client *client.Client, namespace st
...
+		for _, volumeMount := range deployment.Spec.Template.Spec.Containers[0].VolumeMounts {
+			if volumeMount.MountPath == mountPath {
+				return
+			}
+		}
+
+		// append volume mount if its not exists
+		deployment.Spec.Template.Spec.Containers[0].VolumeMounts =
+			append(deployment.Spec.Template.Spec.Containers[0].VolumeMounts, corev1.VolumeMount{
+				Name:      cm.Name,
```
sutaakar, line 158: Minor? In case config map contains more protobuf files the Name parameter here is duplicated for every volume mount entry for those protobuf files. Shouldn't it be rather unique? Though I haven't see any issue with it so far.

sutaakar (approved):
Seems to work as expected. Approving.
Good job.

spolti (approved):

PR #649, opened 2020-11-10
[KOGITO-3720] Add test coverage for Grafana dashboard

sutaakar:

https://issues.redhat.com/browse/KOGITO-3720

### DIFF
--- a/test/features/grafana.feature
+++ b/test/features/grafana.feature
@@ -1,0 +1,25 @@
+Feature: Service Deployment: Grafana
+
+  Background:
+    Given Namespace is created 
+    And Kogito Operator is deployed
+    And Grafana Operator is deployed
+
+  # Testing just simple Grafana connection without Prometheus as I didn't figure out how to test that the dasboard is actually filled with proper data
+  # Integration with Grafana and Prometheus together was tested manually.
+  @smoke
+  Scenario: Deploy dmn-drools-quarkus-metrics service and verify that it successfully exports dashboard to Grafana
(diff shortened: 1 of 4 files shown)


sutaakar:
Timeout with Strimzi
/jenkins test

r00ta (approved):
Hi @sutaakar , thank you very much for this test, looks great to me!

test/features/grafana.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Service Deployment: Grafana
+
+  Background:
+    Given Namespace is created 
+    And Kogito Operator is deployed
+    And Grafana Operator is deployed
```
radtriste, line 6: Why not reusing the step `Kogito Operator is deployed with {XXX} operator` step ?

radtriste (approved):
one comment, otherwise looks good

test/features/grafana.feature
sutaakar: well, it is a thing to consider.
The original reason for creating `Kogito Operator is deployed with {XXX} operator` was that CLI was installing all depending operators in one step, so we need to adjust our steps to be same for CLI and CR approach, which is not relevant any more.

From my perspective the Grafana operator installation is completely separate action from Kogito operator installation, so it deserve separate step. That way each step is responsible on one action only.
It also helps in case of a failure, as we can better distinguish which step failed.

radtriste: oki

PR #658, opened 2020-11-16
[KOGITO-3215] - Support cluster-scoped deployment for operator

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-3215

Apart from the target issue the PR also resolves [KOGITO-3225](https://issues.redhat.com/browse/KOGITO-3225) and [KOGITO-2561](https://issues.redhat.com/browse/KOGITO-2561), Since all the changes were all related
Co-authored-by: Vaibhav Jain <vajain@redhat.com>


deploy/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: operators
```
ricardozanini, line 4: Maybe changing to `kogito-operator-system` to align with the upcoming SDK 1.0 migration? Is this namespace an SDK 0.x recommendation?

pkg/controller/kogitoruntime/deployer.go
```diff
@@ -143,6 +141,18 @@ func onDeploymentCreate(cli *client.Client, deployment *v1.Deployment, kogitoSer
...
+		return fmt.Errorf("error creating Service Account: %v", err)
+	}
+	// create role
+	if err := createRoleIfNotExists(cli, kogitoService.GetNamespace()); err != nil {
+		return fmt.Errorf("error creating role: %v", err)
+	}
+	//create rolebinding
+	if err := createRoleBindingIfNotExists(cli, kogitoService.GetNamespace()); err != nil {
+		return fmt.Errorf("error creating rolebinding: %v", err)
+	}
```
ricardozanini, line 155: Do you mind creating a function to handle this creation? or we can just move this logic to the `KogitoRuntime` controller instead, before using the API to deploy the service. This RBAC creation feature doesn't seems in the right place here.

ricardozanini (commented):
LGTM! Minor comments only, now we need to test some scenarios to make sure that Data Index and other services are deployed and reconciled correctly within the same namespace. Seems that BDD tests are failing as well, maybe we need to change something in there?

deploy/namespace.yaml
Kaitou786: I chose the namespace `operators` to be in coherence with the global-operators namespace when we install OLM in k8s. I am not sure but using yaml the installation would be done in k8s without olm installed, and by default in cli we are deploying the operator in cluster-scope, and to verify if it exists we check if the `kogito-operator` deployment is present in `global-operator-namespace` depending on the env(i.e.: OCP: `openshift-operators`, K8s: `operators`)

Kaitou786:
The BDD tests are failing because of some jenkins issue: ```
oading library jenkins-pipeline-shared-libraries@master

ERROR: Checkout failed

java.io.FileNotFoundException: https://api.github.com/

	at com.squareup.okhttp.internal.huc.HttpURLConnectionImpl.getInputStream(HttpURLConnectionImpl.java:243)

	at com.squareup.okhttp.internal.huc.DelegatingHttpsURLConnection.getInputStream(DelegatingHttpsURLConnection.java:210)

	at com.squareup.okhttp.internal.huc.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:25)

	at org.kohsuke.github.GitHubHttpUrlConnectionClient$HttpURLConnectionResponseInfo.bodyStream(GitHubHttpUrlConnectionClient.java:197)

	at org.kohsuke.github.GitHubResponse$ResponseInfo.getBodyAsString(GitHubResponse.java:313)

	at org.kohsuke.github.GitHubResponse.parseBody(GitHubResponse.java:80)

	at org.kohsuke.github.GitHubClient.lambda$fetch$0(GitHubClient.java:146)

	at org.kohsuke.github.GitHubClient.createResponse(GitHubClient.java:406)

	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:360)
``` @radtriste do you know more about this?

File: pkg/controller/kogitoruntime/deployer.go (hunk 2 above)
Comment (Kaitou786) [reply pkg/controller/kogitoruntime/deployer.go]: Moved to kogitoRuntime_controller

**Review by vaibhavjainwiz — APPROVED** [review]

File: deploy/namespace.yaml (hunk 1 above)
Comment (ricardozanini) [reply deploy/namespace.yaml:4]: > I chose the namespace operators to be in coherence with the global-operators namespace when we install OLM in k8s. 

Makes sense, but when installing without `olm`, as an user I might want to see my operator installed in its namespace instead of a global namespace of any kind, since without OLM that won't make sense whatsoever.

To summarize:

1. On environments with OLM, ~we~ (OLM) will install in the global namespace, no matter the name
2. On environments without OLM, on a cluster-scoped installation, the namespace should follow kubebuilder suggestion: `{operator-name}-system`
3. For namespace scoped installation, doesn't matter.

File: pkg/controller/kogitoruntime/rbac.go [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if err := createServiceViewerRoleBindingIfNotExists(namespace, client); err != nil {
+		return err
+	}
+	if err := createServiceViewerServiceAccountIfNotExists(namespace, client); err != nil {
+		return err
+	}
+	return nil
+
+}
```
ricardozanini, line 104: Although I'd say that moving these functions to its own go file is the right approach, we might want to add them to the `ReconcileKogitoRuntime` structure, this way we avoid passing the `client` context to all of them and restrict the number of static functions. wdyt?

spolti (approved):

deploy/namespace.yaml
Kaitou786: Changed

pkg/controller/kogitoruntime/rbac.go
Kaitou786: done

ricardozanini (approved):
Nice job, man! Many thanks!

sutaakar:
So, the preferred approach for now on would be to deploy the cloud wide Kogito operator, while keeping the possibility to still have the namespace limited operator?

ricardozanini:
> So, the preferred approach for now on would be to deploy the cloud wide Kogito operator, while keeping the possibility to still have the namespace limited operator?

The operator will be installed in a single namespace, watching all namespaces in the cluster. The recommended architecture now is to install supporting and custom services in a single namespace, just because we haven't tested other scenarios and we don't know the outcome.

If you're deploying custom services that do not depend on supporting services (i.e. Data Index), you can install them in any namespace, without having to install the operator.

sutaakar:
The automatic operator installation in CLI doesn't work properly when user doesn't have cluster rights. For example `kogito use-project ksuta-test` results in:
```
[karel@localhost kogito-cloud-operator]$ kogito use-project ksuta-test
Project set to 'ksuta-test'
Error: Error while trying to look for kogito-operator Operator installation: deployments.apps "kogito-operator" is forbidden: User "developer" cannot get resource "deployments" in API group "apps" in the namespace "openshift-operators" 
```

sutaakar:
Seems that I am not able to set `--cluster-scope` flag to false.
Using `--cluster-scope false` results in cluster scope being still set to true.

Kaitou786:
Seems that `cluster-scope` flag is respected when given like: `--cluster-scope=false` but not when given as `--cluster-scope false`
@ricardozanini do you know more about this :/

Kaitou786:
> The operator will be installed in a single namespace, watching all namespaces in the cluster. The recommended architecture now is to install supporting and custom services in a single namespace, just because we haven't tested other scenarios (having supporting services installed on different namespaces), hence we don't know the outcome.
>
I think  it'll fail right now, if the custom/supporting services are scattered around different namespace. Since right now are services like `DataIndex` tries to fetch Instances of `KogitoRuntime` in the same namespace they are running, We would need change how we currently fetch other resources. I have created this [jira](https://issues.redhat.com/browse/KOGITO-3881)
> If you're deploying custom services that do not depend on supporting services (i.e. Data Index), you can install them in any namespace, without having to install the operator.
Yes
> 
> @Kaitou786 I believe that is worth adding this note to the Release Notes, as well as this fix if it's not there already.
Okay, yes it'll need to be added in the `Release notes` so it is clear to user how operator is going to act.

Kaitou786:
I think the `developer` account can be a problem. As if I think a admin can install the operator but any user should be able to install any `kogitoSupportingService` or `KogitoRuntime` in its own namespace. But right now when we do `kogito install data-index` currently it'll try to see if the operator is already installed or `SilentlyInstallOperatorIfNotExists` but a developer user won't have perms to fetch any resource in the global operator namespace. Right now we check for `Deployment` we can also check for `Subscription` but as it is also a `namspace-scoped` resource we will have the same problem. The only way I think is remove this dependency and remove `SilenlyInstallOperatorIfNotExists` method and make it documented that the operator needs to be installed by your cluster-admin before performing any action like installing these services. And when `kogito install data-index` we assume that operator is already installed. But do let me know your thoughts or any better solution you guys think about this problem
^^ @ricardozanini @spolti @vaibhavjainwiz @sutaakar

ricardozanini:
I think a better approach is the CLI to verify if the current user has permissions to install the operator. If it has the permissions will install, if not we display a message that the operator is not there and needs to be installed before deploying Kogito resources in the cluster.

ricardozanini:
FWIW, I believe this problem also happens today. If the CLI is being executed by a non cluster-admin, it will fail as well.

Kaitou786:
> I think a better approach is the CLI to verify if the current user has permissions to install the operator. If it has the permissions will install, if not we display a message that the operator is not there and needs to be installed before deploying Kogito resources in the cluster.

Yes, that is true. But consider this situation. Suppose a cluster-admin has already installed the operator at cluster level. Now  each user has its own namespace where they want to try the `kogitoServices`. At this time when the `developer` user runs `kogito install data-index` the command will fail saying:
```
Error: Error while trying to look for kogito-operator Operator installation: deployments.apps "kogito-operator" is forbidden: User "developer" cannot get resource "deployments" in API group "apps" in the namespace "openshift-operators" 
```
Since before running any cli command we verify if the operator is already installed or not(by checking the deployment in the global operators namespace). Now even though the operator is already installed and the DataIndex would have been running if made by `YAML` the cli will fail at `verify if operator is installed` stage. I don't think this error will come with our current cli as if the cluster-admin has installed the operator in the `developers's` namespace and he runs `kogito install data-index` the deployment will be checked in the target namespace for which the `developer` user has perms to. Please correct me if I am wrong at any point here.

PS: I think this is pretty common scenarios as we are doing the same thing with our `AWS Openshift` with each user has only access to its own namespace for testing trying out the operator.

ricardozanini:
You can try the "light verification", which is using the discovery API to check if the CRD is installed in the cluster (needs to double check if the developer has permissions). Or we can do some more research and see if there's an API that any user can fetch on OLM to verify if a bundle is installed in the cluster or not.

sutaakar:
> Right now the only way I think is remove this dependency and remove SilenlyInstallOperatorIfNotExists method and make it documented that the operator needs to be installed by your cluster-admin before performing any action like installing these services.

I agree. Developer has limited capabilities to check whether the Kogito operator is installed in cluster and also the operator can be installed just by admin. So it would be better to install operator just using `kogito install operator`, for other commands just expect the operator to be installed already.

ricardozanini:
We will do the `CRD` check regardless. But the installation can be done if the user running the command has the required permissions.

hack/deploy-operator-on-ocp.sh
```diff
@@ -33,17 +33,21 @@ echo "Deploying Operator $OPERATOR_NAME"
...
-oc create -f deploy/role.yaml
-oc create -f deploy/role_binding.yaml
+oc apply -f deploy/clusterrole.yaml
+oc apply -f deploy/clusterrole_binding.yaml
 
 # Operator
 sed "s/name: kogito-operator/name: $OPERATOR_NAME/g" deploy/operator.yaml > $TMP_FOLDER/operator.yaml
 sed "s|image: quay.io/kiegroup/.*|image: $IMAGE|g" $TMP_FOLDER/operator.yaml > $TMP_FOLDER/operator-tmp.yaml
-oc create -f $TMP_FOLDER/operator-tmp.yaml
+sed '4 a \ \ namespace: operators' $TMP_FOLDER/operator-tmp.yaml > $TMP_FOLDER/operator.yaml
```
sutaakar, line 49: What is the purpose of this line? And why there is `operators` namespace specified here while other resources like service account have namespace `kogito-operator-system`?

Kaitou786: Ah yes, this needs to changed to `kogito-operator-system` as well. This line puts `namespace` field in our operator.yaml before applying it, since we want the operator to be installed in that namespace only, it is a hack script to deploy operator with yaml, during testing.

Kaitou786: changed

ricardozanini (approved):
LGTM, I believe we need now to review the BDD tests, right @sutaakar ?

sutaakar:
@ricardozanini Right, looking on it....

hack/go-vet.sh
```diff
@@ -25,9 +25,11 @@ operator-sdk generate crds --crd-version=v1beta1
 echo "Generating YAML installer"
 # Generate kogito-operator.yaml
 rm kogito-operator.yaml
-declare deploy_files=("./deploy/role.yaml" "./deploy/service_account.yaml" "./deploy/role_binding.yaml" "./deploy/operator.yaml")
+declare rbac_files=("./deploy/clusterrole.yaml" "./deploy/service_account.yaml" "./deploy/clusterrole_binding.yaml")
+cat ./deploy/namespace.yaml > kogito-operator.yaml; printf "\n---\n" >> kogito-operator.yaml
 for yaml in deploy/crds/*_crd.yaml; do cat "${yaml}" >> kogito-operator.yaml; printf "\n---\n" >> kogito-operator.yaml; done
-for yaml in "${deploy_files[@]}"; do cat "${yaml}" >> kogito-operator.yaml; printf "\n---\n" >> kogito-operator.yaml; done
+for yaml in "${rbac_files[@]}"; do cat "${yaml}" >> kogito-operator.yaml; printf "\n---\n" >> kogito-operator.yaml; done
+sed '4 a \ \ namespace: operators' ./deploy/operator.yaml >> kogito-operator.yaml
```
sutaakar, line 32: @Kaitou786 This should be fixed as well.

Kaitou786: Ah yes, thanks. Changed :)

sutaakar:
BDD test adjustments done and prepared as PR for this PR branch - https://github.com/Kaitou786/kogito-cloud-operator/pull/45

ricardozanini:
@sutaakar up to you man, code looks good. @Kaitou786 any other remarks?

cmd/kogito/command/shared/install_operator.go
```diff
@@ -113,14 +129,18 @@ func installOperatorWithYamlFiles(image string, namespace string, cli *client.Cl
 			return err
 		}
 	}
-
-	if err := decodeAndCreateKubeObject(box, fileServiceAccountYaml, &v1.ServiceAccount{}, namespace, cli, nil); err != nil {
+	if err := decodeAndCreateKubeObject(box, fileNamespaceYaml, &v1.Namespace{}, namespace, cli, nil); err != nil {
```
sutaakar, line 132: Nitpick - namespace should be created just when `namespaced` is false as otherwise operator is installed directly in user namespace.

cmd/kogito/command/shared/install_operator.go
```diff
@@ -113,14 +129,18 @@ func installOperatorWithYamlFiles(image string, namespace string, cli *client.Cl
...
 	}
-	if err := decodeAndCreateKubeObject(box, fileRoleYaml, &rbac.Role{}, namespace, cli, nil); err != nil {
+	if err := decodeAndCreateKubeObject(box, fileClusterRoleYaml, &rbac.ClusterRole{}, namespace, cli, nil); err != nil {
 		return err
 	}
-	if err := decodeAndCreateKubeObject(box, fileRoleBindingYaml, &rbac.RoleBinding{}, namespace, cli, nil); err != nil {
+	if err := decodeAndCreateKubeObject(box, fileServiceAccountYaml, &v1.ServiceAccount{}, namespace, cli, nil); err != nil {
+		return err
+	}
+	if err := decodeAndCreateKubeObject(box, fileClusterRoleBindingYaml, &rbac.ClusterRoleBinding{}, namespace, cli, func(object interface{}) {
```
sutaakar, line 141: Minor - In case the ClusterRoleBinding already exists then it would be fine just to add the new service account there, allowing namespaced installation of several independent Kogito operators.

sutaakar (approved):
Apart from the comments above looks good to me.

cmd/kogito/command/shared/install_operator.go
Kaitou786: This is a edge case in which user wants to install the operator in a specific namespace without the OLM, Since currently we don't have the idempotent approach, which should be implemented in [this](https://issues.redhat.com/browse/KOGITO-3880). So our option would be to delete the object and recreate with the new references. IMO we should not handle this right now, and can be worked on after we have the idempotent approach. Also I think we should also encourage user to use the cluster-wide option instead of the `namespaced` one Specially in the `yamlInstallation`.

cmd/kogito/command/shared/install_operator.go
Kaitou786: Changed

cmd/kogito/command/shared/install_operator.go
sutaakar: ok

ricardozanini:
@Kaitou786 do you still have work to do on this?

Kaitou786:
@ricardozanini .No, this is ready :D

ricardozanini:
@Kaitou786 can you please squash your commits, but the one with @sutaakar ? This way I preserve his contribution to the master.

ricardozanini:
Quay is not feeling weel: https://status.quay.io/incidents/tq5hjzpw3k8t
Will try to rerun OLM by the end of the day.

ricardozanini:
Quay is back

/jenkins test

ricardozanini:
@Kaitou786 seems that BDD is failing, can you take a look tomorrow?

sutaakar:
The test failure is related to https://issues.redhat.com/browse/KOGITO-3935, will be addressed in separate PR.

ricardozanini:
> The test failure is related to https://issues.redhat.com/browse/KOGITO-3935, will be addressed in separate PR.

This means that if I merge now, upcoming PRs will fail?

I'd rather wait for your PR to be ready and merge both. WDYT?

sutaakar:
> I'd rather wait for your PR to be ready and merge both. WDYT?

Fine for me, PR created (https://github.com/kiegroup/kogito-cloud-operator/pull/675)

PR #661, opened 2020-11-17
[KOGITO-3864] - KogitoInfra got delete when refering KogitoRuntime/Ko…

vaibhavjainwiz:
…gitoSupportingService delete

Jira issue : https://issues.redhat.com/browse/KOGITO-3864

### DIFF
--- a/pkg/controller/kogitoinfra/kogitoinfra_controller.go
+++ b/pkg/controller/kogitoinfra/kogitoinfra_controller.go
@@ -109,7 +109,7 @@ func (r *ReconcileKogitoInfra) Reconcile(request reconcile.Request) (reconcile.R
 
 	// make KogitoInfra as self owner so that it will not removed when kogito service referring to it deleted because
 	// kogito services are also become owner of kogitoInfra when reference of infra provided in Kogito services.
-	if resultErr = r.takeSelfKogitoInfraOwnership(r.client, instance, r.scheme); resultErr != nil {
+	if resultErr = r.takeSelfKogitoInfraOwnership(instance); resultErr != nil {
 		return reconcile.Result{}, resultErr
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/kogitoinfra/kogitoinfra_controller.go
```diff
@@ -108,7 +109,7 @@ func (r *ReconcileKogitoInfra) Reconcile(request reconcile.Request) (reconcile.R
 
 	// make KogitoInfra as self owner so that it will not removed when kogito service referring to it deleted because
 	// kogito services are also become owner of kogitoInfra when reference of infra provided in Kogito services.
-	if resultErr = framework.AddOwnerReference(instance, r.scheme, instance); resultErr != nil {
+	if resultErr = r.takeSelfKogitoInfraOwnership(r.client, instance, r.scheme); resultErr != nil {
```
ricardozanini, line 112: Hi! I believe only removing the reference will do, we don't need to add the `KogitoInfra` itself as an owner.

vaibhavjainwiz: Sorry @ricardozanini, I didn't got you here. Could you please explain it more?

ricardozanini: I don't see a reason why a [resource should be owned by itself](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/):

> Some Kubernetes objects are owners of other objects. For example, a ReplicaSet is the owner of a set of Pods. The owned objects are called dependents of the owner object. Every dependent object has a metadata.ownerReferences field that points to the owning object.

We use the `ownerReference` field to define how the object will be garbage collected by Kubernetes. For the `KogitoInfra` object, the deletion should not be tied to any other resource, since I can delete all my services that are linked to this infra, but still want to have it there (for later usage).

Users will control the deletion of the `KogitoInfra` object, not the operator. `KogitoInfra` should not be owned by anyone.

Having said that, we should leave the `ownerReference` of the `KogitoInfra` empty.

vaibhavjainwiz: We had define `ownerReference` of `KogitoRuntime` into `KogitoInfra` because KogitoRuntimeController should listen to any change happened inside its refered KogitInfra and retrigger its reconcilation loop. 

With current implementation, when KogitoInfra updates it check its owner references and only send reconcilation event to those KogitoRuntime which actually refer that infra.

ricardozanini: That's ok adding `KogitoRuntime` in `KogitoInfra`, but not `KogitoInfra` into `KogitoInfra` man:

```go
 framework.AddOwnerReference(kogitoInfra, r.scheme, kogitoInfra)
```

vaibhavjainwiz:
I am closing this PR and creating new PR to incorporate finalizer changes

PR #681, opened 2020-11-30
[KOGITO-3864] - KogitoInfra gets deleted when refering KogitoRuntime/KogitoSupportingService delete

vaibhavjainwiz:
…gitoSupportingService delete

Jira issue : https://issues.redhat.com/browse/KOGITO-3864

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,7 @@
 
 ## Bug Fixes
 - [KOGITO-3866](https://issues.redhat.com/browse/KOGITO-3866) Operator print error logs when KogitoRuntime delete
+- [KOGITO-3864](https://issues.redhat.com/browse/KOGITO-3864) KogitoInfra got delete when refering KogitoRuntime/KogitoSupportingService delete
 
 ## Known Issues
 The protobuf ConfigMap does not update in Spring Boot due to [this issue](https://issues.redhat.com/browse/KOGITO-3406).
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Bug Fixes
 - [KOGITO-3866](https://issues.redhat.com/browse/KOGITO-3866) Operator print error logs when KogitoRuntime delete
-- [KOGITO-3864](https://issues.redhat.com/browse/KOGITO-3864) KogitoInfra got delete when refering KogitoRuntime/KogitoSupportingService delete
+- [KOGITO-3864](https://issues.redhat.com/browse/KOGITO-3864) KogitoInfra gets deleted when refering KogitoRuntime/KogitoSupportingService delete
 
 ## Known Issues
 The protobuf ConfigMap does not update in Spring Boot due to [this issue](https://issues.redhat.com/browse/KOGITO-3406).
--- a/pkg/controller/kogitoruntime/kogitoruntime_controller.go
+++ b/pkg/controller/kogitoruntime/kogitoruntime_controller.go
@@ -28,8 +28,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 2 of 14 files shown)


ricardozanini (commented):
Seems good! The finalizers from Supporting and Runtime services seems pretty much the same thing to me, can't we reuse some of those functions?

vaibhavjainwiz:
> Seems good! The finalizers from Supporting and Runtime services seems pretty much the same thing to me, can't see reuse some of those functions?

yes..there method of addFinalizer and handleFinalization are almost similiar. I was not clear where I should put those methods so that both of controllers could call those method.

radtriste (approved):

RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@
 
 ## Bug Fixes
 - [KOGITO-3866](https://issues.redhat.com/browse/KOGITO-3866) Operator print error logs when KogitoRuntime delete
+- [KOGITO-3864](https://issues.redhat.com/browse/KOGITO-3864) KogitoInfra got delete when refering KogitoRuntime/KogitoSupportingService delete
```
spolti, line 9:
```suggestion
- [KOGITO-3864](https://issues.redhat.com/browse/KOGITO-3864) KogitoInfra gets deleted when refering KogitoRuntime/KogitoSupportingService delete
```

pkg/controller/kogitoruntime/kogitoruntime_controller.go
```diff
@@ -56,7 +58,16 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	err = c.Watch(&source.Kind{Type: &v1beta1.KogitoRuntime{}}, &handler.EnqueueRequestForObject{})
+	pred := predicate.Funcs{
```
spolti, line 61: duplicated code, maybe move to some common package?

spolti (changes requested):

RELEASE_NOTES.md
vaibhavjainwiz: done

vaibhavjainwiz:
> addFinalizer and handleFinalization

I had moved `addFinalizer` and `handleFinalization` to `kogito_service.go` file so that both finalizer can reuse it

pkg/controller/kogitoruntime/kogitoruntime_controller.go
vaibhavjainwiz: I had moved `addFinalizer` and `handleFinalization` method to common package. Are you refering to same duplicate code?

Kaitou786 (approved):

sutaakar (approved):

ricardozanini (approved):

pkg/infrastructure/kogito_infra.go
```diff
@@ -63,3 +64,21 @@ func IsKafkaResource(instance *v1beta1.KogitoInfra) bool {
...
 }
+
+// RemoveKogitoInfraOwnership remove provided kogito service owner reference from kogitoInfra
+func RemoveKogitoInfraOwnership(client *client.Client, owner v1beta1.KogitoService) error {
+	log.Debugf("Removing KogitoInfra ownership for %s", owner.GetName())
+	for _, kogitoInfraName := range owner.GetSpec().GetInfra() {
+		// load infra resource
+		kogitoInfraInstance, err := MustFetchKogitoInfraInstance(client, kogitoInfraName, owner.GetNamespace())
+		if err != nil {
+			return err
```
sutaakar, line 75: Do we need to throw an error in case KogitoInfra is not found? I think that in such case we can just skip it.
Otherwise when for example user deletes KogitoInfra first and then KogitoRuntime, the KogitoRuntime won't be removed as operator will get stuck in reconciliation cycle returning error.

ricardozanini: Agreed.

vaibhavjainwiz: @sutaakar Thanks for pointed it out.  I had incorporated your suggestion

ricardozanini:
@spolti can you take a final look?

/jenkins test

spolti (approved):

PR #705, opened 2021-01-05
KOGITO-3948 - Change trusty infinispan default image name

r00ta:
Jira: https://issues.redhat.com/browse/KOGITO-4091

With https://github.com/kiegroup/kogito-apps/pull/574 we are going to generate 2 different images for the trusty service: one that uses redis and another one that uses Infinispan. 

The related change on kogito-image is here https://github.com/kiegroup/kogito-images/pull/305

This PR aims to use the infinispan image as default, since a bigger discussion around Redis is still to be done. 

Still a draft, since the other PRs are still under review. 


radtriste:
I know this is still draft but I think you should change the issue number in this PR to https://issues.redhat.com/browse/KOGITO-4091 instead of KOGITO-3948 as it relates to https://github.com/kiegroup/kogito-images/pull/305 and we should merge those two PRs at the same time ;)

radtriste (approved):
lgtm

spolti:
one question, would be better do everything on one pr?
it depends on https://github.com/kiegroup/kogito-images/pull/305 which adds both images, does it mean that redis support will be available on images but not on operator?
@r00ta it is part of the discussion you've mentioned above?

ricardozanini:
@spolti speaking about Redis support, in the operator perspective, this doesn't matter since the connectivity configuration can be done via a custom ConfigMap.

r00ta:
Hi @spolti @ricardozanini , I don't think there should be a 1:1 mapping between images and operator. I think it's fine to have a Redis-based image for trusty even if we don't want to add a explicit KogitoInfra for Redis. As Ricardo pointed out the redis connector can be configured by an env variable. Actually also the other images should setup the connectors with env vars/configmaps for production env since KogitoInfra is intended for dev purposes only

ricardozanini:
> Actually also the other images should setup the connectors with env vars/configmaps for production env since KogitoInfra is intended for dev purposes only

This is not right. What's not intended for dev purposes is the creation of the third party instances, which we are about to remove. Connecting and configuring our resources using third-party CRs is ok for production envs.

r00ta:
> > Actually also the other images should setup the connectors with env vars/configmaps for production env since KogitoInfra is intended for dev purposes only
> 
> This is not right. What's not intended for dev purposes is the creation of the third party instances, which we are about to remove. Connecting and configuring our resources using third-party CRs is ok for production envs.

Hi @ricardozanini , my understanding was that KogitoInfra should not be used for prod envs, is this correct? If so, the only way to configure a KogitoSupportingService or a KogitoRuntime to connect to kafka or Infinispan is using env vars like `QUARKUS_KAFKA_BOOTSTRAP_SERVERS` and similar, right? r

ricardozanini:
> > > Actually also the other images should setup the connectors with env vars/configmaps for production env since KogitoInfra is intended for dev purposes only
> > 
> > 
> > This is not right. What's not intended for dev purposes is the creation of the third party instances, which we are about to remove. Connecting and configuring our resources using third-party CRs is ok for production envs.
> 
> Hi @ricardozanini , my understanding was that KogitoInfra should not be used for prod envs, is this correct? If so, the only way to configure a KogitoSupportingService or a KogitoRuntime to connect to kafka or Infinispan is using env vars like `QUARKUS_KAFKA_BOOTSTRAP_SERVERS` and similar, right? r

No, you can use KogitoInfra in prod environments if you are using Strimzi/ISPN Operator. You just have to create the CRs yourself and set them for your needs. KogitoInfra will configure everything for you. Of course, you can always configure the service with ConfigMaps or Envs.

spolti:
@r00ta can you please rebase?

PR #718, opened 2021-01-13
[KOGITO-4133] - API rule violation when running openapi-gen

Kaitou786:
The internal and the exposed name should be same, changing the internal name to `Volumes` since it is a list so makes more sense to me 

See: https://issues.redhat.com/browse/KOGITO-4133

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 
 ## Bug Fixes
 - [KOGITO-4110](https://issues.redhat.com/browse/KOGITO-4110) Single namespaced operator reacts on CRs created in different namespaces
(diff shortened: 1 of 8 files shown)


ricardozanini (changes requested):
`Volume` as `Env` (singular) is what we are using to follow with k8s conventions. The PodSpecTemplate keep them singular, I think we have to keep the same.

EDIT: actually on [`PodSpecTemplate`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core), the attribute is `volumes` (plural). Well, I guess they also missed the coherence point here. :laughing:

ricardozanini (approved):

spolti (approved):

PR #739, opened 2021-01-27
[KOGITO-3979] - Switch to current testing suite by OLM

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-3979


go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/zap v1.15.0
 	golang.org/x/crypto v0.0.0-20201016220609-9e8e0b390897 // indirect
+	golang.org/x/sync v0.0.0-20201207232520-09787c993a3a // indirect
```
ricardozanini, line 31: I don't think we need this update. If we need to use `go get` for our tooling, I believe we can adopt the `go.tools.mod` strategy instead. Every tool we need to install locally will use another mod file, leaving the project's file clean.

ricardozanini (commented):
Awesome man! MANY THANKS for this. Can you please take a look at the comment regarding the update in the mod file?

go.mod
Kaitou786: Do you mean to introduce a tools.go file in our repo and add all our tools there? something like [this](https://marcofranssen.nl/manage-go-tools-via-go-modules/)

ricardozanini: No, I mean this: https://github.com/golang/go/issues/25922#issuecomment-756675225.

In this example you linked, we ensure that the modules are in our mod file. And **don't want that**. So for tooling, we would use this command instead: `go get -modfile=go.tools.mod <the tool>`.

Kaitou786: Oh okay, so do you want me to update all our scripts in which `go get tool` if not found or just revert the changes of `go.mod` and `go.sum` in this PR and have a followup?
^^ @ricardozanini

ricardozanini: I think it's ok to add the `go.tools.mod` in this PR. It's a simple task. wdyt?

Kaitou786: Yeah it'll be okay. I'll add this option everywhere we do `go get `.  not sure if the mod file will be created at this moment from my system since all the tools are already present :)

ricardozanini: That's a good point. You can always clean up your local `$GOPATH`. hahahah

Kaitou786: Did that :p

Kaitou786:
Introduced to `go.tools.mod` =D

ricardozanini (approved):

ricardozanini:
@radtriste, can you take a look? I'd like to merge this before reviewing #740

hack/ci/ensure-image.sh
```diff
@@ -13,19 +13,15 @@
...
-    CLUSTER_NAME=$default_cluster_name
-fi
+version=$(getOperatorVersion)
 
-source ./hack/env.sh
-docker images
-echo "---> Loading Operator Image into Kind"
-kind load docker-image quay.io/kiegroup/kogito-cloud-operator:"$(getOperatorVersion)" --name ${CLUSTER_NAME}
+echo "=====> tagging image"
+docker tag quay.io/kiegroup/kogito-cloud-operator:${version} kind-registry:5000/kiegroup/kogito-cloud-operator
```
radtriste, line 24: shouldn't we use `IMAGE` env instead as the `Makefile` is doing ?
also shouldn't we add a task in the `Makefile` to do it ?

hack/ci/operator-ensure-manifests.sh
```diff
@@ -13,30 +13,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Script responsible for ensuring and correcting manifests as needed.
+set -e
 
-source ./hack/env.sh
+source ./hack/ci/ensure-image.sh
```
radtriste, line 19: why do you need to `source` there ?
any variable needed ?

hack/ci/operator-ensure-manifests.sh
```diff
@@ -13,30 +13,33 @@
...
+#replace image in target CSV
+sed -i  "s|quay.io/kiegroup/kogito-cloud-operator.*|kind-registry:5000/kiegroup/kogito-cloud-operator|g"  "${tempfolder}/community-operators/kogito-operator/${version}/kogito-operator.clusterserviceversion.yaml"
+#
+##ensure correct replace field is there
+sed -i "s|replace.*|replaces: kogito-operator.v${latest_released_version}|g" "${tempfolder}/community-operators/kogito-operator/${version}/kogito-operator.clusterserviceversion.yaml"
+#
+sed -i "s|${latest_released_version}|${version}|g" "${tempfolder}/community-operators/kogito-operator/kogito-operator.package.yaml"
+#
+echo "---> verify CSV updates"
+cat "${tempfolder}/community-operators/kogito-operator/${version}/kogito-operator.clusterserviceversion.yaml" | grep kind-registry:5000/kiegroup/kogito-cloud-operator
```
radtriste, line 41: I see `kind-registry:5000/kiegroup/kogito-cloud-operator` in many scripts ... maybe should we have it in env ?
that would avoid to update all scripts if there is a change...

hack/ci/ensure-image.sh
Kaitou786: Changed, and added make target `make olm-tests`

hack/ci/operator-ensure-manifests.sh
Kaitou786: It is taking IMAGE and KIND_IMAGE vars from ensure-image, also since ensure-image.sh calls env.sh we can have its function in this script like `getLatestOlmReleaseVersion`

hack/ci/operator-ensure-manifests.sh
Kaitou786: done

hack/ci/operator-ensure-manifests.sh
```diff
@@ -13,30 +13,33 @@
...
 
+# Script responsible for ensuring and correcting manifests as needed.
+set -e
 
-source ./hack/env.sh
+source ./hack/ci/ensure-image.sh
 
-OUTPUT="${PWD}/build/_output/operatorhub"
+tempfolder=$(mktemp -d)
+echo "$tempfolder"
```
radtriste, line 22:
```suggestion
echo "Temporary folder is $tempfolder"
```

radtriste (approved):

Kaitou786:
Just squashed the commis :)

ricardozanini:
@Kaitou786 if you have time, can you please also do the same for the Nexus Operator? Right now, we are struggling with their CI. This might fix the next release there.

Kaitou786:
Yeah I agree, I was waiting for the [PR](https://github.com/operator-framework/community-operators/pull/2933) to be merged before doing this

Kaitou786:
I had a talk with OLM guys and since many other are using GH actions to run local OLM tests, so they are planning to create a GH action for it, that'll be really easy for us to use =D

PR #742, opened 2021-02-01
[KOGITO-4332] Fixing role.yaml generation

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

See: https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/operator.20and.20mongo
JIRA: https://issues.redhat.com/browse/KOGITO-4332

While investigating this issue, I've found that the `role.yaml` file was never updated. I fixed that, but the MongoDB permissions were already there. Please see my comment in the JIRA. 

[I also opened a follow-up JIRA](https://issues.redhat.com/browse/KOGITO-4332) to fix the `*` in the resources. We need more time to review that. That's why I haven't done it in this PR. @Kaitou786, do you mind taking this one?

Many thanks for submitting your Pull Request :heart:! 

Please make sure your PR meets the following requirements:

- [x] You have read the [contributors' guide](https://github.com/kiegroup/kogito-cloud-operator/blob/master/README.md#contributing-to-the-kogito-operator)
- [x] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [x] Pull Request contains a link to the JIRA issue
- [x] Pull Request contains a description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Your feature/bug fix has a unit test that verifies it
- [ ] You've tested the new feature/bug fix in an actual OpenShift cluster
- [x] You've added a [RELEASE_NOTES.md](RELEASE_NOTES.md) entry regarding this change

PR #745, opened 2021-02-03
[1.2.x][KOGITO-4332] Fixing role.yaml generation

ricardozanini:
cherry-pick for #742

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/operator-framework/operator-marketplace v0.0.0-20190919183128-4ef67b2f50e9
 	github.com/sirupsen/logrus v1.7.0 // indirect
 	github.com/spf13/cobra v1.1.1
-	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/zap v1.15.0
 	golang.org/x/crypto v0.0.0-20201016220609-9e8e0b390897 // indirect
--- a/go.mod
+++ b/go.mod
@@ -23,13 +23,16 @@ require (
 	github.com/openshift/client-go v0.0.0-20200116152001-92a2713fa240
 	github.com/operator-framework/operator-lifecycle-manager v0.0.0-20200321030439-57b580e57e88
 	github.com/operator-framework/operator-marketplace v0.0.0-20190919183128-4ef67b2f50e9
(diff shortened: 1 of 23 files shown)


ricardozanini:
@radtriste, something broke this commit, and I don't have the time to look at it right now (family appointment). I've been swamped with other things all day. @Kaitou786, since you will be the first one to wake up, can you give a quick look? My guess is the kube client. Only the `cmd` package is failing, and the only difference is your commit, which only has shell script changes.

I'll take a look tomorrow morning.

go.mod
```diff
@@ -23,13 +23,15 @@ require (
 	github.com/openshift/client-go v0.0.0-20200116152001-92a2713fa240
 	github.com/operator-framework/operator-lifecycle-manager v0.0.0-20200321030439-57b580e57e88
 	github.com/operator-framework/operator-marketplace v0.0.0-20190919183128-4ef67b2f50e9
-	github.com/spf13/cobra v1.0.0
+	github.com/sirupsen/logrus v1.7.0 // indirect
+	github.com/spf13/cobra v1.1.1
```
Kaitou786, line 27: Is there reason why we upgraded our cobra version?
The unit test passes when we downgrade to our older version `v1.0.0`
Can we change it back and if we want to update the cobra version can be done in a separate issue with the fix of unit test?
It's the output differs in a failing command with this version of cobra

Kaitou786, line 27: The error in unit tests is because of cobra version being increased, the easy fix would be to revert to `v1.0.0` or apply this [fix PR](https://github.com/ricardozanini/kogito-cloud-operator/pull/13) :)

sutaakar (approved):

go.mod
radtriste: I think the update of cobra to v1.1.1 was done due to gogog v0.11.0 on master. But 1.2.x is still with v0.10.0, so we should not need it.

radtriste: @Kaitou786 I have the rights to push to Zanini's repository :D
So I took your commit in above mentioned PR and cherry-picked it to this PR. Let's see ...

Kaitou786 (approved):

ricardozanini:
@radtriste @Kaitou786 thanks guys! Yeah, I was in a rush yesterday to my appointment and missed the upgrade. Thanks!

PR #755, opened 2021-02-10
[KOGITO-4412] New service installer approach for BDD tests

sutaakar:

https://issues.redhat.com/browse/KOGITO-4412

### DIFF
--- a/test/features/olm/cli_new_project.feature
+++ b/test/features/olm/cli_new_project.feature
@@ -4,13 +4,6 @@
 @cli
 Feature: CLI: Project
 
-  Scenario: CLI create project    
-    When CLI create namespace
-
-    Then Kogito operator should be installed with dependencies
-
-#####
-
   Scenario: CLI create project with Kogito Data Index   
     When CLI create namespace with Kogito Data Index
 
--- a/test/features/olm/cli_use_project.feature
+++ b/test/features/olm/cli_use_project.feature
@@ -7,13 +7,6 @@ Feature: CLI: Project
   Background:
     Given Namespace is created
 
-  Scenario: CLI use project
-    When CLI use namespace
-
-    Then Kogito operator should be installed with dependencies
-
-#####
-
   Scenario: CLI use project and set Kogito Data Index
     When CLI use namespace with Kogito Data Index
 
--- a/test/framework/operator.go
+++ b/test/framework/operator.go
@@ -16,17 +16,12 @@ package framework
 
 import (
 	"fmt"
+
 	"github.com/kiegroup/kogito-cloud-operator/core/infrastructure"
 	"github.com/kiegroup/kogito-cloud-operator/core/logger"
 	"github.com/kiegroup/kogito-cloud-operator/meta"
-	"strings"
 
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-
-	coreapps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
-	rbac "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
@@ -34,7 +29,6 @@ import (
 	"github.com/kiegroup/kogito-cloud-operator/core/framework"
 	"github.com/kiegroup/kogito-cloud-operator/core/operator"
 	"github.com/kiegroup/kogito-cloud-operator/test/config"
-	"github.com/kiegroup/kogito-cloud-operator/version"
 
 	olmapiv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
 	olmapiv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
@@ -52,24 +46,15 @@ const (
 	kogitoOperatorDeploymentName = "kogito-operator-controller-manager"
 
 	// kogitoCatalogSourceName name of the CatalogSource containing Kogito bundle for BDD tests
-	kogitoCatalogSourceName           = "bdd-tests-kogito-catalog"
-	kogitoOperatorSubscriptionName    = "kogito-operator"
-	kogitoOperatorSubscriptionChannel = "alpha"
+	kogitoCatalogSourceName = "bdd-tests-kogito-catalog"
 
 	openShiftMarketplaceNamespace = "openshift-marketplace"
 )
 
-type dependentOperator struct {
-	operatorPackageName string
-	timeoutInMin        int
-	channel             string
-	clusterWide         bool
-}
-
-type operatorCatalog struct {
-	source       string
-	namespace    string
-	dependencies map[string]dependentOperator
+// OperatorCatalog OLM operator catalog
+type OperatorCatalog struct {
+	source    string
+	namespace string
 }
 
 var (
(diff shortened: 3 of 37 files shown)


test/installers/service_installer.go
```diff
@@ -0,0 +1,151 @@
...
+					success = false
+				}
+			}
+		}
+	}
+	return
+}
+
+// UninstallFromCluster uninstalls the cluster wide service from whole cluster. Returns false in case of any error while uninstalling.
+func UninstallFromCluster() (success bool) {
```
radtriste, line 87: `UninstallServicesFromCluster` ?

test/installers/service_installer.go
```diff
@@ -0,0 +1,151 @@
...
+	// Uninstall service from specific namespace
+	uninstallFromNamespace func(string) error
+	// Uninstall service from whole cluster
+	uninstallFromCluster func() error
+}
+
+// Install the service into cloud
+func (installer *ServiceInstaller) Install(namespace string) error {
+	// Store service installer for namespace to use for uninstalling purposes
+	if sis, loaded := installedServices.LoadOrStore(namespace, []*ServiceInstaller{installer}); loaded {
```
radtriste, line 46: storing/loading before the installation ? 
Could that create any problems when uninstalling it if it failed in-between for some reason ?

test/installers/service_installer.go
```diff
@@ -0,0 +1,151 @@
...
+		for _, si := range sis.([]*ServiceInstaller) {
+			// Remove all the CRs created in the namespace, triggered for all installed services in advance as some CRs may have other CRs as owners
+			if err := si.deleteCrsWithoutOwner(namespace); err != nil {
+				framework.GetLogger(namespace).Error(err, "Error removing CRs from namespace")
+				success = false
+			}
+		}
+		for _, si := range sis.([]*ServiceInstaller) {
+			// Wait until CRs are removed
+			if err := si.waitForAllCrsRemoval(namespace, 1*time.Minute); err != nil {
```
radtriste, line 71: put `1*time.Minute` in a constant ?

test/installers/service_installer.go
```diff
@@ -0,0 +1,151 @@
...
+		if len(cr.GetOwnerReferences()) == 0 {
+			crsWithoutOwner = append(crsWithoutOwner, cr)
+		}
+	}
+	return crsWithoutOwner, nil
+}
+
+func (installer *ServiceInstaller) waitForAllCrsRemoval(namespace string, timeout time.Duration) (err error) {
+	return framework.WaitFor(namespace, "Removal of all related CRs without owner", timeout, func() (bool, error) {
+		crs, err := installer.getAllCrsInNamespace(namespace)
```
radtriste, line 144: shouldn't it be `getCrsWithoutOwner` ?

test/installers/knative_installer.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	output, err = framework.CreateCommand("oc", "apply", "-f", fmt.Sprintf("https://github.com/knative/eventing/releases/download/%s/mt-channel-broker.yaml", knativeEventingVersion)).Execute()
+	if err != nil {
+		framework.GetMainLogger().Error(err, fmt.Sprintf("Applying eventing mt-channel broker failed, output: %s", output))
+		return fmt.Errorf("Error applying eventing mt-channel broker")
+	}
+
+	// Start namespace logging
+	go func() {
```
radtriste, line 79: Shouldn't we start the pod log collector before the installation and stop it after the delete of resources ?

test/installers/knative_installer.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Stop namespace logging
+	if err := framework.StopPodLogCollector(knativeEventingNamespace); err != nil {
+		framework.GetMainLogger().Error(err, "Error stopping log collector", "namespace", knativeEventingNamespace)
+	}
+
+	output, err := framework.CreateCommand("oc", "delete", "-f", fmt.Sprintf("https://github.com/knative/eventing/releases/download/%s/mt-channel-broker.yaml", knativeEventingVersion)).Execute()
+	if err != nil {
+		framework.GetMainLogger().Error(err, fmt.Sprintf("Deleting eventing mt-channel broker failed, output: %s", output))
+		return fmt.Errorf("Error deleting eventing mt-channel broker")
```
radtriste, line 100: Returning means we don't even try to remove other resources.
Shouldn't we log and then at the end of all deletion return error if any happens ?

test/installers/service_installer.go
sutaakar: IMHO even in case of installation failure we should try to uninstall the service, to cleanup the environment.
The possible uninstallation failure should be just logged, shouldn't return an error. Uninstallation shouldn't interrupt the flow or fail the otherwise passing test.

test/installers/service_installer.go
sutaakar: Not necessary, this function waits until all the CRs are removed, to make sure that service can be removed without leaving there any hanging CR. Regardless of whether it is with or without owner.
Some CRs have finalizer and need to be processed by service even when there is owner defined.

test/installers/knative_installer.go
sutaakar: I will move the logging logic into `service_installer.go` to have it in one place rather than implementing it for every needed service.

test/installers/knative_installer.go
sutaakar: +1

test/installers/service_installer.go
radtriste: Then I guess, the `WaitFor` text just before should be change or ?

sutaakar: Right, copy paste error :)

sutaakar:
@radtriste Comments should be addressed

radtriste (approved):

sutaakar:
The solution was refactored "a bit", all the shared functionality was abstracted into service_installer.go

sutaakar:
Added a Grafana OLM installation as an OLM approach sample.
Will convert now all the operators we have currently.

test/installers/service_installer.go
```diff
@@ -0,0 +1,287 @@
...
+
+	return installer.waitForClusterYamlServiceRunning()
+}
+
+func (installer *YamlClusterWideServiceInstaller) getAllCrsInNamespace(namespace string) ([]meta.ResourceObject, error) {
+	return installer.getAllClusterYamlCrsInNamespace(namespace)
+}
+
+func (installer *YamlClusterWideServiceInstaller) uninstallFromCluster() error {
+	stopNamespaceMonitoring(installer.installationNamespace)
```
radtriste, line 152: we don't remove here from `installedClusterWideServices` ?

sutaakar: Yes we do. That is what I forgot to fix, thanks for reminding :)

sutaakar: Removing is done as part of public uninstalling functions.

sutaakar:
Should be ready now.
`service_installer.go` contains the common functionality shared by all services.
Every service installation has own go file, there can be several installation options in that file. Currently there is implemented cluster wide or namespaced YAML or OLM installation.
Logic deciding what type of installation is used is specified separately for each service in its function retrieving service installer. So it is possible to combine different approaches for different services.

Finally there is a proper cleanup once scenario or test suite finish.

radtriste (approved):
nice work !
code looks good to me.
2 minor comments, not directly related to the PR

test/framework/operator.go
```diff
@@ -315,15 +231,15 @@ func InstallClusterWideOperator(subscriptionName, channel string, catalog operat
...
+func WaitForOperatorRunning(namespace, operatorPackageName string, catalog OperatorCatalog, timeoutInMin int) error {
 	return WaitForOnOpenshift(namespace, fmt.Sprintf("%s operator running", operatorPackageName), timeoutInMin,
 		func() (bool, error) {
 			return IsOperatorRunning(namespace, operatorPackageName, catalog)
 		})
 }
 
 // WaitForClusterWideOperatorRunning waits for a cluster wide operator to be running
-func WaitForClusterWideOperatorRunning(operatorPackageName string, catalog operatorCatalog, timeoutInMin int) error {
+func WaitForClusterWideOperatorRunning(operatorPackageName string, catalog OperatorCatalog, timeoutInMin int) error {
```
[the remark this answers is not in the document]
sutaakar: Yes, it can

test/framework/operator.go
```diff
@@ -230,67 +196,17 @@ func WaitForKogitoOperatorRunning(namespace string) error {
...
-	return nil
-}
-
 // WaitForClusterWideKogitoOperatorRunningUsingOlm waits for a cluster wide Kogito operator to be running in OLM
 func WaitForClusterWideKogitoOperatorRunningUsingOlm() error {
 	olmNamespace := config.GetOlmNamespace()
 	return WaitForOnOpenshift(olmNamespace, fmt.Sprintf("%s operator in namespace %s running", kogitoOperatorSubscriptionName, olmNamespace), kogitoOperatorTimeoutInMin,
 		func() (bool, error) {
-			return IsOperatorRunning(olmNamespace, kogitoOperatorSubscriptionName, customKogitoOperatorCatalog)
+			return IsOperatorRunning(olmNamespace, kogitoOperatorSubscriptionName, CustomKogitoOperatorCatalog)
```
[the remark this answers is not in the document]
sutaakar: yes

test/framework/operator.go
```diff
@@ -157,7 +123,7 @@ func DeployClusterWideKogitoOperatorFromYaml(deploymentNamespace string) error {
 
 // DeployClusterWideKogitoOperatorUsingOlm Deploy Kogito Operator watching for all namespaces using OLM, return all objects created for deployment
 func DeployClusterWideKogitoOperatorUsingOlm() error {
```
MarianMacik, line 125: Same here.

test/framework/operator.go
```diff
@@ -127,7 +93,7 @@ var (
 // DeployNamespacedKogitoOperatorUsingOlm Deploy Kogito Operator watching for specified namespace using OLM, return all objects created for deployment
 // Will be deployed in the same namespace as it will watch for
 func DeployNamespacedKogitoOperatorUsingOlm(deploymentNamespace string) error {
```
MarianMacik, line 95: Do we still need this?

test/framework/operator.go
```diff
@@ -127,7 +93,7 @@ var (
...
-	return InstallOperator(deploymentNamespace, kogitoOperatorSubscriptionName, kogitoOperatorSubscriptionChannel, customKogitoOperatorCatalog)
+	return InstallOperator(deploymentNamespace, kogitoOperatorSubscriptionName, kogitoOperatorSubscriptionChannel, CustomKogitoOperatorCatalog)
 }
 
 // DeployClusterWideKogitoOperatorFromYaml Deploy Kogito Operator watching for all namespaces from yaml files, return all objects created for deployment
```
MarianMacik, line 99: Same here.

test/scripts/prune_namespaces.go
```diff
@@ -19,6 +19,11 @@ import (
 )
 
 func main() {
+	// Create kube client
+	if err := framework.InitKubeClient(); err != nil {
+		panic(err)
+	}
+
```
MarianMacik, line 26: Was there any issue with that?

test/features/olm/cli_new_project.feature
```diff
@@ -4,13 +4,6 @@
 @cli
 Feature: CLI: Project
 
-  Scenario: CLI create project    
-    When CLI create namespace
-
-    Then Kogito operator should be installed with dependencies
-
-#####
-
```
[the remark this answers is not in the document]
sutaakar: No, we don't. Installing Kogito operator when namespace is created was removed some time ago, this is a leftover which wasn't deleted.

test/features/olm/cli_use_project.feature
```diff
@@ -7,13 +7,6 @@ Feature: CLI: Project
   Background:
     Given Namespace is created
 
-  Scenario: CLI use project
-    When CLI use namespace
-
-    Then Kogito operator should be installed with dependencies
-
-#####
-
```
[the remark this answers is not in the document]
sutaakar: Same as above

test/installers/kogito_installer.go
```diff
@@ -0,0 +1,182 @@
...
+
+// GetKogitoInstaller returns Kogito installer
+func GetKogitoInstaller() (ServiceInstaller, error) {
+	if config.IsOperatorNamespaced() && config.IsOperatorInstalledByYaml() {
+		return nil, errors.New("Installing namespace scoped Kogito operator using YAML files is not supported")
+	}
+
+	if config.IsOperatorInstalledByYaml() {
+		return &kogitoYamlClusterInstaller, nil
+	}
```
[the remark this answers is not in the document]
sutaakar: OLM can install Kogito operator both namespaced and cluster wide, so no need to check namespaced installation?

test/installers/mongodb_installer.go
```diff
@@ -0,0 +1,160 @@
...
+func installMongoDbUsingYaml(namespace string) error {
+	framework.GetLogger(namespace).Info("Deploy MongoDB from yaml files", "file uri", mongoDBOperatorDeployFilesURI)
+
+	if !framework.IsMongoDBAvailable(namespace) {
+		if err := framework.LoadResource(namespace, mongoDBOperatorDeployFilesURI+"crds/mongodb.com_mongodb_crd.yaml", &apiextensionsv1beta1.CustomResourceDefinition{}, nil); err != nil {
+			return err
+		}
+	}
+
+	if err := framework.LoadResource(namespace, mongoDBOperatorDeployFilesURI+"service_account.yaml", &corev1.ServiceAccount{}, nil); err != nil {
```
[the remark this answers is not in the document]
sutaakar: `framework.LoadResource` also set the namespace for created resource. In this case it wouldn't be a problem as namespace is not set in YAMLs. Also `framework.LoadResource` allows adjustment of resources before they are applied (that will come handy for latest MongoDB operator once we will update, the installation approach is significantly refactored).

I think it could be ok to leave the current implementation as it is going to be refactored anyway in near future.

test/framework/operator.go
sutaakar: No, we don't, will delete it. Thanks for noticing.

test/installers/kogito_installer.go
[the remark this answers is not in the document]
MarianMacik: I meant something like:
```
if config.IsOperatorInstalledByYaml() {
    if config.IsOperatorNamespaced() {
        return nil, errors.New("Installing namespace scoped Kogito operator using YAML files is not supported")
    }
    return &kogitoYamlClusterInstaller, nil
}
```
And remove the first if. So in the end there is one less call to `IsOperatorInstalledByYaml()` and both branches (OLM and Yaml) use the same format.

test/scripts/prune_namespaces.go
sutaakar: Yes, uninitialised client was throwing nill dereference error, not sure why or what changed. Maybe a result of recent operator refactoring.

test/installers/kogito_installer.go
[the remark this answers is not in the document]
sutaakar: Ok, will do.

MarianMacik (approved):
Looks good, thank you for the changes!

PR #784, opened 2021-03-08
[KOGITO-233] - Migrate Operator image generation to CEKit

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-233


Kaitou786:
PR checks can be triggered after https://github.com/kiegroup/kogito-pipelines/pull/125 is merged and the image is updated :)

sutaakar (approved):
Looks good to me, seems to work as expected.

Makefile
```diff
@@ -77,11 +77,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build:
-ifeq ($(BUILDER),buildah)
-	$(BUILDER) -t ${IMAGE} bud .
-else
-	$(BUILDER) build . -t ${IMAGE}
-endif
+	cekit -v build $(BUILDER)
```
radtriste, line 80: how do you pass the `IMAGE` information there ?

radtriste (approved):
lgtm. one question though

Makefile
Kaitou786: It is defined in the image descriptor [here](https://github.com/kiegroup/kogito-cloud-operator/pull/784/files#diff-eda18bf907bf5619e4d297d9a06cd67ea32859f6fbff3199904d6e238091788cR12)
I added its entry in the `bump-version.sh` so it is changed at our release time :)

radtriste: yeah but we have this setup: https://github.com/kiegroup/kogito-cloud-operator/blob/master/Makefile#L20
which means that if I set the `IMAGE` var, I would expect the build to create the image with that name. maybe adding a docker tag in the `docker-build` make command ?
moreover, the push would then failed also.

radtriste:
did you also test it on RHBA Jenkins ?

Makefile
Kaitou786: I am not sure how can I define the image tag from cekit at runtime, let try some things with runtime override and update here

Kaitou786:
> did you also test it on RHBA Jenkins ?

Yeah because of kogito-images we already had the dependicies on the node and from the last build of image the cekit version on the node is 3.8.0 so we should be fine :)

Makefile
Kaitou786: I raised this in `CEKit` channel and looks like it won't be possible for now. I raised an [issue](https://github.com/cekit/cekit/issues/689) as advied by there team. So for now I went ahead with the `tag` approach

ricardozanini (approved):
LGTM

radtriste:
> > did you also test it on RHBA Jenkins ?
> 
> Not test but it should work because of kogito-images we already had the dependicies on the node and from the last build of node image the cekit version on the node is 3.8.0 so we should be fine :)

Gonna give it a quick go to be sure

radtriste:
all good :)

radtriste:
@spolti up to you for review and merge

Makefile
spolti: we can use --overrides to do it, e.g.:

```
      Please note that order matters; overrides on the right hand side take precedence! For
      example:

          $ cekit build --overrides '{"from": "custom/image:1.0"}' --overrides '{"from":
          "custom/image:2.0"}'

```

spolti: something like 
```cekit -v build $(BUILDER) --overrides '{"version": "custom-tag"}'```

does it work for this use case?

another option as well would be tag the images manually after build.

spolti:
@radtriste yep, changes sounds good to me, but before merge, lets wait @Kaitou786 response on the suggestion I made above regarding the tags.

Makefile
Kaitou786: > something like
> `cekit -v build $(BUILDER) --overrides '{"version": "custom-tag"}'`
> 
> does it work for this use case?
> 
No it won't be it'll only change the version but we want the whole image name to change if IMAGE is given by developer I assume it is because he would want the image to be pushed in his namespace on quay. But seems it is not possible in cekit with the multi-stage build right now.

> another option as well would be tag the images manually after build.

Yeah, that's what I have done right now in the `make docker-build` target we build the image and then tag it with `${IMAGE}`.

spolti: Ok @Kaitou786 , thanks.

PR #790, opened 2021-03-11
[KOGITO-4469] - Conditions should reflect latest evaluated status

vaibhavjainwiz:

Jira issue : https://issues.redhat.com/browse/KOGITO-4469

### DIFF
--- a/api/conditions.go
+++ b/api/conditions.go
@@ -60,9 +60,9 @@ type ConditionInterface interface {
 
 // ConditionMetaInterface defines the base information for kogito services conditions
 type ConditionMetaInterface interface {
-	SetDeployed() bool
-	SetProvisioning() bool
-	SetFailed(reason KogitoServiceConditionReason, err error)
+	NewDeployedCondition() ConditionInterface
+	NewProvisioningCondition() ConditionInterface
+	NewFailedCondition(reason KogitoServiceConditionReason, err error) ConditionInterface
 	GetConditions() []ConditionInterface
 	SetConditions(conditions []ConditionInterface)
 }
--- a/core/framework/condition_collection.go
+++ b/core/framework/condition_collection.go
@@ -1,0 +1,62 @@
+// Copyright 2021 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import "github.com/kiegroup/kogito-cloud-operator/api"
+
+const maxBufferCondition = 5
+
+// ConditionCollection ...
+type ConditionCollection interface {
+	AddCondition(condition api.ConditionInterface) []api.ConditionInterface
+}
+
+type conditionCollection struct {
+	conditions []api.ConditionInterface
+}
+
+// NewConditionCollection ...
+func NewConditionCollection(conditions []api.ConditionInterface) ConditionCollection {
+	return &conditionCollection{
+		conditions: conditions,
+	}
+}
+
+func (c *conditionCollection) AddCondition(newCondition api.ConditionInterface) []api.ConditionInterface {
+	for i, condition := range c.conditions {
+		if condition != nil && condition.GetType() == newCondition.GetType() {
+			c.removeOldCondition(i)
+		}
+	}
+	c.pushCondition(newCondition)
+	return c.conditions
+}
+
+func (c *conditionCollection) removeOldCondition(index int) {
+	c.conditions = append(c.conditions[:index], c.conditions[index+1:]...)
+}
+
+func (c *conditionCollection) pushCondition(newCondition api.ConditionInterface) {
+	if len(c.conditions) >= maxBufferCondition {
+		c.removeOldestCondition()
+		c.pushCondition(newCondition)
+	} else {
+		c.conditions = append(c.conditions, newCondition)
+	}
+}
+
+func (c *conditionCollection) removeOldestCondition() {
+	c.conditions = c.conditions[1:len(c.conditions)]
+}
--- a/core/framework/condition_collection_test.go
+++ b/core/framework/condition_collection_test.go
@@ -1,0 +1,62 @@
+// Copyright 2021 Red Hat, Inc. and/or its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package framework
+
+import (
+	"fmt"
+	"github.com/kiegroup/kogito-cloud-operator/api"
+	"github.com/kiegroup/kogito-cloud-operator/api/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func TestConditionCollection(t *testing.T) {
+	conditions := make([]api.ConditionInterface, 5)
+	conditionCollection := NewConditionCollection(conditions)
+	newCondition := v1beta1.Condition{
+		Type:               api.DeployedConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+	conditions = conditionCollection.AddCondition(newCondition)
+	fmt.Println(conditions)
+
+	newCondition2 := v1beta1.Condition{
+		Type:               api.FailedConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+	conditions = conditionCollection.AddCondition(newCondition2)
+	fmt.Println(conditions)
+
+	newCondition3 := v1beta1.Condition{
+		Type:               api.ProvisioningConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+	conditions = conditionCollection.AddCondition(newCondition3)
+	fmt.Println(conditions)
+
+	newCondition4 := v1beta1.Condition{
+		Type:               api.DeployedConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+	conditions = conditionCollection.AddCondition(newCondition4)
+	fmt.Println(conditions)
+
+	newCondition5 := v1beta1.Condition{
+		Type:               api.ProvisioningConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+	conditions = conditionCollection.AddCondition(newCondition5)
+	fmt.Println(conditions)
+}
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -20,8 +20,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const maxBufferCondition = 5
-
 // Condition is the detailed condition for the resource
 type Condition struct {
 	Type               api.ConditionType                `json:"type"`
@@ -86,54 +84,31 @@ func (c *ConditionsMeta) SetConditions(conditions []api.ConditionInterface) {
 	c.Conditions = newConditions
 }
 
-// SetDeployed Updates the condition with the DeployedCondition and True status
-func (c *ConditionsMeta) SetDeployed() bool {
-	size := len(c.Conditions)
-	if size > 0 && c.Conditions[size-1].Type == api.DeployedConditionType {
-		return false
-	}
-	condition := Condition{
+// NewDeployedCondition ...
+func (c *ConditionsMeta) NewDeployedCondition() api.ConditionInterface {
+	return Condition{
 		Type:               api.DeployedConditionType,
 		Status:             corev1.ConditionTrue,
 		LastTransitionTime: metav1.Now(),
 	}
-	c.Conditions = c.addCondition(condition)
-	return true
 }
 
-// SetProvisioning Sets the condition type to Provisioning and status True if not yet set.
-func (c *ConditionsMeta) SetProvisioning() bool {
-	size := len(c.Conditions)
-	if size > 0 && c.Conditions[size-1].Type == api.ProvisioningConditionType {
-		return false
-	}
-	condition := Condition{
+// NewProvisioningCondition ...
+func (c *ConditionsMeta) NewProvisioningCondition() api.ConditionInterface {
+	return Condition{
 		Type:               api.ProvisioningConditionType,
 		Status:             corev1.ConditionTrue,
 		LastTransitionTime: metav1.Now(),
 	}
-	c.Conditions = c.addCondition(condition)
-	return true
 }
 
-// SetFailed Sets the failed condition with the error reason and message
-func (c *ConditionsMeta) SetFailed(reason api.KogitoServiceConditionReason, err error) {
-	condition := Condition{
+// NewFailedCondition ...
+func (c *ConditionsMeta) NewFailedCondition(reason api.KogitoServiceConditionReason, err error) api.ConditionInterface {
+	return Condition{
 		Type:               api.FailedConditionType,
 		Status:             corev1.ConditionFalse,
 		LastTransitionTime: metav1.Now(),
 		Reason:             reason,
 		Message:            err.Error(),
 	}
-	c.Conditions = c.addCondition(condition)
-}
-
-// addCondition adds a condition to the condition array ensuring the max buffer
-func (c *ConditionsMeta) addCondition(condition Condition) []Condition {
-	size := len(c.Conditions) + 1
-	first := 0
-	if size > maxBufferCondition {
-		first = size - maxBufferCondition
-	}
-	return append(c.Conditions, condition)[first:size]
 }
--- a/core/kogitoservice/deployer_test.go
+++ b/core/kogitoservice/deployer_test.go
@@ -80,11 +80,10 @@ func Test_serviceDeployer_DataIndex_InfraNotReady(t *testing.T) {
 	assert.Equal(t, reconcileAfter, reconciliationIntervalAfterInfraError)
 	test.AssertFetchMustExist(t, cli, dataIndex)
 	assert.NotNil(t, dataIndex.GetStatus())
-	assert.Len(t, dataIndex.GetStatus().GetConditions(), 2)
-	for _, condition := range dataIndex.GetStatus().GetConditions() {
-		assert.Equal(t, condition.GetType(), api.FailedConditionType)
-		assert.Equal(t, condition.GetStatus(), corev1.ConditionFalse)
-	}
+	assert.Len(t, dataIndex.GetStatus().GetConditions(), 1)
+	condition := dataIndex.GetStatus().GetConditions()[0]
+	assert.Equal(t, condition.GetType(), api.FailedConditionType)
+	assert.Equal(t, condition.GetStatus(), corev1.ConditionFalse)
 }
 
 func Test_serviceDeployer_DataIndex(t *testing.T) {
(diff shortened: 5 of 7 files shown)


core/framework/condition_collection.go
```diff
@@ -0,0 +1,62 @@
...
+	c.pushCondition(newCondition)
+	return c.conditions
+}
+
+func (c *conditionCollection) removeOldCondition(index int) {
+	c.conditions = append(c.conditions[:index], c.conditions[index+1:]...)
+}
+
+func (c *conditionCollection) pushCondition(newCondition api.ConditionInterface) {
+	if len(c.conditions) >= maxBufferCondition {
```
sutaakar, line 52: What is the purpose of the bugger here, when every condition type is available just once?

Kaitou786:
@vaibhavjainwiz If I went though this correctly we will only have the latest `type` in the status. It means if for failure even if the `KogitoRuntime` have multiple failures we will only display the last one(the latest that the reconciler found). So the developer would first fix the latest issue(the only one he can see in the status), and later he would find that there were other failures as well which needs to be fixed. IMO we can keep the same `type/conditions` in the status list if their `message` is different specially in case of failures.

sutaakar:
Failure and success conditions have different type, as a result both failure and success conditions are shown when application starts correctly:
```
    - lastTransitionTime: '2021-03-11T09:54:05Z'
      message: >-
        imagestreams.image.openshift.io "process-springboot-example" already
        exists
      reason: ReconciliationFailure
      status: 'False'
      type: Failed
    - lastTransitionTime: '2021-03-11T09:54:06Z'
      status: 'True'
      type: Provisioning
    - lastTransitionTime: '2021-03-11T09:54:58Z'
      status: 'True'
      type: Deployed
```
Shouldn't we have same type for error and successful deployment? That way used doesn't have to investigate timestamps to find out whether the application runs successfully or not.

core/framework/condition_collection.go
vaibhavjainwiz: `maxBufferLimit` was already there in previous implementation that why I keep it.
yes you are right we can remove it now.

vaibhavjainwiz: done

ricardozanini (changes requested):
@vaibhavjainwiz this looks OK, but I'd rather move to k8s core API instead with `metav1.Condition`. So we can remove completely our own conditions structure. This must be done for every resource (CRD) we have. Then you can use `metav1.StatusFailure` and ` metav1.ConditionFalse` (and vice-versa) to define the right cause. We still have the `errorFor...` functions to define the correct message and to control the reconciliation logic.

Plus, use `SetConditionsMeta` from `k8s.io/apimachinery/pkg/api/meta` package to control the definition of each condition.

core/framework/condition_collection_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+	"github.com/kiegroup/kogito-cloud-operator/api"
+	"github.com/kiegroup/kogito-cloud-operator/api/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func TestConditionCollection(t *testing.T) {
```
spolti, line 25: what is the purpose of this test? shouldn't int be comparing what has been added instead print it?

core/framework/condition_collection.go
```diff
@@ -0,0 +1,51 @@
...
+func NewConditionCollection(conditions []api.ConditionInterface) ConditionCollection {
+	return &conditionCollection{
+		conditions: conditions,
+	}
+}
+
+func (c *conditionCollection) AddCondition(newCondition api.ConditionInterface) []api.ConditionInterface {
+	for i, condition := range c.conditions {
+		if condition != nil && condition.GetType() == newCondition.GetType() {
+			c.removeOldCondition(i)
```
spolti, line 38: shouldn't be better keep, like, the latest X conditions here?

spolti: maybe set a limit and use FILO?

spolti (changes requested):
nice work, just a few comments.

core/framework/condition_collection.go
vaibhavjainwiz: Previous it was implemented in that manner only. But in jira the expectation is to behave like a collection of the latest status of each different type.

core/framework/condition_collection_test.go
vaibhavjainwiz: I add those print stmt for my quick testing but I forget to replace them with asserts in final PR. I fixed them now.

core/framework/condition_collection.go
ruromero: Each condition should be a snapshot of the latest status. It should be have like a key-value map, not like a history

ricardozanini: Please see: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go#L30

We should use this behavior and remove the `buffer` thing.

ricardozanini: We have access to this function, it's just a matter of updating our Status conditions.

spolti: I see, ok, is this question resolved?

vaibhavjainwiz: As discussed, I had implemented condition as a snapshot of the latest status and use `k8s.io/apimachinery/pkg/api/meta` Condition object

vaibhavjainwiz:
> `SetConditionsMeta` from `k8s.io/apimachinery/pkg/api/meta`

@ricardozanini I didn't found `SetConditionsMeta` in `k8s.io/apimachinery/pkg/api/meta`. Could you please help me to find it.

vaibhavjainwiz:
@ricardozanini I had replaced our custom Condition struct with `metav1.Condition` in all CRDs. There is a catch for KogitoInfra, all crds except kogitoInfra have collection of Conditions inside their status but KogitoInfra only contain single Condition object.

ricardozanini:
> @ricardozanini I had replaced our custom Condition struct with `metav1.Condition` in all CRDs. There is a catch for KogitoInfra, all crds except kogitoInfra have collection of Conditions inside their status but KogitoInfra only contain single Condition object.

Yeah, that's wrong. We should have a collection there as well. :)

Can you change it?

ricardozanini:
> `SetConditionsMeta`

Sure!

https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go#L30

go.mod
```diff
@@ -33,24 +33,24 @@ require (
...
+	k8s.io/api v0.20.4
+	k8s.io/apiextensions-apiserver v0.20.1
+	k8s.io/apimachinery v0.20.4
 	k8s.io/client-go v12.0.0+incompatible
-	k8s.io/kube-openapi v0.0.0-20200410145947-bcb3869e6f29
+	k8s.io/kube-openapi v0.0.0-20201113171705-d219536bb9fd
 	knative.dev/eventing v0.18.0
 	knative.dev/pkg v0.0.0-20200922164940-4bf40ad82aab
-	sigs.k8s.io/controller-runtime v0.6.3
+	sigs.k8s.io/controller-runtime v0.6.1
```
ricardozanini, line 43: Why downgrade here?

vaibhavjainwiz:
> > @ricardozanini I had replaced our custom Condition struct with `metav1.Condition` in all CRDs. There is a catch for KogitoInfra, all crds except kogitoInfra have collection of Conditions inside their status but KogitoInfra only contain single Condition object.
> 
> Yeah, that's wrong. We should have a collection there as well. :)
> 
> Can you change it?

KogitoInfra doesn't provision any k8s object. It just connect with existing resources. Currently we are only maintaing two states(success/failure) on bases if KogitoInfra could be able to connect with provided resource and that enough imformation for user. I not sure if KogitoInfra requires collection of Conditions. Wdyt?

go.mod
```diff
@@ -33,24 +33,24 @@ require (
 	golang.org/x/tools v0.0.0-20200916195026-c9a70fc28ce3 // indirect
 	gopkg.in/src-d/go-git.v4 v4.13.1
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.18.8
-	k8s.io/apiextensions-apiserver v0.18.6
-	k8s.io/apimachinery v0.18.8
+	k8s.io/api v0.20.4
+	k8s.io/apiextensions-apiserver v0.20.1
+	k8s.io/apimachinery v0.20.4
```
ricardozanini, line 38: Hmmm this is tricky.. we might lose compatibility with older kube versions. Please, just make sure that we are safe:
https://github.com/kubernetes/client-go#compatibility-matrix

ricardozanini (commented):
Just small comments. Great work man!

ricardozanini:
> > > @ricardozanini I had replaced our custom Condition struct with `metav1.Condition` in all CRDs. There is a catch for KogitoInfra, all crds except kogitoInfra have collection of Conditions inside their status but KogitoInfra only contain single Condition object.
> > 
> > 
> > Yeah, that's wrong. We should have a collection there as well. :)
> > Can you change it?
> 
> KogitoInfra doesn't provision any k8s object. It just connect with existing resources. Currently we are only maintaing two states(success/failure) on bases if KogitoInfra could be able to connect with provided resource and that enough imformation for user. I not sure if KogitoInfra requires collection of Conditions. Wdyt?

I think we might have to implement the collection since it's a "pattern" for Kubernetes resources: https://book-v1.book.kubebuilder.io/basics/status_subresource.html

We need to keep a history of conditions and the last time they transitioned.

api/conditions.go
```diff
@@ -49,20 +48,9 @@ const (
...
 // ConditionMetaInterface defines the base information for kogito services conditions
 type ConditionMetaInterface interface {
-	SetDeployed() bool
-	SetProvisioning() bool
-	SetFailed(reason KogitoServiceConditionReason, err error)
-	GetConditions() []ConditionInterface
-	SetConditions(conditions []ConditionInterface)
+	GetConditions() []metav1.Condition
+	AddCondition(condition metav1.Condition)
+	RemoveCondition(index int)
```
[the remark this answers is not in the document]
vaibhavjainwiz: I had removed `AddCondition` and `RemoveCondition` methods but we still requires `SetConditions` method

core/kogitobuild/status.go
```diff
@@ -64,72 +45,145 @@ func NewStatusHandler(context *operator.Context) StatusHandler {
 }
 
 func (s *statusHandler) HandleStatusChange(instance api.KogitoBuildInterface, err error) {
```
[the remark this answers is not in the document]
vaibhavjainwiz: done

vaibhavjainwiz:
> > > > @ricardozanini I had replaced our custom Condition struct with `metav1.Condition` in all CRDs. There is a catch for KogitoInfra, all crds except kogitoInfra have collection of Conditions inside their status but KogitoInfra only contain single Condition object.
> > > 
> > > 
> > > Yeah, that's wrong. We should have a collection there as well. :)
> > > Can you change it?
> > 
> > 
> > KogitoInfra doesn't provision any k8s object. It just connect with existing resources. Currently we are only maintaing two states(success/failure) on bases if KogitoInfra could be able to connect with provided resource and that enough imformation for user. I not sure if KogitoInfra requires collection of Conditions. Wdyt?
> 
> I think we might have to implement the collection since it's a "pattern" for Kubernetes resources: https://book-v1.book.kubebuilder.io/basics/status_subresource.html
> 
> We need to keep a history of conditions and the last time they transitioned.

@ricardozanini I had implemented collection of condition for KogitoInfra. But regarding your comment `keep a history of conditions and the last time they transitioned`, It will not be there.  As describe by @ruromero `Each condition should be a snapshot of the latest status. It should be have like a key-value map, not like a history`. 
For example in KogitoRuntime, now we are maintaining three type of condition `Failed`, `Provisioning` & `Deployed`. On each reconcilation cycle, we found out whats the current state of these 3 condition and set value of all 3 condition every time. So no history details.

go.mod
vaibhavjainwiz: undo it. To setup complatablity I had made changes in many pkg version. I might miss to reset it

core/infrastructure/deployment.go
```diff
@@ -78,3 +79,12 @@ func (d *deploymentHandler) IsDeploymentAvailable(key types.NamespacedName) (boo
 	}
 	return true, nil
 }
+
+func (d *deploymentHandler) FetchReadyReplicas(key types.NamespacedName) (int32, error) {
+	deployment, err := d.MustFetchDeployment(key)
+	if err != nil {
+		return 0, err
+	}
+	return deployment.Status.Replicas, nil
```
sutaakar, line 88:
```suggestion
	return deployment.Status.AvailableReplicas, nil
```

core/kogitoinfra/status_test.go
```diff
@@ -0,0 +1,57 @@
...
+		Scheme: meta.GetRegisteredSchema(),
+	}
+	statusHandler := NewStatusHandler(context)
+	err := errors.New("error")
+	statusHandler.UpdateBaseStatus(instance, &err)
+	test.AssertFetchMustExist(t, cli, instance)
+	conditions := *instance.Status.Conditions
+	assert.Equal(t, 2, len(conditions))
+	assert.Equal(t, string(api.KogitoInfraFailure), conditions[0].Type)
+	assert.Equal(t, string(api.KogitoInfraSuccess), conditions[1].Type)
```
sutaakar, line 56: Please check also status to make sure what are the values of those types.

sutaakar:
The code looks ok, need to test it....

core/infrastructure/deployment.go
vaibhavjainwiz: done

core/kogitoinfra/status_test.go
vaibhavjainwiz: done

ricardozanini:
> @ricardozanini I had implemented collection of condition for KogitoInfra. But regarding your comment `keep a history of conditions and the last time they transitioned`, It will not be there. As describe by @ruromero `Each condition should be a snapshot of the latest status. It should be have like a key-value map, not like a history`.

I think I wasn't clear in my last comment. This pattern is enforced by the `SetConditions` function, whereas we keep the **latest condition** of a given **type**, not all. My bad if I wasn't clear.

See [reference](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties):

> Objects may report multiple conditions, and new types of conditions may be added in the future or by 3rd party controllers. Therefore, conditions are represented using a list/slice of objects, where each condition has a similar structure. This collection should be treated as a map with a key of type.

go.mod
vaibhavjainwiz: I had tested this functionality on k8s v1.20.0. Do I need to test it with some other k8s version?

api/kogitoinfra_types.go
```diff
@@ -24,10 +24,10 @@ import (
...
 
 const (
-	// SuccessInfraConditionType ...
-	SuccessInfraConditionType KogitoInfraConditionType = "Success"
-	// FailureInfraConditionType ...
-	FailureInfraConditionType KogitoInfraConditionType = "Failure"
+	// KogitoInfraSuccess ...
+	KogitoInfraSuccess KogitoInfraConditionType = "Success"
+	// KogitoInfraFailure ...
+	KogitoInfraFailure KogitoInfraConditionType = "Failure"
```
ricardozanini, line 30: We don't need to replicate these constants here. Please use this instead:
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L719-L722

go.mod
ricardozanini: I think we are fine up to 1.18. @sutaakar, the OCP we run our BDD is based on which k8s version?

ricardozanini (approved):
One nitpick, apart from that, looks really good. Many thanks!

go.mod
sutaakar: good question, @Kaitou786 could know that

api/kogitoinfra_types.go
vaibhavjainwiz: @ricardozanini The link that you had shared above are status but we are using `KogitoInfraSuccess` & `KogitoInfraFailure` as type. In Condition struct we have two fields `type` and `status`. Every type could have its status `success` or `failure`. 
To me as well these type are not looks right, type should be something like `deployed` or `provisioning` which could have their status as `success` or `false`.

ricardozanini: Agreed. We should change this naming then :/

sutaakar:
A nitpick, I have noticed that `lastTransitionTime` changes in every reconciliation cycle, it should be change only when there is a transition in the condition.

api/kogitoservices_types.go
```diff
@@ -21,6 +21,46 @@ import (
...
+	// InternalServiceNotReachable ...
+	InternalServiceNotReachable KogitoServiceConditionReason = "InternalServiceNotReachable"
+	// SuccessfulDeployedReason ...
+	SuccessfulDeployedReason KogitoServiceConditionReason = "AtLeastOnePodAvailable"
+	// FailedDeployedReason ...
+	FailedDeployedReason KogitoServiceConditionReason = "NoPodAvailable"
+	// SuccessfulProvisioningReason ...
+	SuccessfulProvisioningReason KogitoServiceConditionReason = "RequestedReplicasNotEqualToAvailableReplicas"
+	// FailedProvisioningReason ...
+	FailedProvisioningReason KogitoServiceConditionReason = "RequestedReplicasEqualToAvailableReplicas"
```
sutaakar, line 59:
```suggestion
	SuccessfulProvisioningReason KogitoServiceConditionReason = "RequestedReplicasEqualToAvailableReplicas"
	// FailedProvisioningReason ...
	FailedProvisioningReason KogitoServiceConditionReason = "RequestedReplicasNotEqualToAvailableReplicas"
```

sutaakar, line 59: IMHO the reason value should be something more abstract. Provisioning can be set to true also in case there is some  error during reconciliation, while pods are already up and running.
Maybe something like:
```suggestion
	// SuccessfulProvisioningReason ...
	SuccessfulProvisioningReason KogitoServiceConditionReason = "ProvisioningInProgress"
	// FinishedProvisioningReason ...
	FinishedProvisioningReason KogitoServiceConditionReason = "RequestedReplicasEqualToAvailableReplicas"
	// FailedProvisioningReason ...
	FailedProvisioningReason KogitoServiceConditionReason = "ProvisioningFailure"
```

sutaakar:
@vaibhavjainwiz In case of some error the KogitoInfra shows both success and failure types:
```
    - lastTransitionTime: '2021-03-18T08:38:21Z'
      message: >-
        Infinispan resource(kogito-infinispan3) not found in namespace
        ksuta-test
      reason: ResourceNotFound
      status: 'True'
      type: Failure
    - lastTransitionTime: '2021-03-18T08:38:21Z'
      message: ''
      reason: ResourceNotFound
      status: 'False'
      type: Success
```
Does it have sense to show both? What about just using Success type with possible message in case of an error?

core/kogitobuild/status.go
```diff
@@ -64,114 +44,136 @@ func NewStatusHandler(context *operator.Context) StatusHandler {
...
-		secondTransitionTime := instance.GetStatus().GetConditions()[j].GetLastTransitionTime()
-		return firstTransitionTime.Before(&secondTransitionTime)
-	})
+// newSuccessfulCondition ...
+func (s *statusHandler) newSuccessfulCondition(status metav1.ConditionStatus) metav1.Condition {
+	return metav1.Condition{
+		Type:               string(api.KogitoBuildSuccessful),
+		Status:             status,
+		LastTransitionTime: metav1.Now(),
+		Reason:             string(status),
```
sutaakar, line 66: Results in:
```
    - lastTransitionTime: '2021-03-18T08:46:42Z'
      message: ''
      reason: 'False'
      status: 'False'
      type: Successful
```
IMHO we should provide some meaningful reason like BuildInProgress (for the case when build is still in progress), possibly some error reason too for the case of a build error.

core/kogitobuild/status.go
```diff
@@ -64,114 +44,136 @@ func NewStatusHandler(context *operator.Context) StatusHandler {
...
-			Reason:             api.OperatorFailureReason,
-			Message:            err.Error(),
-		})
+// newRunningCondition ...
+func (s *statusHandler) newRunningCondition(status metav1.ConditionStatus) metav1.Condition {
+	return metav1.Condition{
+		Type:               string(api.KogitoBuildRunning),
+		Status:             status,
+		LastTransitionTime: metav1.Now(),
+		Reason:             string(status),
```
sutaakar, line 76: Same here.

sutaakar:
Apart from the points above (and the nitpick `lastTransitionTime` being refreshed in every reconciliation) looks good to me.

go.mod
Kaitou786: Currently we use OCP 4.6.x for our BDD tests which is based on kubernetes 1.19 as stated in the [docs](https://docs.openshift.com/container-platform/4.6/release_notes/ocp-4-6-release-notes.html#ocp-4-6-about-this-release)

spolti (approved):

go.mod
vaibhavjainwiz: I had tested these changes on kubernetes 1.19. They are working fine.

PR #792, opened 2021-03-11
[KOGITO-4453] Fix overwriting propertiesConfigMap

Kevin-Mok:
See: https://issues.redhat.com/browse/KOGITO-4453

## Cause of Bug
The cause of this bug is that the `application.properties` provided in the propertiesConfigMap was not being added to the `appProps` map in `createRequiredResources()`. So when the ConfigMap is created in `createRequiredResources()`, it would not have any of the `application.properties` provided and delete the entire propertiesConfigMap when comparing the deltas of what is created and deployed in the deploy loop.

## Fix
I fixed this bug by adding the `application.properties` from the propertiesConfigMap to the `appProps` map in `createRequiredResources()`.  This has to be done outside of the `if` block that checks if there is a resolvable image or deployment because I came across the case when there wasn't a resolvable image/deployment ready. In this case, the `application.properties` would not be added to the `appProps` map, and the bug would occur again.

## Testing
I tested my code with a propertiesConfigMap added to a KogitoRuntime/KogitoSupportingService:
```yaml
kind: KogitoRuntime
metadata:
  name: qe
spec:
  replicas: 1
  image: quay.io/kmok/process-quarkus-example
  propertiesConfigMap: qe-cm
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: qe-cm
data:
  application.properties: |
    kafka.sals.mechanism=PLAIN
    kafka.security.protocol=SASL_SSL
```

```yaml
kind: KogitoRuntime
metadata:
  name: tq
spec:
  replicas: 1
  image: quay.io/kmok/process-timer-quarkus
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: js-cm
data:
  application.properties: |
    kafka.sals.mechanism=PLAIN
    kafka.security.protocol=SASL_SSL
---
kind: KogitoSupportingService
metadata:
  name: jobs-service
spec:
  serviceType: JobsService
  replicas: 1
  propertiesConfigMap: js-cm
```

I also added a unit test to check if the propertiesConfigMap shows up in the created resources even if there if no resolvable image yet.

## Requirements

### DIFF
--- a/core/kogitoservice/deployer_resources.go
+++ b/core/kogitoservice/deployer_resources.go
@@ -15,6 +15,7 @@
 package kogitoservice
 
 import (
+	"fmt"
 	"github.com/RHsyseng/operator-utils/pkg/resource"
 	"github.com/RHsyseng/operator-utils/pkg/resource/compare"
 	"github.com/kiegroup/kogito-cloud-operator/api"
@@ -27,6 +28,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"reflect"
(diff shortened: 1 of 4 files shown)


ricardozanini (approved):
Great job and great PR description, very well explained. It saved me some time to understand what has been done.

spolti (approved):
well done Kevin.

core/kogitoservice/deployer_resources.go
```diff
@@ -44,6 +45,28 @@ func (s *serviceDeployer) createRequiredResources() (resources map[reflect.Type]
...
+		s.Log.Debug("custom app properties are provided in custom properties ConfigMap", "PropertiesConfigMap", s.instance.GetSpec().GetPropertiesConfigMap())
+		propertiesConfigMap := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: s.instance.GetNamespace(),
+				Name:      s.instance.GetSpec().GetPropertiesConfigMap(),
+			},
+		}
+		if exists, err := kubernetes.ResourceC(s.Client).Fetch(propertiesConfigMap); err != nil {
+			return resources, err
+		} else if exists {
```
sutaakar, line 60: What should we do in case user actually defines config map but it doesn't exist?

I am thinking whether we should rather consider this as an error. Such issue could be caused for example by having a typo in the config map name. 
In case operator act like there is an empty config map the user can be mislead in thinking that everything is fine.

ricardozanini: Good question. I think we should fail, hence updating the statusCondition with a meaningful message so users can easily fix it.

Kevin-Mok: I changed it to throw an error when it can't find the propertiesConfigMap now and added a unit test for it. It'll continually throw the error in the operator log and won't deploy the pod until its fixed, so it's an obvious error for the user now.

ricardozanini:
@sutaakar, are we good to merge this one?

sutaakar:
Testing...

sutaakar (approved):
Looks good to me.

PR #809, opened 2021-03-22
[KOGITO-4731] - Conditions should reflect latest evaluated status - R…

vaibhavjainwiz:
…eview comments

Jira issue : https://issues.redhat.com/browse/KOGITO-4731


sutaakar:
I see one minor thing, in case there is a reconciliation error first and then it is successfully reconciled then the `Failed` condition is still present with value true:
```
  conditions:
    - lastTransitionTime: '2021-03-22T12:25:39Z'
      message: imagestreams.image.openshift.io "process-quarkus-example" already exists
      reason: ReconciliationFailure
      status: 'True'
      type: Failed
    - lastTransitionTime: '2021-03-22T12:28:14Z'
      message: ''
      reason: RequestedReplicasEqualToAvailableReplicas
      status: 'False'
      type: Provisioning
    - lastTransitionTime: '2021-03-22T12:28:14Z'
      message: ''
      reason: AtLeastOnePodAvailable
      status: 'True'
      type: Deployed
```
I would expect that when application is successfully deployed without reconciliation error then `Failed` condition is removed, or possibly set to false.

sutaakar:
Otherwise looks ok.
It seems that we don't watch builds so when the build finishes then the conditions are not refreshed (at least not immediately).
It would be nice to refresh the conditions once then build finishes, but that may be out of scope for this PR.

radtriste (approved):

Kaitou786 (approved):

vaibhavjainwiz:
> I see one minor thing, in case there is a reconciliation error first and then it is successfully reconciled then the `Failed` condition is still present with value true:
> 
> ```
>   conditions:
>     - lastTransitionTime: '2021-03-22T12:25:39Z'
>       message: imagestreams.image.openshift.io "process-quarkus-example" already exists
>       reason: ReconciliationFailure
>       status: 'True'
>       type: Failed
>     - lastTransitionTime: '2021-03-22T12:28:14Z'
>       message: ''
>       reason: RequestedReplicasEqualToAvailableReplicas
>       status: 'False'
>       type: Provisioning
>     - lastTransitionTime: '2021-03-22T12:28:14Z'
>       message: ''
>       reason: AtLeastOnePodAvailable
>       status: 'True'
>       type: Deployed
> ```
> 
> I would expect that when application is successfully deployed without reconciliation error then `Failed` condition is removed, or possibly set to false.

Thanks @sutaakar for catching this scenerio, I completely miss that. Now I had fixed it and added testcases to test this scenerio

ricardozanini:
> I would expect that when application is successfully deployed without reconciliation error then Failed condition is removed, or possibly set to false.

@sutaakar set to `false` should be the right approach, IMO. Since we don't want to lose track of the condition's history.

ricardozanini (approved):
Nice! Seems much better now.

ricardozanini:
>  It would be nice to refresh the conditions once then build finishes, but that may be out of scope for this PR.

I agree, maybe a follow up JIRA then?

vaibhavjainwiz:
> > I would expect that when application is successfully deployed without reconciliation error then Failed condition is removed, or possibly set to false.
> 
> @sutaakar set to `false` should be the right approach, IMO. Since we don't want to lose track of the condition's history.

@ricardozanini If we keep Failed condition in that case then what should be its Reason and Message.

ricardozanini:
> > > I would expect that when application is successfully deployed without reconciliation error then Failed condition is removed, or possibly set to false.
> > 
> > 
> > @sutaakar set to `false` should be the right approach, IMO. Since we don't want to lose track of the condition's history.
> 
> @ricardozanini If we keep Failed condition in that case then what should be its Reason and Message.

You only have to keep the `status` to false. I think that when we always have a failure, `status` would false. :thinking: 
Or am I missing something?

vaibhavjainwiz:
> > > > I would expect that when application is successfully deployed without reconciliation error then Failed condition is removed, or possibly set to false.
> > > 
> > > 
> > > @sutaakar set to `false` should be the right approach, IMO. Since we don't want to lose track of the condition's history.
> > 
> > 
> > @ricardozanini If we keep Failed condition in that case then what should be its Reason and Message.
> 
> You only have to keep the `status` to false. I think that when we always have a failure, `status` would false. 
> Or am I missing something?

`Failure` is a condition type in our case. We put `Failure` condition with status `true` when we receive `error` in `HandleStatusUpdate` function and reason/message are derived form that `error`.
Now if need to keep `Failure` condition in case of success as well then we would mark its status as `false`. Reason is mandatory field in condition struct and I not sure what I have to put in reason when `Failure` condition is `false`. :(

ricardozanini:
> `Failure` is a condition type in our case. We put `Failure` condition with status `true` when we receive `error` in `HandleStatusUpdate` function and reason/message are derived form that `error`.
> Now if need to keep `Failure` condition in case of success as well then we would mark its status as `false`. Reason is mandatory field in condition struct and I not sure what I have to put in reason when `Failure` condition is `false`. :(

Hmmm, I think the point raised by @sutaakar is that we need to modify the `status` to `false` on a `Failure` condition only when we recovered from a previous failure. Can be tricky 'cause you will need to read from the conditions array and only update if it's there. I believe clients will fetch for our status conditions array and look for the `Status: true` to understand the latest condition we are. If we recover from a failure state, but the status keeps indicating `true`, for the caller we will be always there.

@sutaakar is that right?

sutaakar:
Right

vaibhavjainwiz:
@ricardozanini @sutaakar I had incorporated your review comment to set Failed condition status to false when reconcilation recoved from failure

ricardozanini:
@sutaakar, I think we're fine. Do you mind running a final check before we merge this one?

spolti (approved):

sutaakar (approved):
Looks good to me.

PR #813, opened 2021-03-23
[KOGITO-4606] - Remove the hardcoded references to the default images…

vaibhavjainwiz:
… name

Jira issue : https://issues.redhat.com/browse/KOGITO-4606

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -34,4 +34,14 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.annotations['olm.targetNamespaces']
+          - name: builder.image
+            value: kogito-builder
+          - name: runtime.image
+            value: kogito-runtime-jvm
+          - name: runtime.native.image
+            value: kogito-runtime-native
+          - name: image.namespace
+            value: kiegroup
+          - name: image.registry
+            value: quay.io
       terminationGracePeriodSeconds: 10
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -34,14 +34,14 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.annotations['olm.targetNamespaces']
-          - name: builder.image
+          - name: BUILDER_IMAGE
             value: kogito-builder
-          - name: runtime.image
+          - name: RUNTIME_IMAGE
             value: kogito-runtime-jvm
-          - name: runtime.native.image
+          - name: RUNTIME_NATIVE_IMAGE
             value: kogito-runtime-native
-          - name: image.namespace
+          - name: IMAGE_NAMESPACE
             value: kiegroup
-          - name: image.registry
+          - name: IMAGE_REGISTRY
             value: quay.io
       terminationGracePeriodSeconds: 10
--- a/api/v1beta1/zz_generated.openapi.go
+++ b/api/v1beta1/zz_generated.openapi.go
@@ -60,8 +60,9 @@ func schema_kiegroup_kogito_operator_api_v1beta1_Builds(ref common.ReferenceCall
 							Items: &spec.SchemaOrArray{
 								Schema: &spec.Schema{
 									SchemaProps: spec.SchemaProps{
-										Type:   []string{"string"},
-										Format: "",
+										Default: "",
+										Type:    []string{"string"},
+										Format:  "",
 									},
 								},
 							},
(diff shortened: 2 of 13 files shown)


config/manager/manager.yaml
```diff
@@ -34,4 +34,14 @@ spec:
...
+          - name: builder.image
+            value: kogito-builder
+          - name: runtime.image
+            value: kogito-runtime-jvm
+          - name: runtime.native.image
+            value: kogito-runtime-native
+          - name: image.namespace
+            value: kiegroup
+          - name: image.registry
+            value: quay.io
```
ricardozanini, line 46: Please `strings.upperCase` to these vars

ricardozanini (changes requested):

config/manager/manager.yaml
spolti: nad maybe _ instead . ?

spolti (approved):

config/manager/manager.yaml
vaibhavjainwiz: done

ricardozanini (commented):
What's the use case? For us to inform the default image in the product manifests?

vaibhavjainwiz:
> What's the use case? For us to inform the default image in the product manifests?

For KogitoBuild if user don't provide `builderImage` and `runtimeImage` reference, then we pick default images define in Operator. Previously these image name are hardcoded in code which is shared with rhpam-kogito-operator. Now we had make them configurable so that product could define its own builder/runtime image.

sutaakar (approved):

Kaitou786 (approved):

config/manager/manager.yaml
```diff
@@ -34,5 +34,15 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.annotations['olm.targetNamespaces']
+          - name: BUILDER_IMAGE
```
radtriste, line 37: do we really need those values here as they are the default ones ?
Or are they for documentation purpose ?

radtriste (approved):
lgtm

one minor comment as we have default values in code and default env var defined in yaml file.

btw is there a way to update those values in OLM ?

config/manager/manager.yaml
ricardozanini: > btw is there a way to update those values in OLM ?

Not sure. I imagine we will hard code the default image names in the product CSV.

ricardozanini:
> btw is there a way to update those values in OLM ?

Not sure. I imagine we will hard code the default image names in the product CSV.

config/manager/manager.yaml
vaibhavjainwiz: for community operator, these values are not required because these env var default values are defines as per community operator requirement. So if remove these env var from deployment in community operator, then it will pick its default values and works fine. But these value will mandatory to define product operator because default values are invalid in case of product operator.

For now, its work as sample for How we should define these value in product operator.

radtriste: ok fine for em

radtriste:
> > btw is there a way to update those values in OLM ?
> 
> Not sure. I imagine we will hard code the default image names in the product CSV.

So that means the user would need to update the deployment after the installation. Would that work ?
Anyway user can still setup the images in Kogito resources but if we provide a way to setup defaults as env var, user may request a way to update it (and that could be much easier for them ...)
Bot for this PR of course but we need to think about that ;)

radtriste:
@ricardozanini If you are good with latest changes, I guess this can be merged ;)

ricardozanini (approved):

ricardozanini:
> > > btw is there a way to update those values in OLM ?
> > 
> > 
> > Not sure. I imagine we will hard code the default image names in the product CSV.
> 
> So that means the user would need to update the deployment after the installation. Would that work ?
> Anyway user can still setup the images in Kogito resources but if we provide a way to setup defaults as env var, user may request a way to update it (and that could be much easier for them ...)
> Not for this PR of course but we need to think about that ;)

You're right. Can you open a follow-up JIRA, please?

radtriste:
> > > > btw is there a way to update those values in OLM ?
> > > 
> > > 
> > > Not sure. I imagine we will hard code the default image names in the product CSV.
> > 
> > 
> > So that means the user would need to update the deployment after the installation. Would that work ?
> > Anyway user can still setup the images in Kogito resources but if we provide a way to setup defaults as env var, user may request a way to update it (and that could be much easier for them ...)
> > Not for this PR of course but we need to think about that ;)
> 
> You're right. Can you open a follow-up JIRA, please?

Created https://issues.redhat.com/browse/KOGITO-4814

PR #822, opened 2021-03-29
[KOGITO-4781] Fix deployment constantly updating

Kevin-Mok:
See: https://issues.redhat.com/browse/KOGITO-4781

## Description
This PR sets default values for probes so that values not specified for probes in the YAML have the default values in the requested resources. This makes it so that it matches the deployed resources (Kubernetes was automatically adding the default values for unspecified fields) and prevents the operator from continually updating the deployment when it sees a discrepancy between the requested and deployed resources.

## Testing
I updated the probe unit tests to check that the default values are being added when they are not specified. I also deployed the following YAML on both `master` and this branch:
```yaml
kind: KogitoRuntime
metadata:
  name: qe
spec:
  replicas: 1
  image: quay.io/kmok/process-quarkus-example
  probes:
    livenessProbe:
      httpGet:
        path: /q/health/live
        port: 8080
```
Where the operator would continually run in `master` to update the deployment but not in my branch.

## Requirements

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,5 +12,6 @@
 - [KOGITO-4729](https://issues.redhat.com/browse/KOGITO-4729) - Upgrade from Operator 1.3.0 to 1.4.0 fails with reference to Kafka Topics
 - [KOGITO-4905](https://issues.redhat.com/browse/KOGITO-4905) - COPY line in springboot.Dockerfile doesn't work
 - [KOGITO-4894](https://issues.redhat.com/browse/KOGITO-4894) - Remove `image` default value in KogitoRuntime CR (problem in OCP console)
+- [KOGITO-4781](https://issues.redhat.com/browse/KOGITO-4781) - Deployment constantly updates when deployed with custom probes
 
 ## Known Issues
--- a/core/kogitoservice/probe.go
+++ b/core/kogitoservice/probe.go
@@ -40,6 +40,13 @@ type healthCheckProbe struct {
 	liveness  *corev1.Probe
 }
 
+var defaultProbeValues = corev1.Probe{
+	TimeoutSeconds:   int32(1),
+	PeriodSeconds:    int32(10),
+	SuccessThreshold: int32(1),
+	FailureThreshold: int32(3),
+}
+
 // getProbeForKogitoService gets the appropriate liveness (index 0) and readiness (index 1) probes based on the given service definition
 func getProbeForKogitoService(serviceDefinition ServiceDefinition, service api.KogitoService) healthCheckProbe {
 	switch serviceDefinition.HealthCheckProbe {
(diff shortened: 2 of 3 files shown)


sutaakar:
@Kevin-Mok I have noticed one corner case. When I omit `path` and specify only the port then the endless reconciliation is still present (only port is a mandatory field for `httpGet`).
Can you also set default path in case it is not specified?

ricardozanini (approved):

spolti (approved):

Kevin-Mok:
@sutaakar The endless reconciliation is from Kubernetes setting the omitted path to be `/` which would cause the probes to fail as well. I set the default paths to the default Quarkus health ones and updated the unit tests to check for them.

sutaakar:
@Kevin-Mok It is a bit better now.
But, in case I use tcpSocket for Quarkus probe, then `nil pointer dereference` is thrown.
Example CR:
```
kind: KogitoRuntime
metadata:
  name: qe
spec:
  replicas: 1
  image: quay.io/kmok/process-quarkus-example
  probes:
    livenessProbe:
      tcpSocket:
        port: 8080
```
causing:
```
E0331 08:01:18.748426       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 499 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1965ae0, 0x2d0afe0)
	/go/pkg/mod/k8s.io/apimachinery@v0.20.4/pkg/util/runtime/runtime.go:74 +0x92
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.20.4/pkg/util/runtime/runtime.go:48 +0x82
panic(0x1965ae0, 0x2d0afe0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/kiegroup/kogito-operator/core/kogitoservice.setDefaultHTTPGetValues(...)
```

Kevin-Mok:
> Adding of default probe values should be ideally independent on runtime type (so proper defaults are applied to HTTPGet or TCPSocket regardless of runtime type).

This is a good idea. I hadn't even considered applying TCPSocket for Quarkus or HTTPGet for Spring Boot, but it makes sense to deal with these cases. 

> Would you like to address it in this PR or rather as a separate JIRA?

I can address it in this PR.

spolti:
@sutaakar @Kevin-Mok is something missing on this PR or is it ready to merge?

Kevin-Mok:
@spolti I'm working on adding this:
> Adding of default probe values should be ideally independent on runtime type (so proper defaults are applied to HTTPGet or TCPSocket regardless of runtime type).

spolti:
@Kevin-Mok ok, thanks.

ricardozanini:
@Kevin-Mok let's rebase and merge it after CI run.

Kevin-Mok:
> Would you like to address it in this PR or rather as a separate JIRA?

After thinking about the implementation of this some more, I think it would be better to address this in a separate PR. I opened a ticket for this [here](https://issues.redhat.com/browse/KOGITO-4898).

ricardozanini:
@sutaakar do you mind verifying it?

sutaakar (approved):

Kevin-Mok:
This is ready to be merged.

spolti (approved):

spolti:
@ricardozanini do you want to take a final look?

ricardozanini (approved):

PR #840, opened 2021-04-12
KOGITO-4639 Setup BDD tests coverage

radtriste:
https://issues.redhat.com/browse/KOGITO-4639

- build of a new profiling image (with `got test`) to be able to catch the coverage
- kustomization changed a bit to include the construction of profiling files
- need to have a specific data-access pod to get the coverage as the cover file will be created only after operator is stopped. so using a pvc to store the coverage and to retrieve it with data-access pod
- profiling is just an option of BDD tests (and not default)
- profiling is on PR. nightly will do the testing with the correct operator built.

- [ ] You have read the [contributors' guide](https://github.com/kiegroup/kogito-operator/blob/master/README.md#contributing-to-the-kogito-operator)


radtriste:
thinking about having a full BDD tests nightly run with profiling (but not as part of the nightly process) to have a full coverage idea

sutaakar:
@radtriste Wouldn't it have more sense to set ReadWriteMany for PVC and deploy the data access pod at the same time as deploying Kogito operator? That way we remove the risk of PVC being recycled in the time between deletion of Kogito operator and creating data access pod.
Otherwise great work.

sutaakar (approved):

radtriste:
jenkins test

Makefile
```diff
@@ -108,7 +122,7 @@ ifeq (, $(shell which kustomize))
 	KUSTOMIZE_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$KUSTOMIZE_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/kustomize/kustomize/v3@v3.5.4 ;\
+	go get sigs.k8s.io/kustomize/kustomize/v4@v4.0.5 ;\
```
ricardozanini, line 125: Have you tested the `make before-pr` with this new kustomize version? I'm assuming it's working, but they are known to break things between releases. :(

ricardozanini (commented):
It seems really cool! I'm just worried about the kustomize changes. The SDK created the `default` profile. We might run into problems when upgrading the SDK version.

Makefile
radtriste: yeah ... did this change because it is the one used in GHAction PR check ... will run the `before-pr` to be sure

radtriste:
the `default` is still existing, I just abstracted a bit more to be able to reuse. but indeed, let's see when we update. anyway we could still put again in `default` if needed

radtriste:
> @radtriste Wouldn't it have more sense to set ReadWriteMany for PVC and deploy the data access pod at the same time as deploying Kogito operator? That way we remove the risk of PVC being recycled in the time between deletion of Kogito operator and creating data access pod.
> Otherwise great work.

As discussed, AWS cluster storageclass does not support the `ReadWriteMany`:
```
Failed to provision volume with StorageClass "gp2": invalid AccessModes [ReadWriteMany]: only AccessModes [ReadWriteOnce] are supported
```
so will stay with `ReadWriteOnce` and have the `data-access` pod started after.
anyway, we only delete the Kogito operator deployment and not the pvc so we should still have the data in pvc remaining when we start the `data-access` pod, there should not be any recycling

Kevin-Mok:
> /jenkins test

It seems that your PR is having [this error](https://github.com/kiegroup/kogito-operator/pull/841#issuecomment-821730401) as well even after the weekend, and I reran my Jenkins test twice with the same issue. Any idea what could be causing it?

radtriste:
ok looks like the send to codecov from RHBA jenkins is having some struggle ... maybe due to Openstack network.
From AWS Jenkins (PR check), upload command of results has no problem
From RHBA Jenkins (Profiling nightly), upload command is finishing with `400 Bad request`.
I guess there is a network issue ... Created https://issues.redhat.com/browse/KOGITO-4992 to investigate this. Otherwise, the rest is good

sutaakar (approved):
Looks good to me.

vaibhavjainwiz (approved):

ricardozanini (approved):

PR #841, opened 2021-04-12
[KOGITO-4898] Set probe defaults based on YAML action values

Kevin-Mok:
See: https://issues.redhat.com/browse/KOGITO-4898

## Description
This PR changes `probe.go` to be more runtime agnostic and combines all the previous probe getters into a single catch-all `getProbe` function. The `getProbe` function returns the correct probe based on the runtime and probe type. If the handler is empty (user hasn't specified anything for it), then the returned probe will be the defaults for the specific runtime.  If the handler is not empty and is HTTP, then the path and scheme will be set if they aren't already by the user.  If the handler is not empty and is TCP, then nothing in the handler needs to be set by us since the port is required in the YAML. Finally, the probe timer values are set in all cases to not cause reconciliation problems. So in all, the runtime is only considered when the user hasn't specified anything for it, and so we use the default values for it.

## Testing
I wrote unit tests and tested all the various combinations with Quarkus/Spring Boot and HTTP/TCP on OpenShift.

## Requirements

### DIFF
--- a/core/kogitoservice/probe.go
+++ b/core/kogitoservice/probe.go
@@ -21,24 +21,30 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// HealthCheckProbeType defines the supported probes for the ServiceDefinition
-type HealthCheckProbeType string
+// ProbeType defines the types of probes supported by K8s
+type ProbeType string
 
 const (
-	// QuarkusHealthCheckProbe probe implemented with Quarkus Microprofile Health. See: https://quarkus.io/guides/microprofile-health.
-	// the operator will set the probe to the default path /health/live and /health/ready for liveness and readiness probes, respectively.
-	QuarkusHealthCheckProbe HealthCheckProbeType = "quarkus"
-	// TCPHealthCheckProbe default health check probe that binds to port 8080
-	TCPHealthCheckProbe HealthCheckProbeType = "TCP"
+	livenessProbeType  ProbeType = "liveness"
+	readinessProbeType ProbeType = "readiness"
+	startupProbeType   ProbeType = "startup"
 
 	quarkusProbeLivenessPath  = "/q/health/live"
 	quarkusProbeReadinessPath = "/q/health/ready"
+
+	springBootProbeLivenessPath  = "/actuator/health/liveness"
+	springBootProbeReadinessPath = "/actuator/health/readiness"
 )
 
 type healthCheckProbe struct {
 	readiness *corev1.Probe
 	liveness  *corev1.Probe
 	startup   *corev1.Probe
+}
+
+var defaultHTTPGetAction = corev1.HTTPGetAction{
+	Port:   intstr.IntOrString{IntVal: int32(framework.DefaultExposedPort)},
+	Scheme: corev1.URISchemeHTTP,
 }
 
 var defaultProbeValues = corev1.Probe{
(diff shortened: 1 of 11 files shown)


core/kogitoservice/probe.go
```diff
@@ -50,82 +66,61 @@ var defaultProbeValues = corev1.Probe{
 
 // getProbeForKogitoService gets the appropriate liveness (index 0) and readiness (index 1) probes based on the given service definition
```
sutaakar, line 67: Nitpick: please adjust the comment, the index 0 and 1 looks a bit confusing to me

sutaakar:
Disclaimer first - sorry for making seemingly endless story from relatively easy feature.
Would it have sense to actually leverage Actuator for SpringBoot application - https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/reference/html/production-ready-features.html#production-ready-kubernetes-probes?
I am just a bit worried that we cannot enforce its usage in the same way as for Quarkus (by adding healthcheck as a default extension in archetype).
If we would require SpringBoot Actuator then it may make the probe implementation simpler, as HTTP GET would become a default probe type for both environments, while we will still allow to use TCP probe, when user will define it in CR.
Also SpringBoot Actuator would allow having better information about application status and its possible problems, so it can be reflected directly by probes.

Kevin-Mok:
> Would it have sense to actually leverage Actuator for SpringBoot application

I'm not sure myself whether this is something we should require. @ricardozanini, do you have a take on this?

sutaakar:
We can add the Activator as a default dependency in our SpringBoot archetypes - https://github.com/kiegroup/kogito-runtimes/tree/master/archetypes , same way as we did for Quarkus

ricardozanini:
> > Would it have sense to actually leverage Actuator for SpringBoot application
> 
> This sounds useful, but I'm not sure whether this is something we should require. @ricardozanini, do you have a take on this?

I think if we can verify if the SB application has Actuator enabled, then we should definitely use it. Otherwise, won't make much sense.

ricardozanini:
> We can add the Activator as a default dependency in our SpringBoot archetypes - https://github.com/kiegroup/kogito-runtimes/tree/master/archetypes , same way as we did for Quarkus

That would work +1

Kevin-Mok:
/jenkins test

@sutaakar @ricardozanini I opened a PR for [KOGITO-4958](https://github.com/kiegroup/kogito-runtimes/pull/1232) to add Spring Boot Actuator as a default dependency to the Spring Boot archetypes and modified this PR to use the Spring Boot Actuator probe endpoints. I also removed the `HealthCheckProbe` field from the `ServiceDefinition` struct as it's not necessary when we can just check the runtime in `getProbeForKogitoService` to determine what the default probe fields should be.

### Testing
I made an image of process-springboot-example with the Spring Boot Actuator dependency and deployed with the following YAML to test if the default Spring Boot endpoints would work:
```yaml
kind: KogitoRuntime
metadata:
  name: sbe
spec:
  replicas: 1
  image: quay.io/kmok/process-springboot-example:actuator
  runtime: springboot
```

sutaakar:
Such issue is most likely caused by some stuck KogitoRuntime in the CI. Retriggering the tests may help.
/jenkins test

Kevin-Mok:
> Such issue is most likely caused by some stuck KogitoRuntime in the CI. Retriggering the tests may help.

@sutaakar I'm seeing the same issue in [Tristan's PR](https://github.com/kiegroup/kogito-operator/pull/840#issuecomment-822436847) this morning, and it's also happened for the last 3 runs of my PR.

sutaakar:
It should work now, at least for master.
/jenkins test

sutaakar:
Now it should really work.
/jenkins test

Kevin-Mok:
It seems that the tests are good now. The only ones failing are for this reason:
> I imagine that the Spring Boot tests will still fail regardless though because they don't have the Actuator dependency, so the probes will fail.

sutaakar (approved):
@Kevin-Mok Will you also take care of updating Kogito examples so the tests will pass completely?

Kevin-Mok:
@ricardozanini @spolti This is ready for review now.

core/kogitoservice/probe.go
```diff
@@ -21,18 +21,19 @@ import (
...
-	// QuarkusHealthCheckProbe probe implemented with Quarkus Microprofile Health. See: https://quarkus.io/guides/microprofile-health.
-	// the operator will set the probe to the default path /health/live and /health/ready for liveness and readiness probes, respectively.
-	QuarkusHealthCheckProbe HealthCheckProbeType = "quarkus"
-	// TCPHealthCheckProbe default health check probe that binds to port 8080
-	TCPHealthCheckProbe HealthCheckProbeType = "TCP"
+	livenessProbeType  ProbeType = "liveness"
+	readinessProbeType ProbeType = "readiness"
+	startupProbeType   ProbeType = "startup"
 
 	quarkusProbeLivenessPath  = "/q/health/live"
```
spolti, line 32: are these paths configurable on the operator?
what happens if the user decides to customize the heath check paths?

spolti (changes requested):

core/kogitoservice/probe.go
Kevin-Mok: > are these paths configurable on the operator?

Yes, they are. https://github.com/kiegroup/kogito-operator/blob/2b224716e9b712624c5d58d01d870f582acba2e1/kogito-operator.yaml#L1225-L1227

> what happens if the user decides to customize the heath check paths?

The values here are only used if the user doesn't specify a path in the YAML. Otherwise, their's will be used.

spolti: cool, thanks man!

spolti (approved):

PR #860, opened 2021-05-04
BAQE-1695: Test from VS Code interaction to cloud deployment

jomarko:
for more details see:
- https://issues.redhat.com/browse/BAQE-1695

This is still WIP, missing parts are
- [ ] to run actual asset manipulation in VS Code
- [ ] check more properly DMN service (processes return json, while DMN xml)
- [ ] tests more combinations by the data-table

The command for I used for running is stored in jira comment.

- [ ] You have read the [contributors' guide](https://github.com/kiegroup/kogito-operator/blob/master/README.md#contributing-to-the-kogito-operator)


test/pkg/framework/http.go
```diff
@@ -156,36 +157,47 @@ func ExecuteHTTPRequestC(client *http.Client, namespace string, requestInfo HTTP
 }
 
 // ExecuteHTTPRequestWithStringResponse executes an HTTP request and returns a string response in case there is no error
-func ExecuteHTTPRequestWithStringResponse(namespace string, requestInfo HTTPRequestInfo) (string, error) {
+func ExecuteHTTPRequestWithStringResponse(namespace string, requestInfo HTTPRequestInfo) (string, string, error) {
```
sutaakar, line 160: Returning content type as a function result doesn't look much flexible for me.
I think it may be better to refactor the implementation, putting the response body retrieval as a separate helper function.
That way you can adjust ExecuteHTTPRequestWithUnmarshalledResponse function to call ExecuteHTTPRequest, retrieve response body using helper function and access Content-Type header there, without exposing it here in ExecuteHTTPRequestWithStringResponse.

test/pkg/steps/data.go
```diff
@@ -193,8 +197,13 @@ func handleScenarioResult(data *Data, scenario *godog.Scenario, err error) {
...
-	if err != nil {
-		framework.GetMainLogger().Error(err, "Error while deleting temporary examples folder", "folderName", data.KogitoExamplesLocation)
+	temporaryFolders := []string{data.KogitoExamplesLocation, data.KieAssetLibraryLocation}
+
+	for i := 0; i < len(temporaryFolders); i++ {
+		err := framework.DeleteFolder(temporaryFolders[i])
+		if err != nil {
+			framework.GetMainLogger().Error(err, "Error while deleting temporary examples folder", "folderName", temporaryFolders[i])
+		}
 	}
```
sutaakar, line 207:
```suggestion
	for _, temporaryFolder := range temporaryFolders {
		err := framework.DeleteFolder(temporaryFolder)
		if err != nil {
			framework.GetMainLogger().Error(err, "Error while deleting temporary examples folder", "folderName", temporaryFolder)
		}
	}
```

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/cucumber/godog"
+	"github.com/kiegroup/kogito-operator/test/pkg/framework"
+	"github.com/kiegroup/kogito-operator/test/pkg/steps/mappers"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// TODO
+// Should be externalized to config.go ?
```
sutaakar, line 28: Depends whether those values are intended to be hardcoded or should be set/adjusted by user.

test/features/deploy_kogito_build_vs_code.feature
```diff
@@ -0,0 +1,31 @@
...
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+    And Project kie-asset-library is cloned
+    And Project kie-asset-library is built by maven with configuration:
+      | profile | dmn                                    |
+      | option  | -Dactive.structures=quarkus            |
+      | option  | -Dactive.config.sets=scesim            |
+      | native  | disabled                               |
+    Then Project "<project>" is generated in temporary folder
```
sutaakar, line 13: The step text looks misleading. On the first looks I would guess that it reference the kie-asset-library project.
Maybe it could be changed to something like `And Project "<project>" is generated kie-asset-library`

test/pkg/steps/maven.go
```diff
@@ -71,7 +76,9 @@ func (data *Data) localServiceBuiltByMavenWithProfileAndOptions(serviceName stri
 	if mavenConfig.Native {
 		mvnCmd = mvnCmd.Profiles(nativeProfile)
 	}
-	output, err := mvnCmd.Execute("clean", "package")
-	framework.GetLogger(data.Namespace).Debug(output)
```
sutaakar, line 75: Please don't remove this debug entry, can come handy.

jomarko: sure, will revert

test/pkg/steps/kie_asset_library.go
jomarko: ok, if others will not say differently, I will keep it as constants here.

test/features/deploy_kogito_build_vs_code.feature
```diff
@@ -0,0 +1,31 @@
...
+
+
+  Scenario Outline: Build <runtime> binary build with native <native> using KogitoBuild
+    Given Project "<project>" assets are re-marshalled by VS Code
+    And Project "<project>" is built from temporary folder by maven
```
sutaakar, line 18: Thinking about how to properly describe the action done in here.
The text itself doesn't provide much context, there is no hint of kie-asset-library. 
Maybe 'Project "<project>" in kie-asset-library is built by maven'?

sutaakar: With similar syntax in step above.

sutaakar: and below too :)

jomarko: yes, thank you, will update the steps of the feature file, your suggestions sounds good to me

test/features/deploy_kogito_build_vs_code.feature
```diff
@@ -0,0 +1,31 @@
...
+      | option  | -Dactive.config.sets=scesim            |
+      | native  | disabled                               |
+    And Project "<project>" is generated by kie-asset-library
+
+
+  Scenario Outline: Build <runtime> binary build with native <native> using KogitoBuild
+    Given Project "<project>" assets are re-marshalled by VS Code
+    And Project "<project>" in kie-asset-library is built by maven
+      
+    When Deploy <runtime> project "<project>" from kie-asset-library on openshift with configuration:
```
sutaakar, line 20: Please drop openshift mention, the tests could be executed on plain Kubernetes too.

jomarko: good point, addressed in last commit

test/pkg/framework/http.go
jomarko: Please see my latest commit

jomarko:
Command I use locally for running:
```
rm test/logs/ -rf;  make run-tests operator_image=quay.io/kiegroup/kogito-operator operator_tag=1.5.0 local=true cr_deployment_only=true feature=features/deploy_kogito_build_vs_code.feature runtime_application_image_registry=quay.io maven_mirror=https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/public tags=smoke namespace_name=jomarko-testing runtime_application_image_namespace=jomarko
```

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+	framework.GetLogger(data.Namespace).Info("Cloning kie-asset-library project", "URI", KieAssetLibraryGitRepositoryURI, "branch", KieAssetLibraryGitRepositoryBranch, "clonedLocation", data.KieAssetLibraryLocation)
+
+	cloneOptions := &git.CloneOptions{
+		URL:          KieAssetLibraryGitRepositoryURI,
+		SingleBranch: true,
+	}
+
+	var err error
+	reference := KieAssetLibraryGitRepositoryBranch
+	if len(reference) == 0 {
```
radtriste, line 98: same here, `reference` will not be empty

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+		return err
+	}
+	return nil
+}
+
+func (data *Data) projectInKieAssetLibraryIsBuiltByMaven(project string) error {
+	output, errCode := framework.CreateMavenCommand(data.KieAssetLibraryLocation+"/kie-assets-library-generate/target/"+project).
+		SkipTests().
+		Execute("clean", "install")
+	framework.GetLogger(data.Namespace).Debug(output)
```
radtriste, line 127: display output only in error or?

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return errCode
+}
+
+func (data *Data) projectAssetsAreRemarshalledByVsCode(project string) error {
+	installOutput, installErr := framework.CreateCommand("npm", "install").
+		WithRetry(framework.NumberOfRetries(1)).
+		InDirectory(data.KieAssetReMarshallerLocation).
+		Execute()
+	framework.GetLogger(data.Namespace).Debug(installOutput)
```
radtriste, line 139: display output only in error or?

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if installErr != nil {
+		return installErr
+	}
+
+	runOutput, runErr := framework.CreateCommand("npm", "run", "test:it", "KIE_PROJECT="+data.KieAssetLibraryLocation+"/kie-assets-library-generate/target/"+project).
+		WithRetry(framework.NumberOfRetries(1)).
+		InDirectory(data.KieAssetReMarshallerLocation).
+		Execute()
+	framework.GetLogger(data.Namespace).Debug(runOutput)
```
radtriste, line 149: display output only in error or?

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (data *Data) projectIsGeneratedByKieAssetLibrary(project string) error {
+	if _, err := os.Stat(data.KieAssetLibraryLocation + "/kie-assets-library-generate/target/" + project); !os.IsNotExist(err) {
+		return err
+	}
+	return nil
+}
+
+func (data *Data) projectInKieAssetLibraryIsBuiltByMaven(project string) error {
+	output, errCode := framework.CreateMavenCommand(data.KieAssetLibraryLocation+"/kie-assets-library-generate/target/"+project).
```
jstastny-cz, line 124: code format near + operators

jstastny-cz (commented):
I have just minor comments into the part of invocation of kie assets library.

I tend to agree with Tristan that we should start migrating things to kiegroup and then wrap this up. I should be able to do my part during this week - if that's alright?

test/features/deploy_kogito_build_vs_code.feature
```diff
@@ -0,0 +1,32 @@
...
+    Given Namespace is created
+    And Kogito Operator is deployed
+    And Project kie-asset-re-marshaller is cloned
+    And Project kie-asset-library is cloned
+    And Project kie-asset-library is built by maven with configuration:
+      | profile | dmn                                    |
+      | option  | -Dactive.structures=quarkus            |
+      | option  | -Dactive.config.sets=scesim            |
+      | native  | disabled                               |
+    And Project "<project>" is generated by kie-asset-library
```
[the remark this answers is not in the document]
jomarko: understand, good point

test/pkg/steps/data.go
```diff
@@ -86,7 +90,9 @@ func (data *Data) RegisterLogsKubernetesObjects(objects ...runtime.Object) {
 func (data *Data) BeforeScenario(scenario *godog.Scenario) error {
 	data.StartTime = time.Now()
 	data.Namespace = getNamespaceName()
-	data.KogitoExamplesLocation = createTemporaryFolder()
+	data.KogitoExamplesLocation = createTemporaryFolder("kogito-examples")
```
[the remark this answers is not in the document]
jomarko: I tried to apply, needs to be re-reviewed once I push changes

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/kiegroup/kogito-operator/test/pkg/framework"
+	"github.com/kiegroup/kogito-operator/test/pkg/steps/mappers"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// TODO
+// Should be externalized to config.go ?
+const (
+	KieAssetLibraryGitRepositoryURI         = "https://github.com/jstastny-cz/kie-asset-library-poc"
```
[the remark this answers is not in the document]
jomarko: I think I can not convert back to draft, please keep labels to not merge until you think it is needed

test/pkg/steps/kie_asset_library.go
```diff
@@ -0,0 +1,158 @@
...
+	framework.GetLogger(data.Namespace).Info("Cloning kie-asset-re-marshaller project", "URI", KieAssetReMarshallerGitRepositoryURI, "branch", KieAssetReMarshallerGitRepositoryBranch, "clonedLocation", data.KieAssetReMarshallerLocation)
+
+	cloneOptions := &git.CloneOptions{
+		URL:          KieAssetReMarshallerGitRepositoryURI,
+		SingleBranch: true,
+	}
+
+	var err error
+	reference := KieAssetReMarshallerGitRepositoryBranch
+	if len(reference) == 0 {
```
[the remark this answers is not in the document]
jomarko: well spotted , thank you

test/pkg/steps/kie_asset_library.go
jomarko: well spotted , thank you

test/pkg/steps/kie_asset_library.go
jomarko: I was asked earlier to keep debug log message for such commands, see https://github.com/kiegroup/kogito-operator/pull/860#pullrequestreview-651149860

test/pkg/steps/kie_asset_library.go
jomarko: I was asked earlier to keep debug log message for such commands, see https://github.com/kiegroup/kogito-operator/pull/860#pullrequestreview-651149860

test/pkg/steps/kie_asset_library.go
jomarko: well spotted

test/features/deploy_kogito_build_vs_code.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Deploy Kogito Build - after assets processed by VS code
+  
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+    And Project kie-asset-re-marshaller is cloned
+    And Project kie-asset-library is cloned
+    And Project kie-asset-library is built by maven with configuration:
+      | profile | dmn                                    |
```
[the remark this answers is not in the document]
jomarko: tried to apply, needs to be re-reviewed, once pushed

jomarko:
@radtriste @jstastny-cz changes are in

test/features/deploy_kogito_build_vs_code.feature
```diff
@@ -0,0 +1,31 @@
...
+
+  Background:
+    Given Namespace is created
+    And Kogito Operator is deployed
+    And Project kie-asset-re-marshaller is cloned
+    And Project kie-asset-library is cloned
+    And Project kie-asset-library is built by maven with configuration:
+      | option  | -Dactive.definitions=dmn-generated |
+      | option  | -Dactive.structures=quarkus        |
+      | native  | disabled                           |
```
jomarko, line 12: @sutaakar @radtriste hello, We would need this to be different when running community vs product, is that possible to parse and pass some value from the job env?

jstastny-cz: Something like passing env to the maven build on behind, let's say traditional `-Dproductized` to activate a profile in kie-asset-library and we would pass that property from outside of BDD suite (like in CI job) ... is that possible?

radtriste: native enabled scenarios should have the `@native` tag so it is easy to decide which tests would run

radtriste: but this should not be into the `Background` then

radtriste: oh just understood this is not only for native ...
Then I would suggest to have different scenario outline examples with some tag on the scenario

test/pkg/steps/data.go
```diff
@@ -36,13 +36,25 @@ var (
 	logsKubernetesObjects = []runtime.Object{&imgv1.ImageStreamList{}, &v1beta1.KogitoRuntimeList{}, &v1beta1.KogitoBuildList{}, &v1beta1.KogitoSupportingService{}, &v1beta1.KogitoInfraList{}, &olmapiv1alpha1.ClusterServiceVersionList{}}
 )
 
+type thirdPartyProject struct {
```
radtriste, line 39: oh I was confusing in what I said yesterday I guess.
You can directly define a type to be a string instead of having a struct. See https://github.com/kiegroup/kogito-operator/blob/master/api/kogitoinfra_types.go#L24 as an example

test/pkg/steps/data.go
```diff
@@ -36,13 +36,25 @@ var (
...
 // Data contains all data needed by Gherkin steps to run
 type Data struct {
-	Namespace              string
-	StartTime              time.Time
-	KogitoExamplesLocation string
-	ScenarioName           string
-	ScenarioContext        map[string]string
+	Namespace       string
+	StartTime       time.Time
+	Location        map[thirdPartyProject]string
```
radtriste, line 55: rename to ThirdPartyProjectLocation ?

test/pkg/steps/kie_asset_library.go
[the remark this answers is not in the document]
radtriste: Check under the reviewers, you should have the `Still in progress? Convert to draft`

test/pkg/steps/kie_asset_library.go
radtriste: I guess we should "debug" if no error but if error occurs, we should still put it as error log I guess

test/pkg/steps/kie_asset_library.go
radtriste: I guess we should "debug" if no error but if error occurs, we should still put it as error log I guess

test/pkg/steps/kie_asset_library.go
radtriste: I guess we should "debug" if no error but if error occurs, we should still put it as error log I guess

test/pkg/steps/data.go
jomarko: oh, ok, sorry, that is my low skills in go, will fix it

test/features/deploy_kogito_build_vs_code.feature
jstastny-cz: hm, that doesn't scale much - we then need to set kogito version to the mvn build using `-Dversion.kogito`, then the possible `-Dproductized` and there can be more property overrides needed. Especially in case of version.kogito it's not possible to hardcode the value as example.
Is there a way to promote env variables through the go run into the maven invocation? If not, is that something that would be possible to do? And could someone help with that? :-)

radtriste: This is the test script used to launch the tests: https://github.com/kiegroup/kogito-operator/blob/master/hack/run-tests.sh#L22
And the different test options in Golang are defined here: https://github.com/kiegroup/kogito-operator/blob/master/test/pkg/config/config.go

We could have a way to give the Kogito Version to Maven but this is a global setting and not specific to a scenario. So it would be applied to all the Maven commands.
`-Dproductized` is a static configuration for prod scenario or ?

I wouldn't be a big fan of having scenarios which are awaiting for specific test configuration coming via command line.
But like we have the [`native` config](https://github.com/kiegroup/kogito-operator/blob/master/test/pkg/steps/maven.go#L71) on Maven, we could have a `kogito_version` to true which would add the kogito version if it has been given as a test option. But we should not have a strong dependency on it.

Note also that specific scenario for product should be set in https://github.com/kiegroup/rhpam-kogito-operator

radtriste: I guess we could review which maven properties are needed and see how they could be integrated into the framework, that would be more efficient ;)

sutaakar:
Looks good to me.
Just a generic note, originally all BDD steps were designed to be in passive tense. Thinking about it, it may shift the intended sentence meaning and be less descriptive. IMHO it would be better to use active tense for steps which perform some action, keeping passive tense for steps which check the status or result, what do you think?

spolti:
@radtriste  is it still needed?

radtriste:
@jomarko ?

jomarko:
Closing this effort. Similar tests should be probably developed rather in BA Studio testsuite where a deployment a decision on cloud is one of key features.

PR #925, opened 2021-07-08
[KOGITO-5478] - Management console does not work on OpenShift

vaibhavjainwiz:

Jira issue : https://issues.redhat.com/browse/KOGITO-5478

- [x] You've ran `make before-pr` and everything is working accordingly
- [x] You've added a [RELEASE_NOTES.md](https://github.com/kiegroup/kogito-operator/blob/master/RELEASE_NOTES.md) entry regarding this change

### DIFF
--- a/api/v1beta1/kogitosupportingservice_types.go
+++ b/api/v1beta1/kogitosupportingservice_types.go
@@ -107,14 +107,15 @@ func (k *KogitoSupportingService) GetSupportingServiceStatus() api.KogitoSupport
 type KogitoSupportingServiceList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
-	Items           []*KogitoSupportingService `json:"items"`
+	Items           []KogitoSupportingService `json:"items"`
 }
 
 // GetItems ...
 func (k *KogitoSupportingServiceList) GetItems() []api.KogitoSupportingServiceInterface {
 	models := make([]api.KogitoSupportingServiceInterface, len(k.Items))
 	for i, v := range k.Items {
-		models[i] = v
+		item := v
+		models[i] = &item
 	}
 	return models
 }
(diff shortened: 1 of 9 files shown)


sutaakar (approved):
Looks good to me, Management console works as expected.

radtriste (approved):

core/manager/supportingservice_manager.go
```diff
@@ -50,7 +50,7 @@ func NewKogitoSupportingServiceManager(context operator.Context, supportingServi
 }
 
 func (k kogitoSupportingServiceManager) EnsureSingletonService(namespace string, resourceType api.ServiceType) error {
-	k.Log.Info("Ensuring only single instance of supporting service exists")
+	k.Log.Info("Ensuring only single instance of supporting service exists", "resourceType", resourceType)
```
ricardozanini, line 53: We should start investigating [admission controllers](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html) instead of these checks.

api/v1beta1/kogitosupportingservice_types.go
```diff
@@ -107,14 +107,14 @@ func (k *KogitoSupportingService) GetSupportingServiceStatus() api.KogitoSupport
 type KogitoSupportingServiceList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
-	Items           []KogitoSupportingService `json:"items"`
+	Items           []*KogitoSupportingService `json:"items"`
```
ricardozanini, line 110: Not sure if changing the default kubebuilder structure is a good idea. Can you try a different approach?

ricardozanini (changes requested):
@vaibhavjainwiz, I think we might need to take another look at this issue. Changing the default generated type is not a good option unless we have a strong motivation behind it. I couldn't find another example of a CR List being a pointer instead, like you're proposing.

core/manager/supportingservice_manager.go
vaibhavjainwiz: sure Zanini

api/v1beta1/kogitosupportingservice_types.go
vaibhavjainwiz: done

ricardozanini (approved):

spolti (approved):

PR #989, opened 2021-08-23
[KOGITO-5102] - Long KogitoRuntime Name + Namespace causes Route Error in KogitoRuntime

amygaoo:
See: https://issues.redhat.com/browse/KOGITO-5102

## Description
When deploying a KogitoRuntime on OpenShift where `<name>-<namespace>` combined is longer than 63 characters, Kogito fails to create a route because the host does not conform to DNS 1123 naming conventions.

For this fix I added a spec.host field to the KogitoRuntime CRD to allow users to manually set a custom host. In addition, the host input is validated before route creation and defaults to the original route creation steps if host is invalid.

- [x] You have read the [contributors' guide](https://github.com/kiegroup/kogito-operator/blob/main/README.md#contributing-to-the-kogito-operator)
- [x] You've ran `make before-pr` and everything is working accordingly
- [x] You've added a [RELEASE_NOTES.md](https://github.com/kiegroup/kogito-operator/blob/main/RELEASE_NOTES.md) entry regarding this change

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,5 +5,5 @@
 - [KOGITO-5512](https://issues.redhat.com/browse/KOGITO-5512) - controller-gen doesn't generate CRD's
 - [KOGITO-4470](https://issues.redhat.com/browse/KOGITO-4470) - ConfigMap required ownerReference as controller
 - [KOGITO-5573](https://issues.redhat.com/browse/KOGITO-5573) - Kogito Operator Data-Index Unable to connect to Infinispan Instance
-
+- [KOGITO-5102](https://issues.redhat.com/browse/KOGITO-5102) - Long KogitoRuntime Name + Namespace causes Route Error in KogitoRuntime
 ## Known Issues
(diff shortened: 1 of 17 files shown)


sutaakar:
Unfortunately not working as expected on OCP. When I create:
```
kind: KogitoRuntime
metadata:
  name: example-quarkus
spec:
  image: quay.io/kiegroup/examples-process-quarkus-example-nightly:latest
  host: "test.hostname.com"
```
I get an error:
```
{"level":"error","ts":1629797259.442733,"logger":"controller","msg":"Reconciler error","reconcilerGroup":"app.kiegroup.org","reconcilerKind":"KogitoRuntime","controller":"kogitoruntime","name":"example-quarkus","namespace":"ksuta-test","error":"Route.route.openshift.io \"example-quarkus\" is invalid: spec.host: Forbidden: you do not have permission to set the host field of the route","stacktrace":....
```
The issue seems to come from missing permission, see https://github.com/openshift/origin/issues/22612
The question is whether it has sense to provide such permissions or rather address route configuration in different way based on discussions in https://issues.redhat.com/browse/KOGITO-4730

sutaakar:
Works as expected now.
/jenkins test

sutaakar:
While the PR works as expected, I think that we should clarify how to represent the hostname in CR related to https://issues.redhat.com/browse/KOGITO-4730. If we will change the structure later to include skipping Route creation or TSL specification then we introduce backward incompatibilities.

spolti:
will be addressed as part of https://issues.redhat.com/browse/KOGITO-5903

PR #999, opened 2021-09-03
[KOGITO-5740] Unnecessary reconciliations triggered for Service

amygaoo:
See:  https://issues.redhat.com/browse/KOGITO-5740

## Description
Kogito operator isn't able to properly compare Service resources because the deployed Service object has properties `ipFamilies` and `ipFamilyPolicy`, while the requested Service object is missing these properties. 

~~The fix I completed is to skip the comparision for `ipFamilies` and `ipFamilyPolicy` when the requested Service object is missing these properties. If the requested Service object has these properties, then the comparator will function as usual.~~

I refactored our custom service comparator to compare only the service spec fields our operator initializes and disabled the default service comparator from operator utils. This ensures the comparator only catches differences in operator specified fields, instead of auto generated fields from the cluster.

- [ ] You have read the [contributors' guide](https://github.com/kiegroup/kogito-operator/blob/main/README.md#contributing-to-the-kogito-operator)
- [ ] You've ran `make before-pr` and everything is working accordingly
- [ ] You've added a [RELEASE_NOTES.md](https://github.com/kiegroup/kogito-operator/blob/main/RELEASE_NOTES.md) entry regarding this change


ricardozanini (changes requested):
@amygaoo, albeit this is a fine workaround, I think the approach we should take here is to compare only the necessary fields, i.e., the ones we set/manage. Take a look here: https://github.com/kiegroup/kogito-operator/blob/761552207d271d70d6c4f785adb841ae97293103/core/infrastructure/service.go#L42.

This is the service we create. We should only compare the fields we are interested in. This way, we avoid a similar bug in the future.

amygaoo:
> @amygaoo, albeit this is a fine workaround, I think the approach we should take here is to compare only the necessary fields, i.e., the ones we set/manage. Take a look here:
> 
> https://github.com/kiegroup/kogito-operator/blob/761552207d271d70d6c4f785adb841ae97293103/core/infrastructure/service.go#L42
> 
> .
> This is the service we create. We should only compare the fields we are interested in. This way, we avoid a similar bug in the future.

This is a good idea, I will look into a way to create a copy of the service with only the required fields for comparision. In this case, the comparator may have a different approach than the other ones.

amygaoo:
> > @amygaoo, albeit this is a fine workaround, I think the approach we should take here is to compare only the necessary fields, i.e., the ones we set/manage.
> > This is the service we create. We should only compare the fields we are interested in. This way, we avoid a similar bug in the future.
> 
> This is a good idea, I will look into a way to create a copy of the service with only the required fields for comparision. In this case, the comparator may have a different approach than the other ones.

@ricardozanini After looking at how our comparators are called, I would like to check that I am on the right track here. Currently we are using the default comparator from Operator Utils to compare services:  https://github.com/RHsyseng/operator-utils/blob/0f7acfb7a492851cad0ca5eb327b85cee0aa7e10/pkg/resource/compare/defaults.go#L383
So with this function we don't have control over which fields are being compared.

I'm thinking that for services we can not set the default comparator, and instead call the default comparator from our custom comparator https://github.com/kiegroup/kogito-operator/blob/ce1f6eda31269bacfea005b4054211ede602a7ba/core/framework/comparator.go#L202
, where we can create a seperate copy of deployed and requested reasources with only the required fields. 
Or another option is to create our custom comparator to reuse some of the code from the Operator Utils and then only compare the fields we require.  For this option I'm not sure what the policy is about using code from other repos, and if we are allowed to use it if they have an Apache License 2.0?
What is your take on this? Or is there another option that I am missing?

ricardozanini:
@amygaoo, I believe we already have some sort of "custom comparator" in our codebase. I remember creating our own. Also, the `operator-utils` allows injection of custom comparators, so should be pretty straightforward adding our own.

core/kogitoservice/deployer_resources.go
```diff
@@ -217,3 +216,49 @@ func (s *serviceDeployer) mountEnvsOnDeployment(deployment *appsv1.Deployment) {
...
+		var pairs [][2]interface{}
+		pairs = append(pairs, [2]interface{}{svcDeployed.Name, svcRequested.Name})
+		pairs = append(pairs, [2]interface{}{svcDeployed.Namespace, svcRequested.Namespace})
+		pairs = append(pairs, [2]interface{}{svcDeployed.Labels, svcRequested.Labels})
+		pairs = append(pairs, [2]interface{}{svcDeployed.Spec.Ports, svcRequested.Spec.Ports})
+		pairs = append(pairs, [2]interface{}{svcDeployed.Spec.Selector, svcRequested.Spec.Selector})
+		pairs = append(pairs, [2]interface{}{svcDeployed.Spec.Type, svcRequested.Spec.Type})
+		equal := compare.EqualPairs(pairs)
+
+		// TODO: Figure out how to log differences
```
ricardozanini, line 248: If you do, we will have to add it to every log message we have in this code base. :sweat_smile: 
Don't worry, if you prefer, we can open a JIRA to track this down in future sprints.

amygaoo: You are just referring to the TODO comment right? I just forgot to delete that line, since prior to this commit I wasn't able to get logging to work 😅 ... I just deleted the comment.

ricardozanini (approved):
Well done! Many thanks, @amygaoo

vaibhavjainwiz (approved):

sutaakar (approved):
Looks good to me.

spolti (approved):

amygaoo:
> can you please rebase it?

@spolti rebased!

radtriste:
> @radtriste @sutaakar does this fix needs to be on 1.11.1 as well?

That would be great for Product I would say but I think it won't be included into 1.11.1 unless the cherry-pick is done in the next 5 hours ...

radtriste:
@amygaoo Could you cherry-pick this to 1.11.x ?

PR #1033, opened 2021-09-23
[KOGITO-5903] - Configuring Optional Routes in KogitoRuntime

amygaoo:
See: https://issues.redhat.com/browse/KOGITO-5903
## Changes
A `spec.route` boolean field is added to the KogitoRuntime CR to enable/disable route creation in Openshift, which is `False` by default. Routes are always enabled for Supporting Services. 

In the case of route creation failure (like in the case of [Long Namespace and Name Route Error](https://issues.redhat.com/browse/KOGITO-5102)), the KogitoRuntime `externalURI` is updated with the error message and `status.conditions` is updated with with the error as well.

- [ ] You have read the [contributors' guide](https://github.com/kiegroup/kogito-operator/blob/main/README.md#contributing-to-the-kogito-operator)
- [ ] You've ran `make before-pr` and everything is working accordingly
- [ ] You've added a [RELEASE_NOTES.md](https://github.com/kiegroup/kogito-operator/blob/main/RELEASE_NOTES.md) entry regarding this change


ricardozanini (approved):
Seems fine. But this PR is another reason we should drop completely the responsibility of deploying our services via operator. This task can be done by other actors within the cluster.

apis/app/v1beta1/kogitosupportingservice_types.go
```diff
@@ -38,6 +38,11 @@ func (k *KogitoSupportingServiceSpec) GetRuntime() api.RuntimeType {
 	return api.QuarkusRuntimeType
 }
 
+// IsRouteEnabled ...
+func (k *KogitoSupportingServiceSpec) IsRouteEnabled() bool {
+	return true
```
radtriste, line 43: We should allow the user to disable that route if needed. They may want to manage their own route system.

radtriste (commented):
I still think we should implement https://issues.redhat.com/browse/KOGITO-1940 before merging that one. Else kogitoruntime <-> jobs-service will not work out of the box ...

ricardozanini:
> I still think we should implement https://issues.redhat.com/browse/KOGITO-1940 before merging that one. Else kogitoruntime <-> jobs-service will not work out of the box ...

Nice catch! I do agree with this.

apis/app/v1beta1/kogitosupportingservice_types.go
amygaoo: In this case should the`spec.route` boolean flag control both Supporting Service routes and the normal KR routes?

radtriste: kogitosupportingservice should have https://github.com/kiegroup/kogito-operator/pull/1033/files#diff-5612b3ae7d9c41e756f09eb33d1fe50cb6e42d04380fd3fa0dd937a3f6f851dcR47 or maybe a `disableRoute` or similar which would be taken in that method

spolti:
> I still think we should implement https://issues.redhat.com/browse/KOGITO-1940 before merging that one. Else kogitoruntime <-> jobs-service will not work out of the box ...

IMHO it is not mandatory to address KOGITO-1940 first, the default behavior of this new feature must be exactly as it was before, but adding the choice to disable route creation, in that case user must be aware of any route must be created manually, including jobs service.

spolti (approved):

radtriste:
I thought the whole point was to not have automatic creation of route or ?
`in that case user must be aware of any route must be created manually, including jobs service.` I don't get that part

apis/app/v1beta1/kogitosupportingservice_types.go
amygaoo: Yes, I can add the field to [KogitoServiceSpec struct](https://github.com/kiegroup/kogito-operator/blob/178e5d4b83278e6540123a55ca9f745f0cc14bc1/apis/app/v1beta1/kogitoservices.go#L164), so the field exists for both KR spec and Supporting Service spec. If that flag is set to True in KR CR, then routes are configured for both KR and Supporting services; one field in CR controls both route configurations. 
Or, the other solution is to have 2 different fields in the CR, one controlling the SupportingService Routes and one controlling the regular KR routes. The first option seems cleaner, but it depends on how much flexibility we want to give the user.

amygaoo:
> I still think we should implement https://issues.redhat.com/browse/KOGITO-1940 before merging that one. Else kogitoruntime <-> jobs-service will not work out of the box ...

I am wondering since KogitoSupportingServices Routes are always enabled right now, jobs service should work? Although if we make SupportingServiceRoutes configurable as well, then yes jobs-service will not automatically work.

radtriste:
> ondering since SupportingServices Routes are always enabled right now, jobs service s

Yes there are. But when KR is calling the jobs-service to create a job, it does send a callback URL and that one is the KR route for now. So if it is disabled by default, there will be no SERVICE_URL env available to the jobs-service when creating the jobs.
That is why KOGITO-1940 is important, so we can use services instead of routes

apis/app/v1beta1/kogitosupportingservice_types.go
radtriste: well I would prefer flexibility here. Service routes should not depend on KR anyway.
For me, **when KOGITO-1940 is implemented**, we could have:
- Kogito Runtime -> Route is disabled by default and can be enabled by the user
- Kogito Supporting Service -> Route is enabled by default and can be disabled by the user

spolti:
> > > I still think we should implement https://issues.redhat.com/browse/KOGITO-1940 before merging that one. Else kogitoruntime <-> jobs-service will not work out of the box ...
> > 
> > 
> > I am wondering since SupportingServices Routes are always enabled right now, jobs service should work? Although if we make SupportingService Routes configurable as well, then yes jobs-service will not automatically work. In that case, I am cool with putting this on hold until KOGITO-1940 is addressed.
> 
> Yes there are. But when KR is calling the jobs-service to create a job, it does send a callback URL and that one is the KR route for now. So if it is disabled by default, there will be no SERVICE_URL env available to the jobs-service when creating the jobs. That is why KOGITO-1940 is important, so we can use services instead of routes

That's why I said, if user disable route creation, they will have to expose manually. In this case inf needed, manual configuration as well.

vaibhavjainwiz (approved):

core/infrastructure/reconciliation_error.go
```diff
@@ -206,7 +228,6 @@ func (r *reconciliationErrorHandler) GetReconcileResultFor(err error) (ctrl.Resu
 		reconcileError := err.(ReconciliationError)
 		r.Log.Info("Waiting for all resources to be created, re-scheduling.", "reason", reconcileError.reason, "requeueAfter", reconcileError.reconciliationInterval)
 		reconcileResult.RequeueAfter = reconcileError.reconciliationInterval
-		reconcileResult.Requeue = true
```
sutaakar, line 209: Why is the requeue removed here?

radtriste:
> > > > I still think we should implement https://issues.redhat.com/browse/KOGITO-1940 before merging that one. Else kogitoruntime <-> jobs-service will not work out of the box ...
> > > 
> > > 
> > > I am wondering since SupportingServices Routes are always enabled right now, jobs service should work? Although if we make SupportingService Routes configurable as well, then yes jobs-service will not automatically work. In that case, I am cool with putting this on hold until KOGITO-1940 is addressed.
> > 
> > 
> > Yes there are. But when KR is calling the jobs-service to create a job, it does send a callback URL and that one is the KR route for now. So if it is disabled by default, there will be no SERVICE_URL env available to the jobs-service when creating the jobs. That is why KOGITO-1940 is important, so we can use services instead of routes
> 
> That's why I said, if user disable route creation, they will have to expose manually. In this case inf needed, manual configuration as well.

oki so we allow to disable the route creation on both KogitoRuntime and SupportingService.
And we should put a status message when the route cannot be created to disable it and create a manual one. wdyt ?

sutaakar:
Seems to work as expected, waiting for supporting service changes before approving the PR.

spolti:
> oki so we allow to disable the route creation on both KogitoRuntime and SupportingService.
> And we should put a status message when the route cannot be created to disable it and create a manual one. wdyt ?

That's a good Idea, @amygaoo do you think you can add the changes suggested by Tristan?

amygaoo:
Yes, let me just clarify that I understand correctly.
> oki so we allow to disable the route creation on both KogitoRuntime and SupportingService. 

For disabling the route creation for SupportingService, should that be a seperate field on the KR CR (or same field as the already added)? Should supportingService routes be disabled by default?

> And we should put a status message when the route cannot be created to disable it and create a manual one. wdyt ?

This status message is for when the route is enabled, and can I get an example scenerio of when the route cannot be created? Its is currently handled for normal KR routes, just wondering when this happens for supporting services.

core/infrastructure/reconciliation_error.go
amygaoo: It is removed so we can handle no requeue and requeue easily in that code. The [reconciler logic](https://github.com/kubernetes-sigs/controller-runtime/blob/9d7bcf7f784f556882e881c1611240a8a9fdaeb2/pkg/reconcile/reconcile.go#L31) states that if `RequeueAfter` is set, then requeue is automatically true.  We set a`RequeueAfter` interval if and only if it needs to be requeued, so we do not need to explicitly set `Requeue` anymore.

radtriste:
> Yes, let me just clarify that I understand correctly.
> 
> > oki so we allow to disable the route creation on both KogitoRuntime and SupportingService.
> 
> For disabling the route creation for SupportingService, should that be a seperate field on the KR CR (or same field as the already added)? Should supportingService routes be disabled by default?

I guess having a `DisableRoute` on both KogitoRuntime and KogitoSupportingService.
@vaibhavjainwiz @spolti could we maybe put that field in the `KogitoServiceSpec` ?

> > And we should put a status message when the route cannot be created to disable it and create a manual one. wdyt ?
> 
> This status message is for when the route is enabled, and can I get an example scenerio of when the route cannot be created? Its is currently handled for normal KR routes, just wondering when this happens for supporting services.

well it seems it is already handled by the route_reconciler as the supporting services are implementing the KogitoService interface

core/infrastructure/reconciliation_error.go
sutaakar: ok, thanks

vaibhavjainwiz:
> > Yes, let me just clarify that I understand correctly.
> > > oki so we allow to disable the route creation on both KogitoRuntime and SupportingService.
> > 
> > 
> > For disabling the route creation for SupportingService, should that be a seperate field on the KR CR (or same field as the already added)? Should supportingService routes be disabled by default?
> 
> I guess having a `DisableRoute` on both KogitoRuntime and KogitoSupportingService. @vaibhavjainwiz @spolti could we maybe put that field in the `KogitoServiceSpec` ?
> 
> > > And we should put a status message when the route cannot be created to disable it and create a manual one. wdyt ?
> > 
> > 
> > This status message is for when the route is enabled, and can I get an example scenerio of when the route cannot be created? Its is currently handled for normal KR routes, just wondering when this happens for supporting services.
> 
> well it seems it is already handled by the route_reconciler as the supporting services are implementing the KogitoService interface

@radtriste To achieve this we have to move `DisableRoute` field to `KogitoServiceSpec` and need to throw an error from `route_reconciler` when routes are disabled.

radtriste:
After discussion with @vaibhavjainwiz, the route creation should not be blocking for the rest of the deployer. Which means the potential error reported in the route creation should be handled but not block the rest of the deployment. We could so create a k8s event and update the status later about that failure. But application would still be deployed correctly in k8s.
then in the status, we should then suggest to look into the events/logs and disable route if needed, to be created manually (+ add KOGITO_SERVICE_URL env to the application)

spolti:
sounds good to me.
a side note on the docs will be required as well.

amygaoo:
> changes looks good, however needs to be rebased.

Yup, I realized I still need to add another route status conditions part before it is complete.

PR #1215, opened 2022-06-16
KIECLOUD-639 - Remove lookaside cache usage in OSBS

AntStephenson:
JIRA: https://issues.redhat.com/browse/KIECLOUD-639

This PR introduces an additional module for prod/OSBS builds that strips out the artifacts section to avoid impacting local builds, Cachito is already downloading and unzipping the src into a specific directory. With this section removed CEKit will no longer push main.go to the lookaside cache. 

1.13.x PR : https://github.com/kiegroup/kogito-operator/pull/1214

### DIFF
--- a/modules/org.kie.kogito.rhpam.builder.prod/module.yaml
+++ b/modules/org.kie.kogito.rhpam.builder.prod/module.yaml
@@ -1,5 +1,5 @@
 name: org.kie.kogito.rhpam.builder.prod
-version: "7.11.0"
+version: "7.13.0"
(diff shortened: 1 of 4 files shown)


modules/org.kie.kogito.rhpam.builder.prod/module.yaml
```diff
@@ -0,0 +1,10 @@
+name: org.kie.kogito.rhpam.builder.prod
+version: "7.11.0"
```
spolti, line 2:
```suggestion
version: "7.13.0"
```

PR #1488, opened 2023-06-05
[BXMSPROD-2057] Updating references to the new kie-ci actions folder

bacciotti:


.github/workflows/jenkins-tests-PR.yml
```diff
@@ -7,19 +7,18 @@ on:
   pull_request:
```
lampajr, line 7: @bacciotti I am not very convinced about this change as previously the `jenkins-tests-PR` was triggering `.ci/jenkins/dsl/test.sh` which would have cloned `kogito-pipelines:1.13.x`, and executing this script https://github.com/kiegroup/kogito-pipelines/blob/1.13.x/dsl/seed/scripts/seed_test.sh - whereas with this new change we will invoke https://github.com/kiegroup/kogito-pipelines/blob/main/dsl/seed/scripts/seed_test.sh which is slightly different (maybe the final result is the same but I don't know TBH, @radtriste you know?).

Anyway if we are not sure if the behavior is exactly the same and since this is a legacy branch I would keep it as it is mainly because it was not referencing _kogito-pipelines github actions_, so no real need to change it. On the other hand if we can confirm the test is actually the same I agree on updating it.

radtriste: The test script is doing differently ... I would keep as it is and not modify.
Only main should be updated IMO

lampajr: I agree, @bacciotti would you mind to revert the `jenkins-tests-PR` workflow? the rest looks good to me

bacciotti: @lampajr done

lampajr (approved):

radtriste:
@bacciotti please add the branch into the PR title so we know directly on which branch we are. 
Like:
```
[1.13.x] [BXMSPROD-2057] Updating references to the new kie-ci actions folder
```

radtriste (approved):

Ginxo (approved):
LGTM

lampajr (approved):

